PR #506, opened 2014-10-13
Manuscript viewer (First sketch)

uliska:
This is *not* to be merged but to let you comment on the initial infrastructure.
I have:

- a dockable panel with a widget
    - a View(QScrollArea)
        - an AbstractManuscriptContainer with
            - `ManuscriptImageContainer` or a (nonexistent) PDF container

The container is a simple QWidget() containing a list of `ManuscriptPage()` items. These currently are simple labels but will hold more information on the manuscript page later.

In this first sketch a FileDialog will pop up on startup, where you can select one or more images to be displayed. Of course there will have to be more options and actions. Currently I'm only interested in comments about the general approach.


PeterBjuhr:
This is still not ready to be merged. The commit was more of a bump.

I think as a first step we should support viewing an external pdf. Later we could add support for other formats.

uliska:
I've now looked at your new code, and while I have to admit I couldn't fully follow it through the import cascades I have the impression it is what has to be done now.
Having a PDF file open in that viewer and have it already scroll and zoom is quite promising.

So please let me outline how I thought this should be integrated (this is not binding of course).

Regardless of supporting different file formats this tool should support multiple manuscripts for a score. For this I suggest a new metadata file that *can* be present or not (of course the tool is also useful if one can just open arbitrary files through the file dialog).

I suggest supporting a file that is to be stored next to the main document, with a derived name like e.g. `base-document-name.manuscripts.mei`. This file describes many aspects of the manuscripts and should be an MEI file, representing the MEI header and being able to hold the complete metadata of the manuscript(s).
While this may not be the most straightforward approach I think it is a good idea because

- it is nevertheless a *good* approach
- it gives us a "free" opportunity to get first-hand MEI experience
- it may be a good decision if we once should have Frescobaldi be a kind of MEI editor too.

The metadata of the most immediate use that should be stored in that file are the actual files (e.g. one file with N pages or a set of file with a list or regex to match contained page files) and the measure information. This should be at least which measures are on what page/file and can be extended (by the user) with defining the actual locations for each measure. That way the manuscript viewer can by in sync with the input and the Music View. Additionally it should be possible to display the music we are currently editing (provided Frescobaldi's input editor knows that too).

It should then be possible to switch display between the manuscripts or even display the corresponding measures side by side. This will make the manuscript viewer an invaluable tool for scholarly editing (although not an innovation - this is actually what the Edirom tools already offer).
Another thing I would like to see (then) is to have LilyPond compile alternative versions respecting the page layout of the different manuscripts (and have them compiled and cached in background) so switching manuscripts would instantly switch the rendered music.

This deliberately leaves out concrete considerations of the user interface. I think we should first get an idea about the underlying concepts before tinkering around with GUI elements.

uliska:
Here is the [full description](http://music-encoding.org/documentation/guidelines2013/header/) of the `<meiHead>` element. For the actual purpose of the Manuscript Viewer we'll only need a fraction of that, but I think it could be another goal to add a manuscript metadata editor tool at some later point.

PeterBjuhr:
Nice to see that you have many development ideas about this!

One of the questions you're addressing is that about the connection between the manuscript and other documents. If you work with an external viewer you of course have to manually select the appropriate document. But with an internal viewer we can facilitate this. 

Without giving this too much thought I think it would be best if the documents in the manuscript viewer would be remembered on session level; in a session you shouldn't have to reopen any documents but be able to continue directly where you were last time you had the session open.

uliska:
Yes, I think this should be the default behaviour if no dedicated specification document is found. But *if* one is found the default behaviour should be to use that for determining which manuscript to load.

Maybe we should offer an option to create such an MEI file automatically upon manually opening a manuscript (set of ms). There could be a checkbox saying something like "Do you want to permanently associate this manuscript with the LilyPond document XX?" One could even launch the metadata editor from the open dialog (or have a checkbox that triggers the editor after opening the document).

PeterBjuhr:
About the storing of userdata in a MEI file, my spontaneous reaction is that is somewhat contrary to how Frescobaldi normally deals with user data. It would rather seem like an exchange format suitable for an export function. But there can be special considerations for this choice of course...

uliska:
Well, *one* core aspect of MEI *is* the encoding of source metadata. So I would say storing the properties of manuscripts is actually not user data but supporting an additional level of *content* data. the `.ly` file stores the musical content of a score while the `.mei` file stores the properties of the associated manuscript(s).

If we one day should have a way to see LilyPond and MEI files as parallel encodings of the score this could even be merged. Then the source metadata *is* actually part of the encoded content, and we could take the manuscript information from the document itself. And my suggestion is actually targeted at preparing this potential future step.

PeterBjuhr:
So what you are saying is that the data could be useful outside Frescobaldi!? That would strengthen my view that this really is more suitable for export than for internal use...

uliska:
Hm, what I'm saying is that the data is "sibling" data to the musical content. An `<mei>` XML document typically looks like this:

```xml
<mei>
  <meiHead>
  </meiHead>
  <music>
  </music>
</mei>
```

with the meiHead element containing information on the manuscript or other sources.

In a way you could consider writing such an MEI file for internal use *is* equivalent to "exporting" it.

But how I would prefer to see it is:
Frescobaldi gets a new field of work, namely editing metadata for the scores (manuscript) sources. Instead of designing a new file format or squeezing this information into the LilyPond files we make use of an existing technology, namely the MEI specification. This also makes sense because then Frescobaldi can be used as an editing environment for MEI projects, someday in the future.
Along the way the information in this file can be used for handling the properties of the manuscript viewer tool because the information needed for *display* (= internal use) are identical to those that are to be encoded for editorial purposes.

PeterBjuhr:
I don't yet fully grasp the implications of this for all levels of data storage, but at some levels I can definitely see the usefulness of this.

The third topic you introduced was the metadata editor. I think my main question here is what metadata should be read from the file (by ocr technique?) and what should be required by the user?

uliska:
Well, originally the only data that could be retrieved automatically is the file path and the number of pages.
An issue here is that the path will become part of the edition data, and thus relative/absolute paths will become an issue. (ah, and in addition to *files* we could also reference URLs for online resources).

What would be desirable for the purposes of the Viewer is information on the phyiscal location of measures. While one could one day try to incorporate OMR techniques (or call such programs) this will for now have to be edited manually. So we'd need a (modal) dialog for marking up the manuscripts.

I see two ways for that definition (I think the user should decide for the project):
- drawing boxes that define a measure
- defining horizontal split points for systems plus  
  vertical split points for measures within each system.

Any further metadata editing is something to be considered later. Apart from any editing capabilities we would have to investigate options to import from existing library databases.

PeterBjuhr:
One thought, if we somehow are able to read the content of the manuscript to some extent we should try to offer to translate this information to LilyPond source code.

PeterBjuhr:
> drawing boxes that define a measure

What do you mean by this more exactly?

uliska:
>> drawing boxes that define a measure

> What do you mean by this more exactly?

The common approach to this matter is to have each measure represented by a rectangle on the page. So in order to enter this encoding you'll have your image visible and pull up rectangles (like selections in drawing programs). The coordinates of that rectangle are then stored as a reference to the measure.

Later you can the "display m. 34", and the system knows which page and which area on that page to zoom into.

uliska:
> One thought, if we somehow are able to read the content of the manuscript to some extent we should try to offer to translate this information to LilyPond source code.

Definitely. But I think this is quite far from now. Actually this would mean to pull in a workflow into Frescobaldi that is usually done with multiple tools: OMR (Optical Music Recognition) into some format, then to MusicXML and then to LilyPond.

PeterBjuhr:
>>One thought, if we somehow are able to read the content of the manuscript to some extent we should try to offer to translate this information to LilyPond source code.

>Definitely. But I think this is quite far from now. Actually this would mean to pull in a workflow into Frescobaldi that is usually done with multiple tools: OMR (Optical Music Recognition) into some format, then to MusicXML and then to LilyPond.

I don't know the state of the art regarding these things, and I'm not sure this is a task for Frescobaldi. I'm just saying - if we are trying to read the content anyway, why not try to interpret it directly...

PaulMorris:
> On Jun 17, 2015, at 6:09 AM, Urs Liska <notifications@github.com> wrote:
> 
> with the meiHead element containing information on the manuscript or other sources.

> Instead of […] squeezing this information into the LilyPond files we make use of an existing technology, namely the MEI specification.

FWIW, I think it might be simpler to provide parallel data structures for use in LilyPond files.  They could follow the structure of the MEI meta data format 1 to 1.  Probably as scheme data in SXML format (which is less typing, arguably easier to read).  Like “MEI <meiHead> data in LilyPond clothes”.  Then it would be trivial to export that into an MEI file (once guile 2 arrives in LilyPond).

It sounds like a lot of work to keep two files in two different file formats always in sync, displaying both on the screen at the same time, and having to edit one or the other depending on what kind of editing you’re doing.  It’s simpler to have a canonical “master" file that you edit and then export to any other needed formats, working with one file at a time and with both capable of storing all the data.

But I’m speaking as an outsider looking in, Urs you know the most about this MEI area…

PaulMorris:
> On Jun 17, 2015, at 9:42 AM, Paul Morris <paul@paulwmorris.com> wrote:
> 
> It’s simpler to have a canonical “master" file that you edit and then export to any other needed formats, 

Well, I guess the idea is that the canonical file is the MEI file and LilyPond is just there to produce a graphic rendering.  It would still be interesting to consider embedding MEI meta data in LilyPond files in some way.

crantila:
You may be interested in *Lychee*, the "MEI Arbiter" I keep mentioning [elsewhere](http://lists.openlilylib.org/pipermail/mei2ly/2015-June/000042.html). Take a look at [this document](https://jameson.adjectivenoun.ca/crantila/lychee-sketch/blob/master/description.rst) to see if it does what you're hoping for. *Lychee* has bits for "canonical master files," automatic updating between different UI components, version control, and more!

The link above to "this document" uses a self-signed CA certificate for HTTPS. You may download the certificate (via HTTP...) from [here](http://ncodamusic.org/CA-20150228.cert).

PeterBjuhr:
I've just tested this in a real life situation and it actually works much better than I could expect. And much better than I could set up with an external viewer. (I placed the manuscript viewer above the ordinary music view and (with auto-engrave) I could instantly compare the progress.)

I think we should postpone the discussion about advanced features that could be added later, add the basic controls etc and merge it to master.

uliska:
Ok.
What are basic controls? I'd say:

- open dialog
- zoom
- page navigation

Is that right?

Is it possible to make that layout a default? It's the first thing I did too: split the music view horizontally to have the manuscript on top and the Music View on the bottom of the main window's right half.

PeterBjuhr:
>What are basic controls? I'd say:

>- open dialog
>- zoom
>- page navigation

>Is that right?

I think that covers the basic needs!

>Is it possible to make that layout a default? 

Do you think the user base is so large? Many other tools are not default (but the custom layout is remembered).

>It's the first thing I did too: split the music view horizontally to have the manuscript on top and the Music View on the bottom of the main window's right half.

I just realized that with this placement the controls for the Music view have to be moved because they are always at the top.

uliska:
I'm closing this pull request for now. As said above @PeterBjuhr and I will work on that branch to create a working first implementation, and when we think we're ready we'll open a new pull request.

PR #540, opened 2014-12-28
Check if midi file exists before running Timidity, hopefully closes #530

### LINKED ISSUE #530: „File → Export → Export Audio…“ can create empty file
The title says it: If there’s no midi file to convert, frescobaldi creates an empty .wav file – it should better just notify that the midi file is missing and exit.

PeterBjuhr:


PeterBjuhr:
Now I've changed it to using the first of the resultfiles, see #551.

In the future perhaps we could have a dialog with the resultfiles as a drop-down.

PR #552, opened 2015-01-03
Initial stub at a document statistics dialog

uliska:
This is not intended to be merged, but please tell me if I did it right with the integration of the new dialog (it seems I got somewhat out of touch with the architecture ...)

### DIFF
--- a/frescobaldi_app/menu.py
+++ b/frescobaldi_app/menu.py
@@ -296,6 +296,10 @@ def menu_tools(mainwindow):
     m.addAction(mainwindow.actionCollection.file_open_current_directory)
     m.addAction(mainwindow.actionCollection.file_open_command_prompt)
     m.addSeparator()
+    if vcs.app_is_git_controlled() or QSettings().value("experimental-features", False, bool):
+        import statistics
+        m.addAction(statistics.DocumentStatistics.instance(mainwindow).actionCollection.docstatistics)
+    m.addSeparator()
     panelmanager.manager(mainwindow).addActionsToMenu(m)
     return m
 
--- a/frescobaldi_app/panelmanager.py
+++ b/frescobaldi_app/panelmanager.py
@@ -60,7 +60,7 @@ class PanelManager(plugin.MainWindowPlugin):
         self.loadPanel("doclist.DocumentList")
         self.loadPanel("outline.OutlinePanel")
         self.loadPanel("layoutcontrol.LayoutControlOptions")
-        
+
         # The Object editor is highly experimental and should be
         # commented out for stable releases.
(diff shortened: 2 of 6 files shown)


wbsoft:
I think it is good. It is a nice help to debug and explore the output of the music tree. I probably would not assign ctrl+shift+d by default to this dialog. The dialog does not need Cancel/Ok buttons but just a Close button. Maybe 'Run' can be changed to 'Update'. More tabs could be added later to display other statistical properties, such as number of notes, etc.

On a side note: I want to create a panel displaying the music tree as an expandable tree, where items have properties, and that moving the text cursor also shows clearly which item we are in. (And clicking an item moves the text cursor to it.) But before that I want to finish the design of the `ly.music` tree stuff.

uliska:
> I think it is good.
>

OK, then I'll close the pull request ;-)

> It is a nice help to debug and explore the output of the music tree.
>

Hm, the current *behaviour* (i.e. dumping the music tree) is just a stub 
for much more advanced functionality I have in mind. The pull request is 
mainly for the integration of the dialog in the system.
Basically this will become a document statistics dialog like you have in 
word processors, counting (and grouping) all sorts of items in a given 
score.

A more advanced idea I have is the possibility to use presets that match 
item types to values so you can use the document statistics to 
automatically calculate a "value" of a score. That will make it possible 
to devise schemes to account by the actual content of a score instead of 
rather unprecise approaches by page or hour.

> I probably would not assign ctrl+shift+d by default to this dialog.
>

Would you suggest a different shortcut or no default at all?

> The dialog does not need Cancel/Ok buttons but just a Close button. 
> Maybe 'Run' can be changed to 'Update'.
>

I will think about this more closely as soon as I implement real 
functionality.

> More tabs could be added later to display other statistical 
> properties, such as number of notes, etc.
>

See above. The actual functionality and design have not been started so far.

> On a side note: I want to create a panel displaying the music tree as 
> an expandable tree, where items have properties, and that moving the 
> text cursor also shows clearly which item we are in. (And clicking an 
> item moves the text cursor to it.) But before that I want to finish 
> the design of the |ly.music| tree stuff.
>

Sounds very nice. Maybe you'll even find a way to use that for some 
kinds of drag & drop editing. E.g. moving music between voices or staves 
or whatever.

But your above comment made me think of yet another tool: a panel that 
displays the output of \displayMusic for a given music. This would be 
helpful when developing Scheme functions (and learning how to do that). 
It could be accessed similar to "Edit in place", using a selection. If 
it is a panel it could also be configured to automatically update with 
the music element the cursor currently is at.

wbsoft:
Yes a displayMusic tool is also in my todo-list :-) added to TODO

Issue #555, opened 2015-01-05
Release 2.17.2


wbsoft:
Dear friends, I want to release 2.17.2 next week (jan 16th) as a bugfix release, with the fix for issue #543. Can you help test the v2.17.x branch (and maybe cherry-pick some little things to that branch?).

dliessi:
Maybe we could cherry-pick the French translation update dc028dccb98334d23f0dea689c7832e285567f01.
What about 220ed698a537b7d08fa389f68601bfd667aa94b5 (fix for #535) and 2bb9b2f94569d2c89fbfa9e79530c12bab55b070 (partial fix for #541)?
Also, I'd like to cherry-pick a62a4ef4c2a7c97536e220174215e63daa330800 and 0483639590c83ed370f333ec888271a00805b876, that would avoid the need to patch the source in MacPorts.

If you approve I can take care of cherry-picking these.

dliessi:
I opened a pull request for my suggested cherry-picks: feel free to discard it if you don't want some of them.

PeterBjuhr:
I've also added some cherry-picks. Bug-fixes and additions to the userguide that I think could be included...

dliessi:
@wbsoft I'm sorry I don't think I'll be able to upload the new Mac packages today.
I hope tomorrow.

dliessi:
@wbsoft Upload complete, you can announce. :)
MacPorts update will follow.

PR #556, opened 2015-01-05
Cherry-picks for v2.17.2

dliessi:
The cherry-picks I suggested in #555 and an update to the Italian translation.

Dutch, French and Chinese translations need only a few updates, the other translations need a thorough revision.


wbsoft:
thanks!

PR #557, opened 2015-01-07
Python3 stable

proski:
Most of my Python 3 patches are not needed anymore. The remaining four patches are still needed.

### DIFF
--- a/export-help.py
+++ b/export-help.py
@@ -13,6 +13,7 @@ It creates a help/ directory (by default) and puts the HTML and images there.
 """
 
 from __future__ import unicode_literals
+from __future__ import print_function
 
 import os
 import re
@@ -240,7 +241,7 @@ def make_page(lang):
         page_address = footer)
     
     filename = os.path.join(output_dir, make_filename('uguide.html', lang))
-    print 'Exporting to:', filename
+    print('Exporting to:', filename)
     with open(filename, 'w') as f:
         f.write(html.encode('utf-8'))
 
--- a/frescobaldi_app/fileinfo.py
+++ b/frescobaldi_app/fileinfo.py
@@ -61,7 +61,7 @@ def _cached(filename):
     try:
         c = _document_cache[filename]
     except KeyError:
-        with open(filename) as f:
+        with open(filename, 'rb') as f:
             text = util.decode(f.read())
         c = _document_cache[filename] = _CachedDocument()
         c.variables = v = variables.variables(text)
--- a/frescobaldi_app/ly/lex/__init__.py
+++ b/frescobaldi_app/ly/lex/__init__.py
@@ -41,7 +41,7 @@ Usage:
 >>> txt = r"\relative c' { c d e f-^ g }"
 >>> s = ly.lex.state("lilypond")
 >>> for t in s.tokens(txt):
-...     print t, t.__class__.__name__
+...     print(t, t.__class__.__name__)
 
 \relative Command
   Space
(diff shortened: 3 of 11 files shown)


wbsoft:
Thank you! I wasn't aware of your patch set. Very welcome!

proski:
​Hi Wilbert,

Sorry, I was sitting on those patches too long without submitting them. On
the other hand, most of your fixed are better than mine. Some of my fixed
are not needed anymore and I cannot understand why. For example, adding
extra arguments to some GUI functions is not needed now:

https://github.com/proski/frescobaldi/commit/436fb829ad01ec11e58c8c580c10ab33b5264f25

The only thing left is complete elimination of basestring. I think the
simplest approach would be to do

isinstance(args, (type(b''), type(u'')))

Also "if sys.version_info[0] == 3" is not good, it will break in Python 4.

Setting "string_types = str" on Python 3 is questionable, it excludes
bytes. Maybe they were supposed to be excluded, but I'd rather see that
logic with a comment where it's applied, not globally on top of the file.

I cannot get PDF to show with Python 3, but it may be a Fedora 21 issue
(poppler compiled for python2 or something like that).

Pavel

wbsoft:
the poppler-qt4 module works with Python3, it just needs to be compiled.

The extra argument issue was because of a bug in signals.py, I fixed couple of days ago.

Indeed the `(type(b''), type(u''))` is a good idea.

And the best version check is `if sys.version_info >= (3, 0):` etc.

One remaining thing is that I can't yet get portmidi/pypm working.

dliessi:
> The only thing left is complete elimination of basestring. I think the simplest approach would be to do
>
> isinstance(args, (type(b''), type(u'')))
>
> Also "if sys.version_info[0] == 3" is not good, it will break in Python 4.
>
> Setting "string_types = str" on Python 3 is questionable, it excludes bytes. Maybe they were supposed to be excluded, but I'd rather see that logic with a comment where it's applied, not globally on top of the file.

If I understand correctly, your comments refer to edc15de8269f34461c21cc1860ead2af3022edfd.
As far as I understand, byte strings *were* meant to be excluded, at least according to https://docs.python.org/3/howto/pyporting.html, which explicitly advises *against* mixing binary data (`b''`) and text (`u''`).

The way I defined `string_types` is exactly how it is defined in [six](https://pypi.python.org/pypi/six), see [its source](https://bitbucket.org/gutworth/six/src/2c4cfe488c5bb03a6f7315025e62242730eeca89/six.py#cl-35).
I preferred to write those few lines instead of adding a dependency on a whole package.

I can change `==` to `>=` in the `if`, and I can add a comment explaining why I wrote that code, but I don't think that I should use `isinstance(args, (type(b''), type(u'')))`.

wbsoft:
Agreed!

proski:
OK, I get it now. In Python 3, every string is unicode. In Python 2, str("") is a non-unicode string (unless str was redefined), whereas "" may be unicode if "from __future__ import unicode_literals" is used. So we want to match both in Python 2, but only Unicode in Python 3. Thank you for the link, it's very useful.

proski:
By the way, frescobaldi under python 3 cannot open files with umlauts. QApplication is instantiated with correct sys.argv in app.py, but QApplication.arguments() in main.py returns arguments with umlauts stripped away. I don't see any locale settings for QApplication. Locale is en_US.UTF-8. Observed on Ubuntu 14.10 and Fedora 21.

wbsoft:
thanks for catching that, i created an issue for that: #562.

PR #558, opened 2015-01-07
Fix typo in mode shift dialog

uliska:
The key has to be lower case, so the dialog should propose it right.
And I think "e.g." is more appropriate than "i.e." here.


PeterBjuhr:
Thanks! But maybe it should accept upper-case note names??

uliska:
Well,being more oermissive might be good. But liwer case is LilyPond syntax,so I'd keep it.

PeterBjuhr:
OK, we keep it as it is!

PR #559, opened 2015-01-07
Extend mode-shift to handle also non-heptatonic modes

PeterBjuhr:


PeterBjuhr:
A heptatonic scale is of course a scale that has 7 notes per octave... :)

PR #565, opened 2015-01-08
Another import error in Python 3

PeterBjuhr:


frescobaldi_app/widgets/shortcuteditdialog.py
```diff
@@ -31,7 +31,7 @@
 
 import app
 from . import Separator
-from keysequencewidget import KeySequenceWidget
+from . import keysequencewidget
```
wbsoft: Does this work? Now we don't have the global `KeySequenceWidget`. You could also do:
```
from .keysequencewidget import KeySequenceWidget
```

wbsoft:
I fixed this in bd54989923f1655be74909b76694117f6963d152

frescobaldi_app/widgets/shortcuteditdialog.py
PeterBjuhr: Ah, sorry!

Issue #566, opened 2015-01-09
on svg view close, "Script error." is printed to the console


wbsoft:
When a svg view is closed (e.g. when closing a document that has a svg output document in the svg viewer), on the console
```
Script error.
```
is printed. Probably an issue in the javascript that runs inside the SVG view.

PeterBjuhr:
I've traced the error to `svgview.view.clear` and the line:

    self.load(QtCore.QUrl())

Exactly why this triggers the error is not totally clear to me, but it seems to have to do with the browser's same origin policy (see http://en.wikipedia.org/wiki/Same-origin_policy).

If the view could be cleared in another way I think it would be preferable.

wbsoft:
I can't find another way than `QWebView.setUrl(QUrl())` which almost acts like `load()`, but will empty the view immediately instead of keeping old content visible while loading the new content. But `setUrl()` also causes `Script error.` being printed to the console.

PeterBjuhr:
The printout of JavaScript errors in the console is really only for debugging purposes. In the case this error can't be helped I think I should just suppress the errors. But I'll take a closer look first!

PeterBjuhr:
Although I find it kind of ugly we could easily hide this seemingly irrelevant error message.

Another idea which I think is quite nice is to have a default svg-file that is shown when the viewer is empty. It could be the Frescobaldi logo and some text.

What do you think?

wbsoft:
Hiding the message is not needed I think.

I would make the empty view simply grey, like the music view is. Maybe with a very dimly colored frescobaldi "watermark" at the background.

PeterBjuhr:
I'm uncertain of how to reference the file, but I'll give it a go!

PeterBjuhr:
Is the logo dimly enough or is it too apparent?

PeterBjuhr:
Everything seem to be cleared.

PR #568, opened 2015-01-10
Loading a default background instead of an empty QUrl, see issue #566

PeterBjuhr:
It looks nicer and we won't get the Script error.


wbsoft:
Nice. it would be nice if the lily would auto-center and not show scrollbars (unless the lily itself would be bigger than the view). But I don't know if that 'd be possible :-)

PeterBjuhr:
Yes, I agree. I don't know if it's possible either. But what I could do is to not make the image bigger than the lily.

PeterBjuhr:
The image is in 'A4' size now as it is LilyPond's default. 

I guess it would be possible to read the settings of the widget and to adjust the image accordingly.

A much simpler solution would be to remove the lily and only have the grey background (as the Music View). :)

wbsoft:
Try loading this HTML:
```
<html>
<head>
<style type="text/css">
html { 
  background:
  -webkit-linear-gradient(top, rgba(160,160,160,0.7) 0%,rgba(160,160,160,0.7) 100%),
  url(../icons/frescobaldi.svg) no-repeat center center fixed ; 
  -webkit-background-size: cover;
  background-size: cover;
}
</style>
</head>
</html>
```
It can point to the default Frescobaldi icon which will be shown partially transparent (the 0.7; 1.0 is completely transparent).
If you comment the both `background-size` lines, the image will be centered but not streched.

PeterBjuhr:
Thanks! It's much better. The scrollbars are gone. But as I have a rather narrow screen the lily is cropped horizontally.

PeterBjuhr:
Also I get an error because the file isn't an svg.

wbsoft:
We should change `cover` to `contain`...

PeterBjuhr:
Yes, it's better! But I think it's even better to set a percentage value so there is a margin.

I solved the svg error by adding a `svg` tag to the document.

wbsoft:
I think the svg tag complaint comes from the script that's started in svgLoaded after loading the HTML. We should probably not connect to the loaded signal but manually call the javascripts after loading a svg document.

wbsoft:
Sorry, my last comment will be mistaken, load or setUrl return directly, we can install javascripts only when the loading has finished, hence the signal is ok!

PeterBjuhr:
@wbsoft I see you have fixed the loose ends now. Great!

PR #572, opened 2015-01-14
Eliminating unichr

proski:
Python 3 doesn't understand `unichr` and `unicode`. I tend to use the approach where we code for Python 3 and make Python 2 "thoroughly-unicodized" using try-except construct. Thanks to separate namespaces, such tricks don't affect other modules.

The output of hyphenator has be greatly improved. It can deal with non-latin encodings on the command line.

python3 hyphenator.py /home/proskin/src/frescobaldi/frescobaldi_app/hyphdicts/hyph_ru_RU.dic попарнопротивопоставленные
попарнопротивопоставленны – е
попарнопротивопоставлен – ные
попарнопротивопостав – ленные
попарнопротивопо – ставленные
попарнопротиво – поставленные
попарнопроти – вопоставленные
попарнопро – тивопоставленные
попарно – противопоставленные
попар – нопротивопоставленные
по – парнопротивопоставленные

python3 hyphenator.py /home/proskin/src/frescobaldi/frescobaldi_app/hyphdicts/hyph_sv_SE.dic utställningsföremål
utställningsföre – mål
utställnings – föremål
utställ – ningsföremål
ut – ställningsföremål

### DIFF
--- a/macosx/mac-app.py
+++ b/macosx/mac-app.py
@@ -20,10 +20,10 @@ from subprocess import Popen
 # See https://docs.python.org/3/howto/pyporting.html for details.
 # The following code is the same used in package six to define a
 # version independent string type for isinstance() tests.
-if sys.version_info[0] >= 3:
-    string_types = str
-else:
+try:
     string_types = basestring
+except NameError:
+    string_types = str
 
 macosx = os.path.realpath(os.path.dirname(__file__))
 root = os.path.dirname(macosx)
--- a/frescobaldi_app/autocomplete/documentdata.py
+++ b/frescobaldi_app/autocomplete/documentdata.py
@@ -23,6 +23,11 @@ Completions data harvested from a Document.
 
 from __future__ import unicode_literals
 
+try:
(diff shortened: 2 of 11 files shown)


proski:
More patches added. Hyphenation is actually working in Frescobaldi now under Python 3. I used the same try-except approach to deal with all occurrences of `unicode`. Also, basestring is handled that way. In some cases, I replaced fixes made by others to make the code more uniform. Non-unicode strings are re-allowed under Python 2 in dom.ly as it's the documented behavior. The function is called `ifbasestring` for a reason.

macosx/mac-app.py
```diff
@@ -20,10 +20,10 @@
 # See https://docs.python.org/3/howto/pyporting.html for details.
 # The following code is the same used in package six to define a
```
dliessi: Please remove also lines 21 and 22 (the last two lines of the comment) since they are no longer related to the code.

wbsoft:
Thanks!

PR #583, opened 2015-01-19
Plural forms working in Python 3

proski:
We have two places where all exceptions related to language files are silently ignored. That's wrong. The robustness principle says: "Be conservative in what you do, be liberal in what you accept from others". Language files is what we do, even if translations are done by third parties.

Removing those exception handlers immediately exposed problems in *.mo header handling in Python 3. They have been fixed, including functions called by molint.py. The later was also failing to close *.mo files.

The visible result is that plural forms work with Russian translation under Python 3 now (at the moment, they are only used when exporting snippets).

Tabs used for code indentation cause exceptions in Python 3. I've replaced them with spaces using formatting suggested by autopep8.

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -110,7 +110,7 @@ copy_plugins('iconengines')
 
 # copy the frescobaldi_app directory
 subprocess.call([sys.executable, 'setup.py', 'build_py',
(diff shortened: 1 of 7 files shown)


wbsoft:
Thanks for you improvements!

PR #586, opened 2015-01-20
Python 3 issues with MusicXML import/export

proski:
Both commits involve changing files from text to binary mode. Text mode is may be convenient to read ASCII-only output. It may be a good choice for the code that only runs in Python 2 or Python 3. But if we want to support both versions of Python and deal with non-ASCII data (such as a non-latin filename in error output). Text files deal with different types on Python 2 and Python 3, even though both are called "str". Python 3 does the encoding/decoding for us, Python 2 doesn't. That's too much trouble and too little control over the encoding.

Another case of unclosed file has been caught. Python 3 is good at catching such errors.


PeterBjuhr:
Thanks for the contribution!

PR #588, opened 2015-01-20
Windows fixes and git robustness

proski:
I tried Frescobaldi on Windows under ActivePython 2.7.8.10. There were two fatal problems: pypm.Initialize() was missing and there was no "remote" section in .git/config - git for Windows forgot to create it. It turns out "import pypm" imports "PyPM", an ActiveState package manager. I used a simple hasattr() trick to reject it.

As for git, I was tired of errors due to .git/config, so I completely eliminated parsing that file. "git config" provides all that functionality and it's documented. While at that, I fixed running git in shell and disallowed passing strings as arguments - only lists are allowed now. Finally, STGit branches are fully hidden from view now. They don't contain the source tree; checking them out would remove the sources.

### DIFF
--- a/frescobaldi_app/portmidi/__init__.py
+++ b/frescobaldi_app/portmidi/__init__.py
@@ -299,6 +299,10 @@ def _do_import_pypm():
     
     """
     import pypm
+    # Reject incompatible API, such as ActiveState PyPM
+    if not hasattr(pypm, "CountDevices"):
+        del pypm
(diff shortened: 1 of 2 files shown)


wbsoft:
All the git code was added by @uliska , maybe he can help to make it better. I will merge this, because the changes look very good to me.

uliska:
I agree that this looks good (apart from 552b78b, which I can't judge).
Thank you for improving my code.

PR #590, opened 2015-01-21
Git changes

proski:
Somehow I missed some code still relying on parsed .git/config file. That code is not used, so I'm removing it. I also changed the way branches() works. Indicating the current branch with an asterisk is good for text user interface, but not for API that is supposed to work with many VCSes.

It would be nice to wait until Urs has a look.

### DIFF
--- a/freeze.py
+++ b/freeze.py
@@ -29,7 +29,7 @@ import sys
 from cx_Freeze import Executable, Freezer
 
 # access meta-information such as version, etc.
-from frescobaldi_app import info
+from frescobaldi_app import appinfo
 
 # find pypm by adding the dir of pygame to sys.path
 sys.path.append(imp.find_module('pygame')[1])
@@ -160,10 +160,10 @@ Filename: "{{app}}\\frescobaldi.exe";\
  Flags: postinstall nowait skipifsilent;
 
 '''.format(
-    version=info.version,
-    homepage=info.url,
-    author=info.maintainer,
-    comments=info.description,
+    version=appinfo.version,
+    homepage=appinfo.url,
+    author=appinfo.maintainer,
+    comments=appinfo.description,
     target=target_dir,
 )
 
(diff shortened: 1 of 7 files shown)


uliska:
Sorry, can't look into it more closely right now. Just skimmed it and must say I can't immediately judge.

I had intended to look into that area again for quite some time but didn't manage to do so. Maybe you can keep in mind two ideas/tasks I have in mind:

- I started with the code in a way that is somewhat too focused on Frescobaldi's own Git repository. But it actually is intended as a hotbed for continuing in the direction of a generic interface to also manage *project* repositories. I don't know if that has concrete implications on your current work, but you might just keep it in mind.
- Now that there are more and more forks around I thought of adding a function (this should not be done automatically on startup but only on explicit request) that retrieves all forks and checks which of them have branches with differences - and only those branches are shown in the `Git` menu.  
The menu could then be organized by forks (either with simple separators or with submenus) so you can always see who is working on what. I think this could simplify and encourage exchange and testing between different developers.  
Another function could then be added to check for pull requests (which may be done on startup), which would then be highlighted in some way. With that every user running Frescobaldi from Git would be notified about pull requests and would be encouraged to test them.  
All information about forks, branches, pull requests etc. can be retrieved as JSON files through the Github API, without needing any Git operations.

proski:
The removal of upstream goes in the direction of managing any repositories. The upstream is only known for Frescobaldi, but not for an arbitrary project (until Github API is used). The removal of "*" marker from the public API makes sense if VCSes other than git are supported. It also eliminates the need for the callers to do string parsing.

I'm not interested in adding new VCS functionality. I think it doesn't belong to Frescobaldi. An IDE would be a better place to develop such code. If it becomes mature and modular, it could be used by Frescobaldi.

I just want to fix the code that breaks Frescobaldi from running or sets a bad example.

proski:
Sorry, I killed this request inadvertently by using "git push --force" in my repository. Since there were no objections, I've committed the patches to the upstream.

PR #591, opened 2015-01-22
More changes needed after renaming info.py to appinfo.py

proski:
Renaming modules is hard, but this one was needed. There are many "info" variables around the source, it was confusing to have a module of the same name.

Some references to info.py were left after your patch. Some were found with regexes, some were found by pylint, some were found accidentally. I hope I didn't miss anything.


wbsoft:
Thank you :-)

Issue #592, opened 2015-01-23
proposed project restructuring


wbsoft:
I am very happy that now multiple codevelopers help bringing Frescobaldi forward. Also it is desired that python-ly becomes visible and is further developed on its own. To reflect this, and to make some development paths easier, I propopse some changes to how Frescobaldi is structured:

* make `python-ly` a separate package and have Frescobaldi depend on it (can be done as of 2.18)
* (later) make `qpopplerview` a separate package and have Frescobaldi depend on it
* (much later, e.g. next year, or when Frescobaldi supports Qt5 and reaches version 3.0) make a GitHub organisation (named "frescobaldi") and add python-ly and frescobaldi to that organisation.

What do co-developers think?

uliska:
That makes very much sense to me.

While we're at it I would like to add one more thought to that (which I've already expressed earlier but which seems to be related here). I would really like to have one more entity, namely a `frescobaldi-lib` with LilyPond code. For example the code in `layoutcontrol` should be moved there but we could add more material to it.  
This would make it possible to add features to Frescobaldi that rely on LilyPond code being present - without making documents depend on being compiled with Frescobaldi (if someone wants to share documents with non-Frescobaldi users they can be directed to get that library).

I would really like to see such a thing, for example to add support for the `ScholarLY` package that I have just created (and will present on Scores of Beauty soon), in particular I'd like to have an annotation editor in Frescobaldi.

wbsoft:
@uliska, isn't it an idea to add such lilypond code to a library in the openlilylib project. If it is LilyPond code that should also be available to users of other editors, it is not needed (I think) to call that library frescobaldi-lib, as that suggests one can't use that library outside Frescobaldi.

We should give that library of LIlyPond code a generic name and we could add features that use that library to Frescobaldi, and also add a feature to Frescobaldi to automatically install or update the library (just like I want Frescobaldi to be able to install LilyPond and its documentation transparently in multiple versions). Frescobaldi should be able to work perfectly without that library, but some features would be greyed out and an easy way to install the library should be there.

As far as I can see, openlilylib could be a good place for such a library. Please let me know what you think.

uliska:
> @uliska <https://github.com/uliska>, isn't it an idea to add such 
> lilypond code to a library in the openlilylib project.
>

Hm, maybe. I've thought of this too, but see below.

> If it is LilyPond code that should also be available to users of other 
> editors, it is not needed (I think) to call that library 
[...]

PR #595, opened 2015-01-25
Adding assumed path to python-ly as mentioned in issue #592

PeterBjuhr:
If we can assume a setup where python-ly and frescobaldi are next to each other in a directory, this will facilitate debugging latest version of both packages.

### DIFF
--- a/frescobaldi_app/debug.py
+++ b/frescobaldi_app/debug.py
@@ -60,6 +60,11 @@ def modules():
 
 
 
+# include python-ly if possible
+import os
+lypath = os.path.realpath('../python-ly')
+if os.path.isdir(lypath):
+    sys.path.insert(1, lypath)
 
 # avoid builtins._ being overwritten
 sys.displayhook = app.displayhook


uliska:
I think we *can* assume that and explain it like this for developers.  But what happens if python-ly is not there but installed through the package manager?

PeterBjuhr:
Then the installed version will be used. The path is only added if it is present.

proski:
The fix is not working for me:

  File "./frescobaldi", line 11, in <module>
    main.main()                     # Parse command line, create windows etc
  File "/mnt/fedora/home/proski/src/frescobaldi/frescobaldi_app/main.py", line 135, in main
    import mainwindow       # contains MainWindow class
  File "/mnt/fedora/home/proski/src/frescobaldi/frescobaldi_app/mainwindow.py", line 45, in <module>
    import menu
  File "/mnt/fedora/home/proski/src/frescobaldi/frescobaldi_app/menu.py", line 37, in <module>
    import documentactions
  File "/mnt/fedora/home/proski/src/frescobaldi/frescobaldi_app/documentactions.py", line 31, in <module>
    import highlighter
  File "/mnt/fedora/home/proski/src/frescobaldi/frescobaldi_app/highlighter.py", line 33, in <module>
    import ly.lex
ImportError: No module named ly.lex

The sys.path adjustment should happen earlier. Also, why debug.py? It's not like using ../python-ly is only used for debugging.

I think the installed python-ly should take preference. The python-ly directory may contain something not suitable for frescobaldi, e.g. an old version of python-ly. We cannot assume that all frescobaldi users who have the python-ly directory are trying out the code in that directory.

Please see this patch, it fixes that traceback for me.

https://github.com/proski/frescobaldi/commit/08779d3044d78effac5022d5da33df4e39ed23d1

I also think that adding ../python-ly to sys.path should only be done is frescobaldi is run from the source directory. I'll try to implement that.

uliska:
Am 26.01.2015 um 08:05 schrieb Pavel Roskin:
> I also think that adding ../python-ly to sys.path should only be done 
> is frescobaldi is run from the source directory. I'll try to implement 
> that.

We have vcs.app_is_git_controlled() which should make that simple

proski:
Good point. I got an answer in another thread that debug.py is not loaded automatically, tso the fix wasn't supposed to work for me.

I'm on the fence whether to leave the code it as is or implement python-ly detection under some narrow circumstances.

PeterBjuhr:
>I got an answer in another thread that debug.py is not loaded automatically, so the fix wasn't supposed to work for me.

@proski so the error above was not from running debug.py at all?

wbsoft:
@PeterBjuhr yes I wrote that as a comment below 08779d3044d78effac5022d5da33df4e39ed23d1 .
I am reluctant to test in `main.py` for the existence of source tree of python-ly because then that happens on every Frescobaldi start.

We could devise a debug script instead of the `frescobaldi` script, that is only for developing and testing. That one could explicitly try to import `ly` from a source tree and also print various console messages (i.e. enable some kind of debug mode.

But I do not really think that is needed, because we already have the possibility of `from frescobaldi_app import debug.py` as mentioned in README-development.

PeterBjuhr:
I totally agree, that script shouldn't be moved to `main.py`. It was only meant for the special case of starting the app through the debug script. And even then maybe it's assuming to much.

PeterBjuhr:
It seems the answer may be redundant at this point, but I'll keep it short:

>Also, why debug.py?

It seems to me that when running the app through the debug script using the PYTHONPATH construct would be the normal case and then it would be more convenient to add it to the script instead...

proski:
@PeterBjuhr Correct, I wasn't loading debug.py.

I think we are overengineering the code and adding unexpected behavior. Why would I want to use a non-installed version of python-ly if and only if debug.py is loaded? Python is capable of reporting stacktraces of the installed code. Switching to different code when debugging is an unexpected behavior that would complicate debugging. Have you heard of Heisenbug? :)

I thought the intention of the change was to run frescobaldi without installing anything, python-ly of frescobaldi itself. It looks like everybody has a different idea what should happen. In that case, the best solution would be to stick with the default behavior. It may be non quite convenient, but at least it's consistent.

PeterBjuhr:
@proski sorry for the misunderstanding about what the change was suppose to do!

What you are saying is really that we shouldn't use debug.py at all!? Or do I take the interpretation too far?

proski:
I'm saying we shouldn't be adding python-ly to sys.path in debug.py. It's not the expected behavior.

There is a huge downside. If the version of python-ly in the source directory is different from the installed version of python-ly, loading debug.py changes the behavior of frescobaldi.

We have no reason to assume that the source version of python-ly is the same as the installed version plus possible debug statements.

wbsoft:
I actually agree with  @proski , we should simply import ly. A user who wants to test Frescobaldi with a different version of `ly` (or any other external package, e.g. pypm or poppler-qt4) can use the PYTHONPATH variable to handle that. Frescobaldi should not do such things by itself.

PeterBjuhr:
@wbsoft agreed, I have several times expressed reluctance for this commit for the reasons you mention. But it was interesting to create the pull request for the sake of the discussion. I think I should create my own debug script as was my inital intent.

proski:
I've reverted the patch in d41cff632c173d27024a1527657f70b738c5a76f

PeterBjuhr:
Thanks!

I'm sorry, I should have been more explicit that the pull request needed more careful review as @uliska have done on several occasions!

wbsoft:
it's ok!

PeterBjuhr:
I'm beginning to set things up nicely locally. @wbsoft I hope you don't mind a question; why do you import PyQt stuff in the debug script.

(I'll also take @proski 's advice and not overuse running from debug scripts...)

wbsoft:
To make it easier to type stuff at the promt, use Qt objects etc.

PeterBjuhr:
I see! Thanks.

PR #600, opened 2015-01-28
Reimplement "Export Audio" using a dialog

proski:
I tested the code carefully, but I don't do much GUI programming, so I'm submitting this patch as a pull request.

As few stylistic questions about the code.

Would it be better to have a separate file for the audio export dialog? Or is the code too trivial for that?

Is it OK to do so many things in the constructor? Or should I move some of that stuff to another function? If so, should it be a slot or just a class method?

Any better ideas what to write in self.setMessage()?

### DIFF
--- a/frescobaldi_app/engrave/lytools.py
+++ b/frescobaldi_app/engrave/lytools.py
@@ -37,7 +37,7 @@ import widgets.dialog
 def show_available_fonts(mainwin, info):
     """Display a dialog with the available fonts of LilyPond specified by info."""
     dlg = Dialog(mainwin)
-    dlg.setWindowTitle(app.caption("Available Fonts"))
+    dlg.setWindowTitle(app.caption(_("Available Fonts")))
     dlg.run_command(info, ['-dshow-available-fonts'], _("Available Fonts"))
     dlg.setMessage(_(
         "List of fonts detected by {version}").format(version=info.prettyName()))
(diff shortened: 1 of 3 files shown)


wbsoft:
Thanks, all changes look good so I merged them.

wbsoft:
Regarding the audio exporting dialog: This dialog is basically about running any external command and examining its output. I'll probably move it to a generic dialog and have the caller set the command etc.

proski:
I agree, it can be generalized. But please note that the code includes logic specific to file generation. The output file is deleted if the operation is canceled.

Why did you add another call to setAttribute(Qt.WA_DeleteOnClose)? It's already in AudioExportDialog.__init__. Besides, isn't it too late to set it after dlg.show()?

wbsoft:
>Why did you add another call to setAttribute(Qt.WA_DeleteOnClose)? It's already in AudioExportDialog.init. Besides, isn't it too late to set it after dlg.show()?

Sorry, I didn't notice. I'll keep the one in the caller. I'd rather not have in in `__init__`, init just constructs the thing, and only the caller decides to use it only once and then throw it away. I will change it a little to demonstrate what I mean.

proski:
That would be great. I was feeling uneasy about stuffing all code into the constructor, but I didn't know what belongs there and what doesn't. I'll be happy to check your commits and learn.

wbsoft:
I tried to make a generic class and inherit it in `file_export/__init_.py`. See what you think of it. I also learn a lot of your coding! :-)

proski:
Very nice!

PR #617, opened 2015-02-16
Adding abc2ly and midi2ly imports

PeterBjuhr:
The code is restructured with a generic import dialog, and abc2ly and midi2ly imports are added.

Some issues remaining:

1. abc2ly doesn't work in the same way as musicxml2ly and midi2ly.
2. The previous dialog has to be "forgotten" when switching between different imports.

I've indicated the problematic places in the code.


wbsoft:
> When the previous import dialog is remembered (self._importDialog) switching between imports is impossible. Should a new dialog simply be created in each occasion?

Yes, probably per import file type(?) so some user preferences are remembered. I'll merge and we can improve.

PR #624, opened 2015-02-19
midiinput: fixed byte representation of midi event

mamoch:
I didn’t test the midiinput in frescobaldi on python3, but I checked that python2 and python3 return the same result if you feed them `bytearray([...])`, and they do.


proski:
Indeed, it turns out we had some Python3 specific code in Frescobaldi. The patch fixes MIDI capture in Python 2. Thank you!

Just two minor points. The patch adds trailing whitespace. We don't need that. We have more that enough of that in the code.

The new comment added to the code is not useful there. It belongs to the patch description only. It's not practical to comment (in the code) why every function was chosen. Developers reading the code in the future would probably be interested in some other issue than Python 2/3 compatibility. Nobody would suddenly feel the urge to change bytearray back to bytes :)

Maybe you could remove that comment and I would merge your patch?

mamoch:
You’re welcome. :-) But I also needed that feature working. That’s why I added `midiinput` in the first place. ;-)

And the trailing space was in that unnecessary comment, so it’s gone now, too.

proski:
Committed. Thank you!

wbsoft:
Thanks!

PR #627, opened 2015-02-24
Implementing rest tools from ly.rests as Frescobaldi tools

PeterBjuhr:
Please take a look at all text in the app so everything look comprehensible and ok!


wbsoft:
Thanks!

PR #658, opened 2015-03-14
Change all Python '#!' lines to '#!/usr/bin/env python' (fix #640)

### LINKED ISSUE #640: Change/fix #! line in executable Python scripts
See https://github.com/wbsoft/frescobaldi/issues/637#issuecomment-78126940 and https://github.com/wbsoft/frescobaldi/issues/637#issuecomment-78131722.

@proski:
> Another idea is to replace /usr/bin/python in executable python scripts with /usr/bin/python3 to make Python 3 the default. But that may be misinterpreted that Python 2 is no longer supported. So it's not a dot release material.

@dliessi:
> Why don't we change it to `/usr/bin/env python` instead?
> That's common practice and ensures that the script is run with the Python interpreter preferred by the user (or by the system administrator), i.e. the first one in the `$PATH`.
> That can still be overridden by explicitly calling the Python interpreter or by placing a symlink to the preferred Python interpreter earlier in the `$PATH`.
>
> ([Here](http://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my) you can find some interesting remarks.)

Independently from what we choose between the above solutions:
> Also, the interpreter in the `#!` line should have an absolute path, see [here](http://unix.stackexchange.com/questions/77512/why-not-use-pathless-shebangs) and [here](http://www.in-ulm.de/~mascheck/various/shebang/4.0BSD_newsys_sys1.c.html).

dliessi:
I didn't touch `frescobaldi_app/language_names/data.py`, but I changed the relevant part in `generate.py`.

### DIFF
--- a/frescobaldi
+++ b/frescobaldi
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 import sys
 
 from frescobaldi_app import toplevel
--- a/frescobaldi_app/debug.py
+++ b/frescobaldi_app/debug.py
@@ -1,4 +1,4 @@
-#! python
+#!/usr/bin/env python
 # This module helps with debugging Frescobaldi.
 #
 # Start a Python shell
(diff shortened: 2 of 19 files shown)


wbsoft:
I'd rather remove the shebang line for those scripts that are not meant to be executed directly.

PR #685, opened 2015-05-12
translated zh_CN.po to popplerprint.py

benluo:
I will continuely translate zh_CN.po. Please merge the new po to master.


wbsoft:
thanks!

Issue #688, opened 2015-05-21
popup window not on focus so keyboard shortcuts don't work


fedelibre:
This is a very minor issue and, I guess, also easy to be fixed.

If you close an unsaved document, a window pops up asking what to do. The focus does not change to this new popup window so using keyboard shortcuts (Alt+...) is not possible. The shortcuts work only after clicking on the popup window, making the shortcuts no more useful at all.

PeterBjuhr:
Sorry if I'm misunderstanding something here, but I can't confirm this; when the Close Document popup appears the keyboard shortcuts do work - using `Ctrl+c` cancels and `Ctrl+s` gives the save dialog.

Could it be something related to your system? What exact shortcut was you trying (Alt+?)?

fedelibre:
Yes, I guess that it depends on the desktop and window manager used. I'm on Debian, Gnome 3.

I did not specify the key after Alt as it depends on the language. Now I've started frescobaldi in english and the shortcut which should work are Alt+C to cancel and Alt+S to save. No shortcut provided for closing (this is another bug), as no character is underlined:

![fresco-close](https://cloud.githubusercontent.com/assets/1644647/7746973/042e3820-ffb7-11e4-9e38-f06d1fa1212c.png)


Here Ctrl+C and Ctrl+S do not work in the close document popup.

PeterBjuhr:
>Now I've started frescobaldi in english and the shortcut which should work are Alt+C to cancel and Alt+S to save. 

So does the use of `Ctrl` or `Alt` vary between systems? On mac I guess it would be `⌘` instead of `Ctrl`.

>Here Ctrl+C and Ctrl+S do not work in the close document popup.

I don't think you mean that, but you aren't actually hitting Ctr+Shift+c to get the upper case C? I guess there's a potential misunderstanding from the button texts.

fedelibre:
I was not precise: Ctrl+C and Ctrl+S do work but only when I click on the window and I get the focus on that window.
Does Alt work in your system? I'm used to use Ctrl+... for main window and Alt+... for pop up windows.

No, I'm not using the Shift key.

PeterBjuhr:
>Does Alt work in your system? I'm used to use Ctrl+... for main window and Alt+... for pop up windows.

Yes, either Alt or Ctrl works. I've never heard of using Alt for popups before.

PeterBjuhr:
>No shortcut provided for closing [without saving] (this is another bug), as no character is underlined.

[...]

PR #695, opened 2015-06-10
Creating an instance (with parent) instead of using a static QMessage…

PeterBjuhr:
…Box function

See #688.

### DIFF
--- a/frescobaldi_app/mainwindow.py
+++ b/frescobaldi_app/mainwindow.py
@@ -325,10 +325,11 @@ class MainWindow(QMainWindow):
         if not doc.isModified():
             return True
         self.setCurrentDocument(doc, findOpenView=True)
-        res = QMessageBox.warning(self, _("dialog title", "Close Document"),
+        msgbx = QMessageBox(QMessageBox.Warning, _("Close Document"),
             _("The document \"{name}\" has been modified.\n"
             "Do you want to save your changes or discard them?").format(name=doc.documentName()),
-            QMessageBox.Save | QMessageBox.Discard | QMessageBox.Cancel)
+            QMessageBox.Save | QMessageBox.Discard | QMessageBox.Cancel, self)
+        res = msgbx.exec_()
         if res == QMessageBox.Save:
             return self.saveDocument(doc)
         else:
--- a/frescobaldi_app/mainwindow.py
+++ b/frescobaldi_app/mainwindow.py
@@ -329,6 +329,7 @@ class MainWindow(QMainWindow):
             _("The document \"{name}\" has been modified.\n"
             "Do you want to save your changes or discard them?").format(name=doc.documentName()),
             QMessageBox.Save | QMessageBox.Discard | QMessageBox.Cancel, self)
+        msgbx.raise_()
         res = msgbx.exec_()
         if res == QMessageBox.Save:
             return self.saveDocument(doc)


PeterBjuhr:
@fedelibre I can't test this myself, will you give a try?

fedelibre:
I've pulled your branch and run the binary in the git tree, but I can't see any change.
I confirm what I wrote in the issue:

>The focus does not change to this new popup window so using keyboard shortcuts (Alt+...) is not possible. The shortcuts work only after clicking on the popup window, making the shortcuts no more useful at all.

PeterBjuhr:
Thanks! I added a commit with another try at a solution. Can I ask you to test again?

fedelibre:
It doesn't work.. but I've just realized that my report was not correct: the problem arises only when I close Frescobaldi using the `Alt+F4` shortcut. If I use `Ctrl+W`, then the dialog box is on focus and works correctly (also on master, not only in this branch you are working on).

Sorry for this late find!

PeterBjuhr:
I see! How about if you use `Ctrl+Q`, any difference?

fedelibre:
`Ctrl+Q` works correctly as well

PeterBjuhr:
Ok, good. Can you tell what (window) actually has focus when you use `Alt+F4`?

Is `Alt+F4` equivalent of using the x-button to close the window?

fedelibre:
I would say that no window has focus, not even main Frescobaldi window. In fact whatever shortcut I try doesn't work. I made a screenshot:

fresco-alt-f4

PeterBjuhr:
I pushed a new commit in the same branch. Will you try again?

fedelibre:
Same problem unfortunately.
You cannot reproduce it in your system if you use `Alt+F4`?

PeterBjuhr:
No unfortunately I can't - the focus is kept all the way.

PeterBjuhr:
I'm trying out different Qt stuff to try to emulate the behavior in your system. But I can't find any...

fedelibre:
Let me know if I can help you somehow.
Can't you get access to a Gnome 3 desktop? (a friend computer? a live CD?)

PeterBjuhr:
Yes, I was just thinking about that myself! :) I'm interested to try it out anyway...

PeterBjuhr:
Now I have a Debian OS I can test on, and I can reproduce the issue there!

To sum up, if `Alt+F4` is used to close the app the whole application loses focus (Frescobaldi is no longer shown as the active program in the top list) and the user have to activate it again before any shortcuts work.

Any other way of closing the document or the application keeps the app in focus (until it's properly ended).

I have tried different Qt stuff but haven't found anything that can override the system's behavior. And I'm not sure that it's a good idea to do so anyway.

wbsoft:
The screenshot by @fedelibre shows a behaviour (by the window system) where the MessageBox gets no window border but (because the MessageBox is modal) the app is shown darker (because the MessageBox is the only widget that can be dealt with).

That the MessageBox gets no keyboard focus is really a Qt/window system issue. Maybe we can workaround by creating a QMessageBox (`m`) instance by hand and then:
```python
m.show()
m.setFocus(Qt.ActiveWindowFocusReason)
result = m.exec_()
```

PeterBjuhr:
That doesn't work unfortunately. I think basically the problem is not that the message box is not in focus but that the app is not in focus (deactivated by the window system).

But I noticed that if you create the `QMessageBox` without a parent the app remains active in the top list. I'm not sure if that will help us though.

PeterBjuhr:
No new ideas have popped up here. For now the workaround seems to be to use `Ctrl+Q` instead of `Alt+F4`. #688

PR #706, opened 2015-07-08
Abstract viewer panel

uliska:
I create this as a pull request because it's the most convenient way of starting a review/discussion. @PeterBjuhr please also read the commit messages as they contain the reasoning.

Basically I want to start factoring out common functionality in the different viewers, and I was surprised how smooth that went so far. Anyway, I'm not feeling too sure about the architectural implications so please look quite closely.

Basically what I did is make `MusicViewPanel` a new and nearly empty class inheriting from `AbstractViewPanel`, which started off as a copy of the former class.

The next steps should be to identify the stuff/functionality that is specific to the music view and push the respective code to the new child class. Then we should try to make the manuscript viewer *another* child of that base class. Apart from the toolbar this might automatically give our manuscript viewer all the functionality of the music view (e.g. navigation and copying image excerpts (yes, this is useful too).


uliska:
if you agree with my approach I suggest continuing on this branch until the refactoring of the current music view can be considered finished for now and then merte it back to the music-viewer branch.

PeterBjuhr:
I think the approach is promising, but I would like @wbsoft to take a look before we take this any further.

I have to dig deeper into the details here but at a first glance I'm not sure if you're slicing this exactly right. I'll start with two examples:

`jumpToCursor` isn't that related to the point-and-click functionality?

`DocumentChooserAction` has a connection to the current editor document in a way I don't think we want in the Manuscript viewer.

uliska:
I haven't done any serious "slicing" yet - this is what I referred to as the next steps.  What I wanted is a first start that actually runs the same as before.
. (So your questions are both valid)

PeterBjuhr:
Sorry, I went straight for the commits and missed the next-step part in your comment somewhat!

But as long as it's only a matter of inheritance it shouldn't affect the functionality so where to slice is the main issue here.

uliska:
Yes, that's the issue to be discussed now. Going through the actions I found that practically everything should at least conceptually be available from the base class. In some cases we'll have to override the implementation, but apart from the document chooser (which I don't really see through) I think all actions should be in the base class.

PeterBjuhr:
If everything is retained in the base class it might be easier and less intrusive regarding the present code to just inherent from `MusicViewPanel` directly!?

uliska:
This is something to be considered, definitely. But I think there *are* differences, particularly on the implementation side. They will be more prominent once we add the more advanced features to the manuscript viewer. Or when we (finally) start adding significant right-click editing features to the music view.
I think on the long run an abstract base class might be more natural (and maybe also efficient). But I wouldn't argue about it.

uliska:
I've done one more preliminary but substantial step, and now I'd also say that @wbsoft should have a look and comment on our approach.

I have

* Moved both the `musicview` and the new `manuscript` modules *inside* a new `viewers` module.  
This better reflects the hierarchical structure. And it will make it cleaner to add more viewers in the future. Conceivable candidates are:
    * A Verovio viewer
    * an integrated annotations viewer (PDF or HTML files with formatted renderings of annotations and point-and-click links to the input files
    * Viewers in the context of to-be-developed Git workflows
    * etc.
* Made `ManuscriptViewPanel` also inherit from the new `AbstractViewPanel`

I think (hope) this approach is a good one. What has to be considered is that if we want to integrate the SVG viewer in that system (for example to give that the same functionality and interface) we may have to add another layer of abstraction to discern between Poppler and QWebView based viewers.

*Note:* That state *still* doesn't provide a thorough consideration of where to draw the line between actions/implementation in the base and in the concrete classes.

uliska:
What is also to-be-done is factoring out common functionality from the actual Widget classes.

PeterBjuhr:
>I've done one more preliminary but substantial step, and now I'd also say that @wbsoft should have a look and comment on our approach.

Yes, let's wait for his response!

>Conceivable candidates are:
>A Verovio viewer

The Verovio output is SVG so a Verovio viewer should be related to the SVG viewer.

(I can also repeat that it is already possible to view a Verovio SVG in the current SVG viewer.)

>What has to be considered is that if we want to integrate the SVG viewer in that system (for example to give that the same functionality and interface) we may have to add another layer of abstraction to discern between Poppler and QWebView based viewers.

Agreed, it could be considered. But I have some initial doubts about forcing different viewers into the same system.

uliska:
Am 09.07.2015 um 11:24 schrieb Peter Bjuhr:
>     I've done one more preliminary but substantial step, and now I'd
>     also say that @wbsoft <https://github.com/wbsoft> should have a look
>     and comment on our approach.
> 
> Yes, let's wait for his response!
> 
>     Conceivable candidates are:
>     A Verovio viewer
> 
> The Verovio output is SVG so a Verovio viewer should be related to the
> SVG viewer.
> 
> (I can also repeat that it is already possible to view a Verovio SVG in
> the current SVG viewer.)

Yes, but we may want them to do slightly different things some day. So
it's a *candidate*, not a concrete suggestion.

> 
>     What has to be considered is that if we want to integrate the SVG
>     viewer in that system (for example to give that the same
>     functionality and interface) we may have to add another layer of
>     abstraction to discern between Poppler and QWebView based viewers.
> 
> Agreed, it could be considered. But I have some initial doubts about
> forcing different viewers into the same system.

I can see that too, but from the user's perspective it doesn't matter
very much, and for example one day we should try to add functions like
"export image" to the SVG viewer too, or have them use the same
interface for page navigation, and then it will be good to have that
hierachical class structure.

PeterBjuhr:
A more moderate approach where existing files remain untouched were decided. I have manually and by cherry-picking merged in all other changes so this can be closed.

Issue #736, opened 2015-07-18
Viewers should get a contextmenu submenu giving access to the alternative documents


uliska:
Now that we can hide the viewer's toolbar the main thing that would probably be needed to be accessible is the list of open documents/manuscripts.

I suggest implementing a contextmenu subment that opens up to the list of open manuscripts and switches to the selected doc.

PeterBjuhr:
The contextmenu is quite crowded already so I'm not sure this is a good idea. Also following the reasoning behind why the toolbar could be hidden in #732 it had much to do with working with a singular manuscript. Perhaps it is no problem that while the toolbar is hidden some commands are in fact not accessible.

uliska:
> The contextmenu is quite crowded already so I'm not sure this is a good idea. 
You're right about the contextmenu. However, I think it *is* a good idea (now more than before).

The fundamental reason for hiding the toolbar is not in using a single manuscript but in saving screen estate when everything important (especially (and explicitly) zooming and scrolling) are available through other means (mouse, contextmenu). This is equally true when working with one or multiple manuscripts. 

I think when we allow hiding the toolbar we should make sure all important functionality can be accessed without having to show the toolbar first. The print button may be an exception because that's (probably) not a regular task. But switching between documents should definitely be accessible through the contextmenu and/or keyboard shortcuts.

I would rather suggest combining things further. The zoom actions could definitely be grouped in a submenu (as you'll probably do that with the mouse wheel most of the time), probably also the three close actions.

I will first implement that grouping and then come back to the original question.

uliska:
OK, I've implemented it and completely reviewed the context menu. It is now grouped by meaningful parameters:

- actions affecting the current link/selection  
*Note:* The "edit in place" item doesn't show up anymore (#740)
- actions affecting the opened documents  
(I will add the "Show" submenu above the "Close" submenu)
- actions affecting the viewer's global state.

When checking make sure to also open it on a link and on a selection.

uliska:
*Edit:* The "edit in place" issue was a false alarm.

uliska:
The main issue has been merged with the #742 pull request. Everything else has been solved.

PR #742, opened 2015-07-22
Viewer show submenu

uliska:
I need some help now, after fiddling around for some (intermittent) time over a few days.

I try implementing a context-sub-menu giving access to the alternative open documents in the manuscript viewer. Building the menu seems to work, but I don't get the actions to trigger properly.

I have two different approaches (should be clear from the comments in `manuscript.contextmenu`) and both don't work. In one case the slot isn't executed at all - this is probably due to the fact that the object has already been deleted before the action is triggered. In the other case it's always the *last* item that defines the action - so you can only switch to the last document in the list.

### DIFF
--- a/frescobaldi_app/viewers/contextmenu.py
+++ b/frescobaldi_app/viewers/contextmenu.py
@@ -91,6 +91,11 @@ class ViewerContextMenu(QObject):
         sm.addAction(ac.music_zoom_out)
         sm.addAction(ac.music_zoom_original)
 
+    def addShowActions(self):
+        """Add actions to show alternative documents.
+        This is not implemented in the base class"""
+        pass
+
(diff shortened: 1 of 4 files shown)


uliska:
Relates to #736

PeterBjuhr:
I think you need to group the actions with `QActionGroup` and connect the group to the slot.

uliska:
OK, I'll look in that direction, thanks.

uliska:
Unfortunately this isn't the case. I keep the implementation as ActionGroup because it looks good and is more consistent with the `Sessions` and `Documents` menus. But the problem persists, I think when the action is triggered the target object doesn't exist anymore.

uliska:
Fixed. But I'll merge manually

Issue #753, opened 2015-08-03
Create a Tools->Viewers submenu


uliska:
By now we have three viewers: Music View, Manuscript View and SVG View. And more might come (e.g. the Pre-View as suggested in #569).

It might be more consistent to create a submenu in the Tools section where all viewers can be combined. What do you think @PeterBjuhr and @wbsoft  (or anybody else)?

PeterBjuhr:
A submenu makes the menu items harder to access so I wonder what is the reason for this change?

>It might be more consistent[...]

There are now other submenus among the tools so are you thinking about consistency in regards to the pitch, rhythm etc menus?

I think the relevant reason should be that the list of tools is growing too long...

uliska:
There is a trade-off between "complexity" of additional navigation to a
submenu and having to "browse" a flat list of more or less unrelated items.
In my perception the "Tools" list (i.e. the generated list of
dockpanels) has already become too long and crowded, and the additional
manuscript viewer is just that extra bit that is too much. And
considering that there might (probably will) be even more viewer entries
in the future I found it's time to add that extra level.

However, this is probably very much a matter of taste or expectations,
that's why I didn't simply push this on our branch but opened a PR.

PeterBjuhr:
Grouping all viewers is of course an obvious way of creating a submenu. Another possible grouping is the MIDI player and MIDI input. 

A less obvious way of thinking is to try to keep easy access to tools that is expected to be opened often and put the other tools in (a) submenu(s).

The contextmenu with a list of tools is growing equally and could also be considered in this regard.

uliska:
I won't delete the `viewers-menu` branch although I closed the PR (#754). We'll consider this after the manuscript viewer is merged.

uliska:
This has now long be done.

Issue #750, opened 2015-07-29
[manuscript dev] Viewer preferences


PeterBjuhr:
Although this probably is related to #735 and more or less on topic there, I think it's better to open a new issue for this. 

In preferences->tools are the preferences for the music viewer. We should have a look which of these settings also applies to the manuscript viewer. If they do apply to both viewers that should be clear I think.

I also intend to add a preference for sorting the manuscripts in the document chooser.

Maybe we should add a new section `Viewers` to the preferences. And have generic preferences for all viewers, Music view preferences and manuscript view preferences!??

uliska:
I had also thought about this and had the following idea:

- create a Preferences Page "Viewers" (as you suggest)
- access the different viewers through a tab interface
- add an option to "link" the options for the different viewers
  (that is for those that make sense, e.g. magnifier scale)
Maybe this will include all options, maybe not.

I think it would be good if the user can choose to have the settings of
the different viewers identical or not.

PeterBjuhr:
I don't know about the tabs. With an own page there would be plenty of room.

I have also doubt about the linking. If the settings are the same by default and the user could change that wouldn't that be enough?

But perhaps there's no point in generic settings. Music view preferences and manuscript view preferences that probably have the same defaults...

PR #754, opened 2015-08-04
Add "Viewers" submenu to Tools menu

uliska:
Implements #753.

What do you think?

### DIFF
--- a/frescobaldi_app/menu.py
+++ b/frescobaldi_app/menu.py
@@ -285,7 +285,8 @@ def menu_lilypond_generated_files(mainwindow):
 
 def menu_tools(mainwindow):
     m = Menu(_('menu title', '&Tools'), mainwindow)
-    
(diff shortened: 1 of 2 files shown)


PeterBjuhr:
I would prefer if we didn't merge this into the manuscript-viewer branch but waited and created an own pull request for this. The reason is that the upcoming pull request (from the manuscript-viewer branch) is quite overwhelming already both in number of commits and in files/lines of code to review. And I would like the PR to retain as much focus as it can.

For the same reason I also think we should wait with the preference stuff (#750) to a later stage.

uliska:
OK. This will then also allow us to have a broader discussion about the actual "design" of the menu, and possibly a somewhat cleaner implementation (if there should be more than one submenu I think we should have a better solution than a hard-coded `if .. in` conditional, maybe a nested list. Or querying if the panel is from the `viewers` module (where the others will eventually be moved).

PR #755, opened 2015-08-05
Widget creation

### LINKED ISSUE #735: Store and retrieve state(s) from viewers in preferences
This came up in #732.
It's not clear yet whether we need a "Viewers" Preferences page, but in any case I think some stuff should be saved/restored using the QSettings interface.
Among them should be the visibility of a viewer's toolbar and the "synchronize with cursor" option. As the latter is already saved we should look into that for a start.

uliska:
This does *not* fix #735 at all, but hopefully makes the process of creating the widget and handling sessions more sane.

Now the session handlers are called *after* the widget has been created, avoiding the loop caused by the implicit calls of  `widget()`.

### DIFF
--- a/frescobaldi_app/viewers/__init__.py
+++ b/frescobaldi_app/viewers/__init__.py
@@ -135,6 +135,18 @@ class AbstractViewPanel(panel.Panel):
         ac.viewer_show_toolbar.triggered.connect(self.slotShowToolbar)
         self.mainwindow().allDocumentsClosed.connect(self.closeAllViewdocs)
 
+    def widget(self):
+        """Ensures that our widget() is created and intitialized and returns it.
+        Overrides the default implementation of Panel because we need to call
+        the session handler *after* the widget is created."""
+        w = super(panel.Panel, self).widget()
+        if not w:
+            w = self.createWidget()
+            self.setWidget(w)
+            import sessions
+            self.slotSessionChanged(sessions.currentSession())
+        return w
+
     def _createConreteActions(self):
         """Create the actionCollection.
         Subclasses must override this method."""
--- a/frescobaldi_app/viewers/__init__.py
+++ b/frescobaldi_app/viewers/__init__.py
@@ -181,9 +181,10 @@ class AbstractViewPanel(panel.Panel):
         This is the lowercase classname, right-stripped
         of a trailing 'panel'.
         To be used for accessing the QSettings group."""
-        result = type(self).__name__.lower()
-        result = result if not result.endswith('panel') else result[:-5]
-        return result
+        if not hasattr(self, '_viewerName'):
+            name = type(self).__name__.lower()
+            self._viewerName = name if not name.endswith('panel') else name[:-5]
+        return self._viewerName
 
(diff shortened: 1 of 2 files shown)


PeterBjuhr:
I haven't had time to look into this fully. But my initial thought is that the code should really be restructured to avoid the problem in the first place instead of a workaround created. Have you considered that?

uliska:
Hm. Of course I'm not completely sure about that, but I had the impression my reasoning was sound:

- The session handler must only be connected/called after the widget has been created and not upon creation of the panel (because the session handler *has* to access `widget()`).
- So we have to explicitly call the session handler after creating the widget (because at that moment there won't be a sessionChanged signal).
- When this is called from within the widget's `__init__()` method we can't avoid the `widget()` loop.
- That means we have to find a way to call the session handler *after* the widget has been created, but *immediately* after that.
- Moving that call to an overridden `widget()` method was the only way that I could think of implementing that.

I have the impression it's quite straightforward and not actually a hack: Instead of connecting the sessionChanged signal when the panel is created we connect it while the widget is created and execute it once immediately after the widget is created.

Does that make sense?

PeterBjuhr:
The basic idea is to call `self` instead of `widget()` that is handle things from the inside instead of the outside. You mentioned yourself moving `slotSessionChanged` back to the widget. I think that is a good start. 

To go more into details, one of the problematic lines is this (https://github.com/wbsoft/frescobaldi/blob/manuscript-viewer/frescobaldi_app/viewers/__init__.py#L711):

````
self.parent().widget()._positions[doc] = position
````

There was an issue with this from the start because you're accessing a protected variable (I've reached my quota with nagging about protected variables so I didn't report it). But now with the widget init loop the problem is of course acute.

Let me try to restructure the code. If I fail we can return to your solution!

I will first deal with #756 because down the line the same method is affected so the implementation will  be dependent of what happens there.

uliska:
> The basic idea is to call |self| instead of |widget()| that is handle
> things from the inside instead of the outside. You mentioned yourself
> moving |slotSessionChanged| back to the widget. I think that is a good
> start.
> 

I had already tried this but gave up on it when I realized that the
problem is *not* in the session handler but in the document chooser (as
you quote below). So I don't think anymore that moving the session
handler will help.

PeterBjuhr:
Therefore I called it 'a good start' and not the full solution. :) I think you can see from the rest of my comment that it involves more restructuring than that.

uliska:
@PeterBjuhr how does this PR relate to what you did recently? Should we close that branch or is it worth sorting out the merge conflicts?

PeterBjuhr:
I did add 9453c3790f90ed40f124eef5701b1fb935007edc. But I did manage to restruct the code as mentioned so I don't think b037f3e12402d6ce840fac8f6e72274887e83e75 is relevant any more.

uliska:
OK. Then **This pull request is not intended to be merged anymore**, but we'll leave it open for a while.

uliska:
Thinking about it again I don't see any use in keeping this, it's only confusing things.
(However, I won't immediately delete my local branch)

PR #772, opened 2015-10-26
Adding macro functionality for snippets

PeterBjuhr:
I got the idea of macro snippets, that is snippets that put together other snippets and also other Frescobaldi menu options.

For snippets the `name` variable has to be used both in the original snippet and in the macro.

For menu options I use the name of the action which is a bit awkward and not at all straight-forward from an user perspective. Perhaps a matching against the `action.text()` could be used instead.

Also it's perhaps a bit clumsy to collect the snippets and the actions here when they are already collected elsewhere...

### DIFF
--- a/frescobaldi_app/snippet/insert.py
+++ b/frescobaldi_app/snippet/insert.py
@@ -53,6 +53,8 @@ def insert(name, view):
         # insert the snippet, might return a new cursor
         if 'python' in variables:
             new = insert_python(text, cursor, name, view)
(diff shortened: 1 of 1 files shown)


wbsoft:
I think it is a nice idea. I would like to make menu actions available as commands anyway, a bit like emacs. So we can keep this in place, and when there is a nicer way to call actions (instead of finding te QAction and triggering it).

PeterBjuhr:
Great.

Issue #169, opened 2013-08-20
Comprehensive Tap Tempo in the Score Setup Wizard


pl-gauthier:
TL;DR: The wish if for "An averaged tap tempo counter with manual reset". 

The behavior I'm looking for, to my knowledge, cannot be found in any software and yet it seems fairly easy to implement.

Also, we know it is somewhat impossible for a human to tap exactly on time on each down time. So if the tempo calculation is made at each tap, without past taps in context, the value will change each time a user clicks on the tap tempo button.

So how would it work:
Let's say the song is paced at 120 BPM. One would click on the tap tempo button on each down time. At the first click the timer starts and after each click, an evermore precise average is made. If the user hasn't stop clicking on each down time, the tempo value can only be more and more precise and accurate, even if he's not tight at all! If the process didn't go as expected, a manual reset button would reset the timer and set the tap tempo calculator on standby waiting for the first click.

So, the more you tap, the more accurate is the tempo.

That means (Just to explicit the behavior):
If during a 4/4 song of a 12 bars song paced at 120 BPM a user taps only once at the beginning of the song and then, during the last bar, clicks all the 47 remaining down beats (like only a robot could do): the tempo indicator will still display a strong 120.000 BPM.

Would it be a good idea?
Thanks for considering.

PR #792, opened 2016-01-12
Round tap tempo value to nearest common tempo

sebleblanc:
This commit introduces a checkbox underneath the tap tempo widget in the score wizard. This checkbox is labelled "Round tap tempo value" and allows the user to enter a very precise tempo using the tap tempo button.

When the checkbox is set (default value for newer versions of Frescobaldi), behaviour is not changed, however when it is turned off, the tempo entry box will show the raw tempo from the button. For example, if the user tapped at exactly 117 BPM, the value shown will not be rounded to 116, i.e. one of the values in the tempo box.

This modification is a request from a friend, @pl-gauthier.


pl-gauthier:
#169

wbsoft:
thanks!

Issue #683, opened 2015-04-28
Auto completion of \repeat (unfold) doesn't work in some contexts


PeterBjuhr:
Reported on the Frescobaldi mailing list:

>I've noticed many times that autocompletion goes inactive if I'm in a context other than a normal note-containing voice.

>````Lilypond
>{ \repeat unfold 4 { c' d' e' f' } }  % autocompletion works

>\drummode { \repeat unfold ... }  % nope, you're typing by hand

>\chordmode { \repeat unfold ... }  % same problem 
````

PR #795, opened 2016-01-21
Add also support for drum mode

PeterBjuhr:
See #683.


dliessi:
@PeterBjuhr Most of the changed lines in your commit are whitespace changes.
May I suggest that you separate whitespace changes and the actual content you added in two commits?
That would keep the commit history easier to read.

PeterBjuhr:
Sorry about that! I've cleaned up this PR now!

PR #819, opened 2016-05-01
Spanish translation update.

pacovila:
Made against 3.x branch; maybe it is valid for 2.20 as well?


wbsoft:
Many thanks!! Will also apply to 2.2 branch.

PR #846, opened 2016-08-16
New Swedish translations

dag:
Covers most of the UI but little of the documentation.

Based on unupdated .pot as it was sometime in June, IIRC.

Fixes some apparent mistakes in the source messages WRT things like
inconsistent capitalization.  I added a comment to many of those messages.

Conservative WRT menu accelerators: only added when the original
accelerator letter is available.  This means there are some 70 menu items
that lose their accelerator and don't provide any new one.

I tried to stay at least somewhat consistent with the Swedish translation
of MuseScore.


wbsoft:
Thanks!

PR #849, opened 2016-08-27
Retinasupport in qpopplerview

### LINKED ISSUE #816: HiDPI or retina displays for preview
First: thanks for creating Frescobaldi!
Second: I am using Frescobaldi on a Macbook Pro with Retina display and Frescobaldi’s preview window looks low-res / pixelated. Is there an option to make the preview window retina resolution as well?
Best regards
Ilja

drydh:
This resolves #816 (high DPI rendering on Mac Retina displays in music view) + some minor fixes. Only qpopplerview has been modified, not qpageview (replacement of qpopplerview in development).

### DIFF
--- a/frescobaldi_app/macosx/icon_drag_eventhandler.py
+++ b/frescobaldi_app/macosx/icon_drag_eventhandler.py
@@ -26,7 +26,8 @@ Currently this makes only sense on Mac OS X.
 
 
 from PyQt5.QtCore import QEvent, QMimeData, QObject, QPoint, Qt
-from PyQt5.QtWidgets import QApplication, QDrag, QStyle
+from PyQt5.QtWidgets import QApplication, QStyle
+from PyQt5.QtGui import QDrag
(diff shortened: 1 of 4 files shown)


drydh:
Cached images for Retina screens gets quite big when using the magnifier, typically something like 8000 x 6000 pixels. This is about 200 MB so default max cache size of 100 MB is used up by a single magnified picture. Not a big problem but perhaps the user should be able to set the max cache size in the preferences? (e.g., when magnifying and moving between two pages, only one is kept in cache at high resolution at a time)

wbsoft:
Thanks! I can probably also add this functionality to qpageview based on your work :-) I already planned that, but unfortunately I have no machine to test. Thanks again!

drydh:
I could add that functionality to qpageview at some point as well (or test your implementation). How complete is qpageview? Is it soon ready to replace qpopplerview?

PR #851, opened 2016-08-28
Feature: zoom with pinch gesture in popplerview

drydh:
Enable zoom using pinch gesture on trackpad (works at least under Mac OS X).


wbsoft:
thank you!

PR #852, opened 2016-08-28
Fix debug-layout-options for LilyPond 2.19.22+ (#801).

### LINKED ISSUE #801: Update layout control options to LilyPond 2.19.21 syntax
Many of the layout control options are affected by a fundamental syntax change about using the `parser` argument in music-etc-functions.

The task is to wrap the offending calls in version predicate conditionals because Frescobaldi has to support older _and_ newer LilyPonds.

drydh:
Fixes #801.

### DIFF
--- a/frescobaldi_app/layoutcontrol/debug-layout-options.ly
+++ b/frescobaldi_app/layoutcontrol/debug-layout-options.ly
@@ -40,32 +40,39 @@
 
 debugLayoutOptions =
 #(define-void-function (parser location)()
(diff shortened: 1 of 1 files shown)


uliska:
LGTM.
I don't merge it because I'm only looking at it online.

Thank you for taking that.

uliska:
@drydh I have taken a slightly different approach but (finally) fixed this embarrassing issue separately in 80fed97ecbe96a27cc4f2ec1226ff21fefc00381

Issue #797, opened 2016-02-03
MIDI input doesn't work on Windows 8.1


unmostro:
MIDI input from external synthesizer (Yamaha) doesn't work on Windows 8.1 both with MIDI--USB and native MIDI-MIDI interface. At the same time MIDI output works well.
MIDI input interface itself is OK since the other application (MIDI-OX)  displays all input MIDI events correctly.

unmostro:
Denemo version 2.0.0 with the same MIDI settings handles MIDI input perfectly.

unmostro:
Denemo version 2.0.0 with the same MIDI settings handles MIDI input perfectly.

HeyItsJustMe:
FYI same on Windows 10. So frustrating. Here: https://github.com/wbsoft/frescobaldi/issues/853

Issue #853, opened 2016-08-29
MIDI Input doesn't work on Windows 10 - BROKEN


HeyItsJustMe:
Sorry but MIDI input is not working on Windows 10 64bit. I've tried several MIDI keyboards, none work. But I can confirm MIDI is being transmitted correctly from these devices, but Frescobaldi will not do MIDI input. It just does nothing. Start and stop capture = completely broken on Win10.  MIDI-OX shows perfectly fine MIDI data transfer FYI.

Can this be fixed please? It's such a hassle now. Thank you.

Note: I'm happy to report that the MIDI devices are working with MIDI INPUT in Frescobaldi on Windows 7 64. This is ONLY a problem with Windows 10 and 8.1 and Frescobaldi MIDI INPUT. Really hope you can fix this soon.

drydh:
MIDI input doesn't work for me either (on Mac OS 10.10). MIDI output is fine. This is with Frescobaldi 2.19.0. I'll investigate this further when I've been able to get pyportmedia/pypm/pygame running under Python3 so I can debug (or I'll get the Python 2.x version running).

Migrating to mido (#579) or pyrtmidi would perhaps fix these issues but could involve quite some work.

HeyItsJustMe:
But it works perfectly fine in Windows 7. What changed with Windows 10 that MIDI input is so terribly unusable? It is great in Win7.

Such a headache. I love Frescobaldi but this is a tough situation.

drydh:
I found the bug causing my problem: Frescobaldi has **never** honored the choice of the midi input device in the preferences due to a confusion between the prefixes 'midi.midi' and 'midi'! It therefore always uses the default input device. To check if this is also causing your problem, find the preferences for Frescobaldi (in the registry on Windows?) and change the key midi.midi.input_port to midi.input_port and see if this fixes your problem.

HeyItsJustMe:
Would that have been really so different on Windows 7 where it only would cause a problem on 8.1 and Win10? Odd. I will try your suggestion and post my results here. Thank you for the tip!

drydh:
I don't have a Windows machine but looking at other issues, the key in the registry is probably (at least under Windows 8.1) called:

HHKEY_CURRENT_USER\software\Frescobaldi\frescobaldi\midi\midi\input_port

and should be changed to

HHKEY_CURRENT_USER\software\Frescobaldi\frescobaldi\midi\input_port

drydh:
[...]

PR #855, opened 2016-08-30
BUGFIX: MIDI Input setting was never used!

drydh:
Could fix issues #797 and #853.

- midiinput/__init__.py used "midi/input_port" and preferences used "midi/midi/input_port".
- Bug has existed since first implementation of midi-input in commit 9473aa87 (2013-11-12).


drydh:
Added another midi input fix: discard all events besides NoteEvents.

wbsoft:
thanks!

PR #874, opened 2016-12-02
Allow MusicXML export to follow \include's

### LINKED ISSUE #873: Follow \include-s for MusicXML export
Currently MusicXML export only works for the open document and doesn't respect \includes. Which can be seen in https://github.com/wbsoft/frescobaldi/blob/master/frescobaldi_app/file_export/__init__.py#L53, specifically in [this line](https://github.com/wbsoft/frescobaldi/blob/master/frescobaldi_app/file_export/__init__.py#L65).

This is  major annoyance because it simply doesn't export *anything* in files that make use of any modularization. I think Frescobaldi is able to follow includes, as it does for example with autocompletion. So it should be rather painless to do.

I'd like to look into this but ask here because someone might be faster telling me where to look that I can find myself. @PeterBjuhr @wbsoft ?

uliska:
This is functionality that has already been available with python-ly
but simply not "activated" in Frescobaldi.

Closes #873

I hope this is sufficient as it simply passes the filename of the
currently open tab. If anything like respecting master documents
should be respected some more work would be involved (but actually
I don't think that should prevent this from being merged).

**NOTE:** I have tested this on the `2.19` branch but I don't think there should be any issue applying it to current master as well.

### DIFF
--- a/frescobaldi_app/file_export/__init__.py
+++ b/frescobaldi_app/file_export/__init__.py
@@ -62,7 +62,7 @@ class FileExport(plugin.MainWindowPlugin):
             return False # cancelled
         import ly.musicxml
         writer = ly.musicxml.writer()
-        writer.parse_text(doc.toPlainText())
+        writer.parse_text(doc.toPlainText(), orgname)
         xml = writer.musicxml()
         # put the Frescobaldi version in the xml file
         software = xml.root.find('.//encoding/software')


uliska:
@PeterBjuhr bump?

PeterBjuhr:
As far as I can tell this seems good! But I would like @wbsoft to take a look aswell!

uliska:
I can't imagine any serious side effects here, so I take the liberty to merge it now.

Issue #899, opened 2017-03-04
The GUI's zh_CN translation is not complete.


codeworm96:
Hi everyone,

I am a new user of frescobaldi. I found that some text in the GUI is not translated when language is set as zh_CN in frescobaldi.

As a native speaker of Chinese, I would like to provide some help in improving zh_CN translation, if needed.

uliska:
Hi Zhang,

Am 04.03.2017 um 06:00 schrieb Zhang Yuning:
>
> Hi everyone,
>
> I am a new user of frescobaldi. I found that some text in the GUI is
> not translated when language is set as zh_CN in frescobaldi.
>

Most translations are not completely up-to-date, because this relies on
people like you to do the work. So thank you for getting in touch with us.

> As a native speaker of Chinese, I would like to provide some help in
> improving zh_CN translation, if needed.
>

Did you already read
https://github.com/wbsoft/frescobaldi/blob/master/README-translations?
Basically all you have to do is take the file
https://github.com/wbsoft/frescobaldi/blob/master/frescobaldi_app/po/zh_CN.po
and edit it to your liking.

If anything is still unclear don't hesitate to ask.

Best
Urs

codeworm96:
Thank you for providing me those useful links. I am trying to improve the Chinese translation, but it will take some time.

PR #900, opened 2017-03-05
Update Traditional Chinese translation

snyoung:
It has nothing to do with #899, it was my own translation to zh_TW (Traditional Chinese), with some corrections and some new ones.

Please do not merge my translations directly into the zh_CN ones; some terms might not be compatible even after the character conversion. And sorry for not knowing which terms are compatible, since I can not read Simplified Chinese.

Those terms with no conventional Traditional Chinese translation (or I don't know for sure) and those files in the /userguide directory are yet not translated, I may translate them later.


snyoung:
I added back the translating contribution by Anthony Fok to the credits, which I deleted by some mistake. The mo file might need some recompiling.

Issue #872, opened 2016-11-29
wish: modify how Frescobaldi creates default filename for new score


wbsoft:
From a user:
> Is it possible to modify how Frescobaldi creates file names for new 
> scores? I love the fact that a file name is created based on the
> title and composer in the header block, but I would like to use the
> format "Title (Composer)" instead of the current "Composer-Title"
> form.
> 
> Is there template somewhere that I can modify or is this hardcoded?

This is currently hardcoded in [`documentinfo.defaultfilename()`](https://github.com/wbsoft/frescobaldi/blob/eec60717e4b5992f2abbdef30fe9a6e546b0001e/frescobaldi_app/documentinfo.py#L68-L99).

It would be nice if this function could accept a user-generated string from the preferences dialogs so the way a new default filename is created is configurable.

wbsoft:
Fixed in pull #902

PR #902, opened 2017-03-05
Support custom default file name template

codeworm96:
This is my attempt to provide the function described in #872.
I add a new option in the preferences dialog, user can turn this function on and provide a template string to configure the way a new default filename is created.

### DIFF
--- a/frescobaldi_app/documentinfo.py
+++ b/frescobaldi_app/documentinfo.py
@@ -96,9 +96,9 @@ def defaultfilename(document):
     if s.value("custom_default_filename", False, bool):
         template = s.value("default_filename_template", "{composer}-{title}", str)
         for k in fields:
-            if k not in d:
-                d[k] = "unknown"
-        filename = template.format(**d)
+            if k in d:
+                template = str.replace(template, '{' + k + '}', d[k])
+        filename = template
     else:
         filename = '-'.join(d[k] for k in fields if k in d)
     if not filename:
--- a/frescobaldi_app/documentinfo.py
+++ b/frescobaldi_app/documentinfo.py
@@ -91,8 +91,16 @@ def defaultfilename(document):
     # make filenames
     for k in d:
         d[k] = re.sub(r'\W+', '-', d[k]).strip('-')
-    
-    filename = '-'.join(d[k] for k in fields if k in d)
+
+    s = QSettings()
+    if s.value("custom_default_filename", False, bool):
+        template = s.value("default_filename_template", "{composer}-{title}", str)
+        for k in fields:
+            if k not in d:
+                d[k] = "unknown"
+        filename = template.format(**d)
+    else:
+        filename = '-'.join(d[k] for k in fields if k in d)
     if not filename:
         filename = document.documentName()
     ext = ly.lex.extensions[i.mode()]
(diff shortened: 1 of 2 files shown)


uliska:
Thank you for this pull request. I assume you are one of those who reached out to us because of GSoC? It would be good then to know who of them you are ...

One technical issue with your commit: it should *not* include the `.pot` file, so you should re-commit it.

I've tried out your code and at first sight it is very nice - the preference looks good and works, and when using `title` and `composer` it does what is expected upon saving the file.

However, using any other header field in the template string triggers a KeyError exception. This is because it isn't present in the `fields` tuple.

The straightforward solution would be to add all possible header fields to the `fields` tuple, but the problem is that a user can specify arbitrary non-standard header fields and may want to use them in the default filename as well.

So you should try to igenerate the field list from those present in the template string (of course only if the option is activated).

codeworm96:
> I assume you are one of those who reached out to us because of GSoC? It would be good then to know who of them you are ...

Yes. I am seeking a suitable project to apply. After familiarizing myself with frescobaldi and trying to make some contribution. I decide to choose frescobald. I will introduce myself in the mailing list soon. 

> One technical issue with your commit: it should not include the .pot file, so you should re-commit it.

Fixed. I modified that commit to remove the .pot file, so you may need a '--force' option when pulling.

> However, using any other header field in the template string triggers a KeyError exception.

Sorry for my carelessness. An end user application should be as robust as possible against unexpected input. I thought about the problem again. Maybe the format() method is too flexible that user have a lot of ways to construct a template that causes some errors(for example, non-standard header, out-of-range index and extremely long padding). As to this function, maybe a str.replace() is enough? In the fixed code, I just replace supported headers and leave alone unknown headers.

uliska:
Thanks, this is definitely better already, as it prevents the exception.

However, this way it will only support the two fields `title` and `composer` again, which is of course not enough to pass as "custom" ...

I see two possible improvements (actually as said previously):
* add all known header fields to the tuple definition in line 79
* Retrieve all field names from the actual template string and feed these to the tuple

The second one would require some more reworking of the code, but I think this would be the way it should be (because the user can use arbitrary fields if they tend to use them in the header).

uliska:
> I am seeking a suitable project to apply. After familiarizing myself with frescobaldi and trying to make some contribution. I decide to choose frescobald. I will introduce myself in the mailing list soon.

Ah, I thought you were one of those who had already introduced themselves. Well, opening a pull request is a good way to start that process ...

uliska:
Thinking again, it may actually be sufficient to add all header fields that are used by LilyPond's titling. It's about the *preset* filename after all and doesn't limit what can be used in the Save dialog.

codeworm96:
I adopted the second way to allow user to use more fields in their template.
It retrieves all field names from the template string and try to find them in the document now.

uliska:
From *my* POV this is good to go now and I'd like to pass it on to @wbsoft to actually merge it.

One thing that *could* be discussed is the behaviour for unset variable. Instead of evaluating to `-unknown-` it might be suppressed, otherwise I end up with a filename like `My-title-unknown-unknown.ly` - which might be better reflected as simply `My-title.ly`. But I wouldn't make this a requirement for/against merging.

wbsoft:
Thank you, Zhang!, And Urs, for your help! We could fix the 'unknown' thingy if it's needed later.

Issue #890, opened 2017-02-15
wish: (programmable) keyboard shortcut for accidentals in midi input


wbsoft:
User question:
> Is it possible to have a keyboard short-cut for the Accidentals in the 
MIDI Input panel?  A keyboard short-cut for sharps or flats.
> 
> Or, can I somehow program that now.

uliska:
Implemented in https://github.com/wbsoft/frescobaldi/pull/905

PR #905, opened 2017-03-07
Keyboard shortcut for Midi input accidentals

prabhanshuguptagit:
Added a keyboard shortcut for switching between accidental styles (flats or sharps), requested in Issue #890. I've kept the default shortcut to be Ctrl+Shift+3 and icon transpose-tools.svg.

### DIFF
--- a/frescobaldi_app/midiinput/tool.py
+++ b/frescobaldi_app/midiinput/tool.py
@@ -22,6 +22,7 @@ class MidiInputTool(panel.Panel):
         ac = self.actionCollection = Actions()
         ac.capture_start.triggered.connect(self.slotStartCapturing)
         ac.capture_stop.triggered.connect(self.slotStopCapturing)
+        ac.accidental_switch.triggered.connect(self.slotAccidentalSwitch)
         actioncollectionmanager.manager(mainwindow).addActionCollection(ac)
         mainwindow.addDockWidget(Qt.BottomDockWidgetArea, self)
     
@@ -34,7 +35,10 @@ class MidiInputTool(panel.Panel):
     
     def slotStopCapturing(self):
         self.widget().stopcapturing()
-    
+
+    def slotAccidentalSwitch(self):
+        self.widget().switchaccidental()
+
     def createWidget(self):
         from . import widget
         return widget.Widget(self)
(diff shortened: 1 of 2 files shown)


uliska:
As you can see from the referenced new issue I've tried to check your PR. On my machine it doesn't seem to have any effect, but I'm not sure if that's caused by your code or an underlying secondary issue.

Does your code *do* what you expect on your machine?

prabhanshuguptagit:
> While investigating #905 I realized that the shortcuts for the MIDI input tool aren't listed/editable in the Preferences.

The shortcut is getting registered in my preferences list. Also the shortcut is working as it is expected to.

image

Could the issue be due to a version incompatibility? These are the versions I am running:

```
Frescobaldi: 3.0.0
Python: 3.5.2
python-ly: 0.9.5
Qt: 5.8.0
PyQt: 5.8
sip: 4.19.1
```
Does anything seem off?

uliska:
OK, I've found it. Sorry, but I searched for "MIDI" because it should be part of the miditools ...

a)  
Either the name "Switch accidental style" should be changed to something clearly indicating that this is for MIDI input or (actually better) the MidiTool shortcuts should be grouped. It's not related to your pull request, but I think having Start/Stop capturing as such top-level entries is also inappropriate.  
@wbsoft how would one create a Preference group?  

b)  
How is the shortcut working as expected? I can't test it in MIDI capturing because I don't have a MIDI keyboard available, so I don't see if it does that while entering music. But on my machine it doesn't have any effect on the check box in the panel. Does it do change that for you? (I don't think it's a version issue).

c)  
The box around the Accidentals radio buttons expands vertically to fill the panel:

miditool

It would be a nice extra challenge if you could fix this in the context of this PR (of course this doesn't have anything to do with *your* code, it's just an opportunity.

prabhanshuguptagit:
The shortcut does toggle between the two radio buttons on my machine. I'll check if I've used any incompatible methods. 

>a) 
Either the name "Switch accidental style" should be changed to something clearly indicating that this is for MIDI input or (actually better) the MidiTool shortcuts should be grouped. It's not related to your pull request, but I think having Start/Stop capturing as such top-level entries is also inappropriate.
@wbsoft how would one create a Preference group?
 
Should MIDI input and player tools be grouped together or separately? I think creating a single MidiTool group would serve the purpose.

uliska:
> Should MIDI input and player tools be grouped together or separately? I think creating a single MidiTool group would serve the purpose.

I thinks so too. But this kind of questions should always be discussed with @wbsoft too.

uliska:
I'm sorry to admit that my problems were related to a misconfiguration of my keyboard. So Federico's comment made me double-check everything - and now it works.

I've taken the liberty to basically revert the last commit and replace it with a much simpler (one-line) solution in https://github.com/wbsoft/frescobaldi/commit/a1437ca12ec1a9896f9679fa40f29d2e10a924d6

prabhanshuguptagit:
Glad to know that it worked.

> Either the name "Switch accidental style" should be changed to something clearly indicating that this is for MIDI input or (actually better) the MidiTool shortcuts should be grouped. 

The regrouping of the shortcuts list is yet to be done.

Issue #299, opened 2013-11-28
wish: better autocompletion


jan-warchol:
Here are some ideas about autocompletion that might be interesting:
- since it is very rare to have commands differ only by letter case, it would be nice to make the dropdown list with suggestions case-insensitive (i.e. when i write \sluru, it should still list \slurUp as a suggestion)
- allow partial autocompletion with TAB key, like in bash - i.e. if i write \sl and hit TAB, it will be autocompleted to \slur (because there are no commands that begin with \sl but don't begin with \slur)
- add an option for automatic autocompletion - i.e. when there is only one possibility to autocomplete it will be inserted automatically.  This should be turned off by default, but it could be very handy for advanced users.  For example, it would be awesome to be able to type just \voo for \vo(ice)O(ne).

uliska:
I think accepting the suggestion by `Enter` is a good thing, and I wouldn't want to have it autocompleted automatically (if that's what was suggested).

The TAB suggestion is very nice. Additionally I suggest using the TAB key for iterating through the suggestion list. I think TAB is much easier to reach than the arrow keys on most keyboards.

jamesthechamp:
hey,
i am a new contributor would love to contribute and participate for gsoc 2017.
i made a patch for this issue #918
here i implement the functionality to iterate over suggestion list as show follow
![peek 2017-03-23 03-59](https://cloud.githubusercontent.com/assets/11019397/24223846/023c43c6-0f7e-11e7-8126-a648573d33fc.gif)
please have a look on it.

uliska:
Implemented so far (by @jamesthechamp or me):

* Partial autocompletion with TAB
* Common subsets are automatically filled (but selected so they can be killed again)
* TAB cycles through suggestions if there's nothing to be partially completed
* Accepting by ENTER is still necessary, but one doesn't have to select an item with arrow keys or mouse

uliska:
Implemented case insensitivity.

Now matching is done case insensitive. But the document text is only updated upon acceptance

uliska:
I will not implement automatic autocompletion. As I mentioned I prefer having to accept with ENTER.

The sequence to enter `\voiceOne` (example from the original issue) is now

`\vo` `TAB` `o` `ENTER`

The only missing feature now is the automatic selection upon partial completion (see https://github.com/wbsoft/frescobaldi/pull/918#issuecomment-288719192 for explanation and screenshot). I tried to do it but for some reason it didn't do what I thought ...

[...]

PR #918, opened 2017-03-22
fixes: #299 implemented autocomplete with tab key

jamesthechamp:
related to issue #299 
while typing one can use iterating through the suggestion list with TAB key 
please have a look on it.


uliska:
Thank you for the PR.
It does what it promises (which is a small but really useful improvement for daily work :-) ), and I can't find anything wrong with the code.

So I suggest @wbsoft has a final look before merging it, otherwise I'd do it in a few days.

@jamesthechamp if you wish to continue on that track (of #299) I suggest you give a try at partial autocomplete.

For example, if you type `Dr` you get `DrumStaff` and `DrumVoice` as suggestions, which are equal up to the `Drum`

autocomplete

It would be even better if in that case the `um` would be added *and selected*. Hitting `Esc` or `Backsp` would cancel the whole operation while or remove the implicitly added text. Every other letter would be added *after* the selection. So if you write `Dr` you'd get the dropdown and `um` selected, so `S` or `V` would immediately select the right suggestion.

Please give a hint on whether you'd like to do that or not because in that case we'd postpone the merge and do it all together.

---

*(Just a general remark: even when you're forking the repository it would be good to work and submit your work in separate branches instead of on `master`)

uliska:
Another (tiny but neat) improvement would be to automatically make the first element selected when the pop-up appears. Right now what happens is:

* I type `Dr`
* The suggestions pop up like in the image above
* When I hit `Enter` the first item is applied

This is actually correct behaviour but slightly incorrect display, as the first item should be highlighted already.

Consequently the first hit to TAB should not select the *first* item but the *second*.

uliska:
@jamesthechamp If you are doing this because you're interested in GSoC and haven't done so (if I'm not mistaken) please introduce yourself on the Frescobaldi mailing list (maybe referencing this PR).

Otherwise just keep working on this please :-)

jamesthechamp:
hey @uliska ,

i made changes and implemented autocomplete as you mention
now it partially autocomplete input as show following
peek 2017-03-23 18-07
and select first as popup opens as shown following
peek 2017-03-23 18-06
and sorry ... i should had made a new branch for issue i will take care of its from now on.
and sure i will shoot a mail introducing myself and my interest to project.

thanks for reviewing and providing guidance. :)

uliska:
Thank you, this is becoming really nice!

But I have a wish that will make it somewhat more intuitive to use (actually this is what I had in mind but maybe I didn' make myself clear enough).

When the pop-up appears the common text should be already in the document, but selected, looking like this:
autocomplete

* TAB can then accept this partial suggestion (basically what you already have implemented)
* ENTER will accept the first (selected) item from the dropdown
* ESC will abort the process and (probably) remove the partial suggestion (or leave it selected, I don't know and really care)
* Typing any other character would accept the partial suggestion and continue to enter text. In the case of my image typing `V` would result in this state:

autocomplete

So entering `DrumVoice` would consist of the sequence `Dr` => `V` => `Enter`

---

Of course it would be nice if (as someone suggested) the match from input to word list would be case insensitive ...

uliska:
Moved to https://github.com/wbsoft/frescobaldi/pull/922 to have it on a proper branch.

@jamesthechamp thank you for your contribution so far.

PR #922, opened 2017-04-07
WIP: #299 Improve autocompleter

uliska:
Continued from https://github.com/wbsoft/frescobaldi/pull/918 (created a new branch within the upstream repository.

Implements feature requests from #299. Started by a potential GSoC student. I hope to continue completing the wish list from the issue because I think this is a very useful improvement.

Anyone wanting to continue working on this should do it on this branch.


uliska:
OK, there are two things open now that I'm aware of:

* Upon partial autocompletion with `TAB` the added text should be selected.  
  I don't know why https://github.com/wbsoft/frescobaldi/pull/922/commits/b62e934d379b16032e3d0b7828072e848014a69e doesn't do that
* The autocompleter should *not* prevent `Alt-TAB` to perform its default operation to initiate the desktop environment's task switcher. Right now Alt TAB only aborts the completer so the *next* press will start the task switcher. The problem in https://github.com/wbsoft/frescobaldi/pull/922/commits/f128d8c46a6236451a050de644b51e3eae1c08bc is that *only* the `TAB` part is passed through and not the Alt modifier.

Additionally, I don't know for sure what using the `Ctrl`modifier for cycling backwards through suggestions does for Mac users' expectations.

uliska:
See http://stackoverflow.com/questions/43294525/how-can-i-pass-through-alttab-in-a-pyqt-event-filter for the Alt-TAB question

uliska:
OK, I *hope* this is now ready to be merged. Everything except the capturing of the Alt-TAB issue has been fixed, as far as I'm aware.

The behaviour of autocomplete now is:

* after two characters (or the manual command) the completion list pops up
* If there's a partial match it is automatically inserted but selected
* The first suggestion is selected
* typing any character will overwrite the selection  
  (this is good if you want to type something that isn't suggested)
* hitting TAB will "accept" the selection and and highlight the next suggestion
* (the Ctrl modifier will cycle backwards through the list)
* typing another character can now produce the next partial match (like above)
* At any point the currently highlighted suggestion will be accepted with ENTER

NOTE: matching is now case insensitive. That is: upon ENTER the correct capitalization is copied from the suggestion.

uliska:
Yet another improvement ;-) According to @jan-warchol 's original request typing further characters *also* accepts partial completions:

`\vo`

auto1

`\vo => f`

auto2

`\vo => f => ENTER`

auto3

Note that upon ENTER the correct capitalization is taken from the completion list.

uliska:
As I haven't got any objections (only some confirmations from users that it works), I'm going to merge this now.

This branch leaves two open ends:

* Opening the completer popup "captures" the TAB key, so `Alt-TAB` doesn't properly work to initiate the OS's task switcher. Instead it closes the autocompleter, so `Alt-TAB` has to be used *twice* to switch to another program.
* I haven't got any feedback on the question of `Ctrl-TAB` behaviour and expectations from Mac users.

Issue #917, opened 2017-03-22
Whitespace clean-up and policy


uliska:
We have lots of files with lots of lines that contain trailing or only whitespace (e.g. lines between method definitions where the editor kept the indentation).

This leads to the situation where contributors with editors that clean up files (as Frescobaldi does with LilyPond files BTW) either produce awkward commits with lots of "noise" or have to awkwardly prepare their staging to produce meaningful commits.

I would like to see

1. A full whitespace clean-up of the code base (I think this would have to include all branches?)
2. A policy for contributors that no "offending" whitespace should be introduced with new commits

With 2) I'm not sure how to enforce that. Maybe we could add a Travis test that automatically checks each push and pull request? Or a pre-receive hook that automatically cleans up and adds a "whitespace" commit if necessary?

PR #924, opened 2017-04-08
Remove (trailing) whitespace

uliska:
Relates to #917 

The sed command

for f in $(find . -print | grep \.py$); do sed -i 's/\s*$//g' $f; done

cleans all whitespace-only lines and removes trailing whitespace
throughout all the .py files in the repository

I suggest doing this for *all* currently published branches to have a consistent codebase without trailing whitespace.

The thing is only that we should provide tools to ensure that no trailing
whitespace gets back in in the future. Some editors don't clean up upon saving while others do. And particularly with Python there are often lines
filled with indents between function definitions.

I see two options:
a)
provide a pre-commit hook that cleans up *before*
committing. The problem is that this hook has to be installed manually
by any contributor (no way to automatically distribute it).
b) a pre-receive hook on Github that rejects pushes. Problem is that this may be quite off-putting for contributors.

Both could be accompanied by having a clean-up script available in the
repo that can easily be invoked, even by new contributors.

@wbsoft what do you think? I would volunteer doing this clean-up for all current branches, but it only makes sense if we establish a policy for later.

### DIFF
--- a/frescobaldi_app/about.py
+++ b/frescobaldi_app/about.py
@@ -38,40 +38,40 @@ import userguide.page
 
 class AboutDialog(QDialog):
     """A simple 'About Frescobaldi' dialog.
-    
+
     Most of the information is taken from the info module.
-    
+
     """
     def __init__(self, mainwindow):
         """Creates the about dialog. You can simply exec_() it."""
         super(AboutDialog, self).__init__(mainwindow)
-        
+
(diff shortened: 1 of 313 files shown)


uliska:
Just as I expected, there are already many conflicts arising.

I think this is too ambitious, and I'll close the PR

PR #926, opened 2017-04-11
Add library path to LilyPond invocation

uliska:
This will fix a long-standing annoyance - but is only implemented with downloaded and self-compiled versions on Linux. I assume distro packages will work as well (or don't need it anyway), but I have no idea about Mac and Win. Maybe the problem doesn't apply to them, but I want to wait for feedback on that before merging this PR.

Not adding this can cause compilation failures when the bundled versions of libraries (like Ghostscript) don't match the system-provided ones.

### DIFF
--- a/frescobaldi_app/engrave/command.py
+++ b/frescobaldi_app/engrave/command.py
@@ -84,6 +84,7 @@ def defaultJob(document, args=None):
     j.directory = os.path.dirname(filename)
     command.append(filename)
     j.command = command
+    j.environment['LD_LIBRARY_PATH'] = i.libdir()
     if s.value("no_translation", False, bool):
         j.environment['LANG'] = 'C'
         j.environment['LC_ALL'] = 'C'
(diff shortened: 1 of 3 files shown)


uliska:
There's one thing I'm not really sure about, and that's who `abscommand` and `command` relate, and if that's relevant to my patch

uliska:
See also http://lists.gnu.org/archive/html/lilypond-user/2017-04/msg00236.html for the question of other OS/installation types

uliska:
It seems the issue is no issue on Windows as probably all dependencies are bundled in the installer.

So that leaves OS X. I don't know how LilyPond is packaged/bundled there, but somehow I have the feeling it's bundled as well.

As I don't recall this error being raised by non-Linux users I think we can merge this patch and wait if any users raise the issue again.

uliska:
As I haven't got any further feedback I assume that

* this issue is only relevant for Linux
* the fix doesn't affect others either

So I'll merge this after fixing the indicated merge conflicts

PR #927, opened 2017-04-11
676: Add "Goto Line" command

### LINKED ISSUE #676: go to linenumber
reported by a user: Is there a go to line number function in the Frescobaldi editor?

I think it'd be nice to have it. It should be in the View menu, have Ctrl+G shortcut and implementation in the `mainwindow.py` file, where also methods are for search, etc.

uliska:
Fixes #676

For now the default shortcut is Ctrl+Alt+G because Ctrl+G
is already taken


uliska:
I think this is pretty uncontroversial (if I haven't overlooked any silly things and assuming no substantial UI wishes are expressed).

One could imagine a real dialog with sort of a preview (similar to the edit-in-place dialog. This could preview a range of, say, +/- 5 lines around the line entered in the chooser edit, so one sees the target before finally accepting it and be brought to that in the main editor. But I'm not inclined to do that right now, and it's in no way necessary for the wished functionality.

So if there are no objections I'll merge this myself in a few days.

uliska:
I'm merging this now. Any additional feature requests can be created as issues, and if the Default shortcut is an issue people may also complain later.

Issue #904, opened 2017-03-06
CTRL+Click to jump to includes and variables


bastl:
In other IDEs it is possible to use identifiers like hyperlinks by holding CTRL-key and clicking the name.

I'd like to use that for:  `\variable` and for `\include "myInclude.ly"`

uliska:
If I'm not completely mistaken this is exactly what Frescobaldi does.

I close this issue, but feel free to reopen it if I should have overlooked anything.

bastl:
You are right the function is working fine, I overlooked it...

But I did so because there was no visual feedback when hovering over the item with CTRL held down. I expected to see the identifiers underlined in that case as I am used to from eclipse (IIRC).

Feel free to change this into a wish if you like. At least I know how to do it now :-)

uliska:
OK, this makes sense.
It would be cool to have items highlighted on hovering, either by underlining or through a tooltip (which could also indicate the location).

I can see this as a resource problem because it would permanently have to search for targets, while the current behaviour triggers the search only upon actual user interaction. But I'm not fully aware if the document already has such a list available.

bastl:
Hmm in the source code editor component of eclipse IDE in Linux and on Windows as well this is a standard feature. 

I am relatively sure that one some how has to register where/what the identifiers are. Then the UI component will take care of the rest. Quote sinilar to Syntax Highlighting id expect.

uliska:
Sure. But we did have some performance issue with collecting the autocomplete and syntax highlighting data, so I wanted to add this disclaimer.

wenxin-bupt:
I plan to cache the information of included files in `documentinfo`. And to update them according to the changes of the text.

I found highlighting the included files or adding tooltips is related to the inline git editor technically. Because they are both adding widget into the `QPlainTextEdit` view.  

 To highlight the included files seems feasible.   But I can't find a way to integrate tooltip into `QPlainTextEdit`.  The only widget inside the `QPlainTextEdit` I found  now is the autocompleter which is built from `QCompeleter`.  Is there a way to add a widget into `QPlainTextEdit`? Otherwise I need to reconsider how to implement the inline editors.

uliska:
As long as the GSoC selection process hasn't been completed I can't comment on the Git editing widgets (in order not to give any indication about our plans or decisions in one or the other direction.

[...]

PR #935, opened 2017-04-24
Highlight the included files

wenxin-bupt:
This is my try of highlighting the included file while "Ctrl" is pressed. It highlights the file names (paths) by adding bold font and adding an underline. While highlighting,  it doesn't check whether the file is able to be opened. 

The purpose of this function is to give the user a hint when he pressed the "Ctrl" button.


uliska:
Hi Wen Xin,
thank you for this pull request!

Before investigating the code I'd like to comment on the interface itself (and I'd be happy about comments by others as well), a few glitches I noticed when trying out your branch, and some additional ideas/wishes.

##### Interface

1) The highlighting through bold and underlined font looks good to me.
2) Your suggestion triggers the highlighting through pressing the `Ctrl` key
    * IIUC this is somewhat unusual - but that doesn't say it's bad
        * Usually one would expect the highlighting to appear when the mouse is hovering over the text.
        * Actually I think it's good that one can see all possible places at a glance
        * But we should at least discuss that

#### Glitches

1) When clicking with the mouse in the editor window the highlighting is canceled.  
    But when clicking in any other place of Frescobaldi while `Ctrl` is still pressed the highlighting is *not*
    canceled and won't go away even after releasing `Ctrl`.
2) Just hitting `Ctrl` (i.e. having the include files highlighted for a time) marks the document as modified. That means when you open a document, press `Ctrl` and close the document again you are asked for saving the modified document. This is of course not acceptable.  
Also, when opening an included file through `Ctrl`-Click the newly-opened document is immediately marked as modified.

#### Additional Ideas

1) The information about the included file is already present, I think. So it should be possible to show a tooltip displaying the file location when hovering over a highlighted filename.  
I think it should be possible to capture the mouse movement (when the highlighting is active) and determine whether the mouse is over one of the highlighted names.  
(This is a "would-be-nice-to-have" wish.)
2) The "Click-to-open" functionality also works for definitions. For example if you have a file `included.ily`
```lilypond
music = { c' }
```
and a main file
```lilypond
\include "./included.ily"

\score {
  \new Staff \music
}
```
clicking on `\music` will open the included file at that position. I think this should be included as well (and was listed in the original description of #904 as well).

wenxin-bupt:
Sorry for replying late. 

I did have some problems of implementing the hovering-tooltip function.  So I tried this highlighting version. 

**For the two bugs you found**:

The first one is because of the losing focus of current document. After losing focus, the highlighted document doesn't receive the "key release" event. So the "unhighlight" function doesn't  execute. 

The second one is due to my way of adding highlight format. I directly adding the format to current content. And this is considered as changing the document.  

I'll fix these bugs.

**The "Click-to-open" functionality**：

I haven't thought about this. I assume that implementing this functionality need to visit the syntax tree (just like implementing a auto completer)?  

**This is my initial thoughts of this "hint" functionality**: 

When user's mouse hovering on a included user name, a tooltip appears with the words "  Ctrl + Click to open 'path/the_file_name.ly'  ".

Detecting the included files / paths. Highlighting the files / paths that could not be opened with the dot-underline. This function seems to belong to a syntax checker. 

It's more natural and useful.

uliska:
We have completely implemented this in the new PR #941

PR #941, opened 2017-05-09
Add Ctrl + mouse hovering to display tooltips for include files

wenxin-bupt:
I write this new version.

 A user will get a tooltip displaying the exact path of a include file when the mouse is hovering near the include file with the Ctrl button pressed. 

For invalid files it will display "This is an invalid include file"

### DIFF
--- a/frescobaldi_app/open_file_at_cursor.py
+++ b/frescobaldi_app/open_file_at_cursor.py
@@ -19,17 +19,88 @@
 
 """
 Opens a file the current textcursor points at (or has selected).
+
+Generate the include files' tooltips
 """
 
 
 import os
 
-from PyQt5.QtCore import QUrl
+from PyQt5.QtCore import QUrl, Qt
+from PyQt5.QtGui import QTextCharFormat, QFont, QTextCursor  
 
 import documentinfo
 import browseriface
 
+def genToolTipInfo(cursor):
+    """Return a list of infomation of the inlude files.
+    
+    For each include file,  generate a ToolTipInfo instance. Insert it into the list. 
+    
+    A ToolTipInfo instance has two attribute:
+    
+    - content: the content of the tooltip for this include file
+    
+    - num:     number of the block where this include file locates 
 
+    """    
+    dinfo = documentinfo.info(cursor.document()) 
+    
+    # get the path of current file
+    filename = cursor.document().url().toLocalFile()
+    directory = os.path.dirname(filename)
+    if filename:
+        path = [directory]
+    else:
+        path = []
+    path.extend(dinfo.includepath())
+    
+    # structure being used to store included file information
+    class ToolTipInfo:
+        pass    
+    toolTipInfo = []
+    
+    # startIndex is the content where every find begins
+    startIndex = 0
+    
+    # find the first '\include'
+    pointCursor = cursor.document().find('\include', startIndex)
+    
+    # find an include file each loop
+    while not pointCursor.isNull():
+        block = pointCursor.block()
+        text = block.text()
+        head = block.position()
+        tail = head + block.length()
+        i = dinfo.lydocinfo().range(head, tail)
+        fnames = i.include_args() or i.scheme_load_args()
+        
+        # update startIndex
+        startIndex = tail
+        if not fnames:
+            continue
+        
+        # file information unit
+        info = ToolTipInfo()
+        info.num = block.blockNumber()
+        
+        # whether we can find a valid file path
+        valid = False
+        for f in fnames:
+            for p in path:
+                name = os.path.normpath(os.path.join(p, f))
+                if os.access(name, os.R_OK):
+                    info.content = name
+                    valid = True
+                    break
+        if valid == False:
+            info.content = "This is an invalid include file"
+        toolTipInfo.append(info)
+        
+        # find next '\inluded'
+        pointCursor = cursor.document().find('\include', startIndex)
+    return toolTipInfo        
+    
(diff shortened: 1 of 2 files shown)


uliska:
Thank you for this PR which looks really nice.

I had a quick glimpse on the code and have some ideas where I expect some questions, but I'll have to look more closely into it.

In the UI it basically does what it promises, with one tiny wish left open: When I move the mouse over the filename and click `Ctrl` nothing happens - I then have to move the mouse to trigger the tooltip. So with your current implementation you should try to manually trigger the event from within the keypress event.

As a discussion point: there is also a tooltip event. One could equally connect to that and check if there's the Ctrl key pressed and an include under the cursor. I don't have a preference here, just wanted to mention it.

From the perspective of least surprise for the user it would be good to have the functionality of your other PR included as well: highlight the words as bold underlined as soon as Ctrl is pressed. That is, if you manage to do that without setting the document status to modified.

Please say if you want to integrate both things here. Otherwise we can finish and merge the functionality you have in the current PR first.

wenxin-bupt:
I have figured out a way to display the include files in a different style without showing the file is modified. However, I think it's kind of hacking style. We can add it until I find a more appropriate way.

Personally, I think this PR is less surprising (or annoying) than the highlight version.  Since a user doesn't get the hints every time he presses the Ctrl button.  I'd like to only merge this PR now.  

> As a discussion point: there is also a tooltip event. One could equally connect to that and check if there's the Ctrl key pressed and an include under the cursor.

And... I don't figure out your meaning here.

uliska:
> I have figured out a way to display the include files in a different style without showing the file is modified. However, I think it's kind of hacking style. We can add it until I find a more appropriate way.

Am I right that this solution is *not* included in your PR yet?

> Personally, I think this PR is less surprising (or annoying) than the highlight version. Since a user doesn't get the hints every time he presses the Ctrl button. I'd like to only merge this PR now.

OK, we can leave it at that for now. If people should start complaining we may consider a change, but I don't expect this to happen.

>> As a discussion point: there is also a tooltip event. One could equally connect to that and check if there's the Ctrl key pressed and an include under the cursor.

> And... I don't figure out your meaning here.

Right now you're intercepting the "KeyPress" event and check whether some conditions are met. There's also an event that is triggered when the tooltip is about to be displayed. This could also be used as the hook to check the conditions. But we don't need to do that.

I have another idea, but I'll try it out and send it to you as a a PR if I'm happy with it.

wenxin-bupt:
> Am I right that this solution is not included in your PR yet?

Yes, it is not included in this PR till now.  

> Right now you're intercepting the "KeyPress" event and check whether some conditions are met. There's also an event that is triggered when the tooltip is about to be displayed. This could also be used as the hook to check the conditions. But we don't need to do that.

I understand it. 

In this way we add a tooltip for the entire editor and we check the mouse position when the tooltip is triggered. If it meets the condition, we give the tooltip a specific content and the position it will display.  If not, we will hide the tooltip. 

And I agree we don't need to do it in this way.

wenxin-bupt:
> However, there's one thing I'd like to change right now: since we are checking the MouseMove event anyway I think it would be even more useful if the tooltip would show up without having the Ctrl key pressed.

Ctrl button in my PR is a trigger for updating the include files position list.  If we don't change the logic (generate a list) that I am using now,  we need to find a new trigger to update the info list.  I think there could be efficiency problems.   But since your have a new idea of showing the tooltip without update the whole info list, there seems no need to worry.

I have also found a problem in my code. I named variables in Camel Case style. For example,  `startIndex`  which should be name as `startindex`. That should be fixed latter.

> As I have already started looking into it may I suggest that I try to achieve what I have in mind and break it down in small commits discussing what I mean? This will not be a final implementation but a next state for discussion.

Looking forward to learning your idea through these commits! I will add my fixing commit after them.

uliska:
> Ctrl button in my PR is a trigger for updating the include files position list. If we don't change the logic (generate a list) that I am using now, we need to find a new trigger to update the info list. I think there could be efficiency problems. But since your have a new idea of showing the tooltip without update the whole info list, there seems no need to worry.

I have seen that, and I'm aware that we have to be careful when doing this from the usual mouseMove event. But I hope my approach is lightweight enough not to cause problems:

- it is only triggered when the mouse moves to a current line or after the document has been modified
- we (nearly) look only on the current line and not the whole document.

I have nearly finished what I wanted to achieve, with the exception of the issue that we're always working on the whole `block` instead of the actual mouse position. It will be very simple to implement but I'd like to wait for feedback on #947.

If you (generally or completely) agree with what I have added the next thing would be to remove the `open_file_at_cursor.genToolTipInfo()` function (sorry ...)

wenxin-bupt:
Great improvement!

There is an obvious improvement in my view. In my PR, the tooltip will still be hanging there for some seconds when user's mouse has left the block.  In your commits, when the mouse left the block which contains include file, the tooltip is hided immediately.  

I ran several tests to compare the time usage of my version and your version by simply adding a timer in `mouseMoveEvent`.  

The time cost of running the check-and-show tootip code (calling super class's function is not included) is mostly 10^-5 s for both two versions. The time cost of updating my file-info list is 1-100 ms. When the mouse is moving in the area of "\include",  time cost both increases.  To generate a tooltip, my version need 1-10 ms in most of time, sometimes it increases to 10-100 ms. While the time cost in your version keeps in range 1-10 ms.  Your version is at least as efficient as mine according this test.

I completely agree with what you have added. It's much nicer.

PR #943, opened 2017-05-11
Score Wizard: Provide options for landscape/rotated

### LINKED ISSUE #768: Score wizard: distinguish ‘proper’ and ‘rotated’ landscape formats
There has been a small discussion on page orientation in the thread starting from
http://www.mail-archive.com/bug-lilypond@gnu.org/msg40138.html.
I’m quite sure that I learned about the
   `#(set-paper-size "a6" 'landscape)`
syntax through the Frescobaldi score wizard, and it turns out that the more reliable way to invoke landscape formats, especially WRT output formats other than .pdf, is
   `#(set-paper-size "a6landscape")`

I can’t completely wrap my head around the essence of this distinction, but I think it would be better to use the latter in the score wizard, or provide more sophisticated support allowing a choice between both. Documentation is in [NR 4.1.2](http://lilypond.org/doc/v2.19/Documentation/notation/paper-size-and-automatic-scaling).

uliska:
Closes #768

### DIFF
--- a/frescobaldi_app/scorewiz/build.py
+++ b/frescobaldi_app/scorewiz/build.py
@@ -172,7 +172,8 @@ class Builder(object):
         self.smartNeutralDirection = generalPreferences.neutdir.isChecked()
         self.showMetronomeMark = generalPreferences.metro.isChecked()
         self.paperSize = generalPreferences.getPaperSize()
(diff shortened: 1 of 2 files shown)


uliska:
As I didn't get any negative feedback, even after asking on the mailing lis,t I decided to merge.

Issue #724, opened 2015-07-16
Add support for -dpreview mode


sincere-music:
Currently, in order to engrave using -dpreview one has to use ‘Engrave (custom)’ and manually edit the command. What’s more, there is a naming conflict with the ‘Engrave (preview)’ mode, which actually refers to Lilypond’s default with point-and-click.
I suggest renaming the current ‘Engrave (preview)’ mode to ‘Engrave (default)’ and add a new mode ‘Engrave (preview)’, which calls
$lilypond -dpreview $filename.

uliska:
I don't think this warrants a new engraving mode altogether.

What I suggest is adding a checkbox to the 'Engrave (custom)' dialog to make that mode accessible.

Renaming things could be discussed, although I'd also consider choosing a different name for the `-dpreview` interface. Actually I find this LilyPond option somewhat misnamed as it isn't really a preview but closer to what an incipit does. So we could label the checkbox for example "Engrave first system only" and add the reference to the `-dpreview` option to the tooltip.

uliska:
@sincere-music please test Pull Request #944 and discuss the namings.

uliska:
I've now merged #944 but I'm only partially satisfied.

There is now the option to choose "First system only" in the Engrave mode dropdown in the Engrave (Custom) dialog. But having this as a GUI element makes the limitation more problematic that the resulting single-line file isn't present in Fresocbaldi's list of produced files.

PR #944, opened 2017-05-11
WIP: Add "First system only" option to Engrave (Custom)

uliska:
This is a preliminary Pull Request to have people test the feature and discuss before finishing it.

Addresses #724

Adds an entry "First System Only" to the "Engraving Mode" combo in the Engrave (Custom) dialog.
Please test and discuss the naming conventions used.

**Missing:** The compiled preview is *not* displayed in the Music View and only available through `LilyPond=>Generated Files`. A quick look into the code didn't tell me how to fix that as well.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,8 +2,10 @@ ChangeLog of Frescobaldi, http://www.frescobaldi.org/
 =====================================================
 
 
-Changes in 3.0.1 -- 
+Changes in 3.0.1 --
 
+* New features:
+  - New "First System Only" option in Custom Engrave
 * Bug fixes:
   - fixed #895 seeking in MIDI player during playing stops sound
 * Improvements:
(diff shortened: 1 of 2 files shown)


fedelibre:
Just tested, nice!

I don't miss too much the fact that the compiled preview does not appear in the Music View. After all, these preview files are usually produced just to share the files with other people (typically for bug reports) and not for the real music preview.

fedelibre:
BTW, @uliska I've added some comments to this issue about -dpreview unwanted cuts off:
https://sourceforge.net/p/testlilyissues/issues/2968/

uliska:
> I don't miss too much the fact that the compiled preview does not appear in the Music View. After all, these preview files are usually produced just to share the files with other people (typically for bug reports) and not for the real music preview.

That's true, but I *do* think the preview should be considered a compilation result because

* Usually I (and presumably most users) want to review the preview before sharing. Right now I can't even open it through the "generated files" menu.
* If I change the file before compiling the preview the score in the Music View isn't up-to-date, which can be very confusing

uliska:
OK, as I currently don't have an idea how to fix the remaining problem (and as it's not a regression through my patch but has always been the case) I will give this a day or another and then merge if noone objects.

PR #949, opened 2017-05-22
WIP: UrlRequester additions

uliska:
Two (non-breaking) additions to the UrlRequester (which I need for something else):

* `editingFinished` signal  
Allows to perform validations that don't interfere with manual typing in the line edit (i.e. only validate after it loses focus)
* `mustExist` property  
If set to `True` the UrlRequester rejects entries that don't point to an existing file/directory (depending on `fileMode()`. In that case a message box is displayed and focus returned to the line edit. Empty paths are allowed, though

### DIFF
--- a/frescobaldi_app/widgets/urlrequester.py
+++ b/frescobaldi_app/widgets/urlrequester.py
@@ -21,10 +21,11 @@
 UrlRequester, a lineedit with a Browse-button.
 """
 
+import os
 
 from PyQt5.QtCore import pyqtSignal
 from PyQt5.QtWidgets import (
-    QFileDialog, QHBoxLayout, QLineEdit, QToolButton, QWidget)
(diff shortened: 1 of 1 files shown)


uliska:
I realized that this PR isn't ready to merge yet (and I'd be glad about any assistance).
The `editingFinished` signal is also emitted when the tool button is pressed, but this should not take place. The reason is that editingFinished is used for validation, and when I manually edit the line edit and then click on the tool button I do *not* want such validation to occur.

In other words: changing the line edit text and firing up the browse dialog should be considered *one* editing stage.

See also my question on [stackoverflow](http://stackoverflow.com/questions/44108178/suppress-qlineedit-editingfinished-signal-when-certain-button-is-clicked)

uliska:
I've now found the solution (setting a flag when clicking on the file open button),
and the improvements are now implemented as part of #1075, in 81cc2263623aaf3e1af25eef63b22739f208788b

PR #974, opened 2017-06-12
Rewrite vcs

wenxin-bupt:


wenxin-bupt:
2017/6/12

1.  Investigate git-version's format and Python's PEP 440
     https://www.python.org/dev/peps/pep-0440/
2.  Implement the version() in gitjob
3.  Investigate how to fetch the committed file and the file in Index
    
    Command:
    Index:                   **git cat-file --filters :[*file's relative path*]**
    specific commit:   **git cat-file --filters [*commit*]:[*file's relative path*]**

    Reference:
    https://stackoverflow.com/questions/4044368/what-does-tree-ish-mean-in-git
    https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html#_specifying_revisions

4. Implement the _read_committed_file()

wenxin-bupt:
In the last commits above:

1. Error or exception handling is waiting to be implemented. 
  I need some time to investigate into this topic after finish implementing basic functions.  
  I agree that we shouldn't simply raise an exception. Exceptions should lead to disability of the git functions and be transparent to users in most of the time.

2. I don't know how Frescobaldi and Lilypond handle the line ending (CRLF?  LF? depending on the OS?)

3. Till now all the functions are running synchronously.

uliska:
Thank you for opening the pull request. In fact this is the right infrastructure to discuss the code (better than individual issues), and I now get notifications.

I'm sorry to say that I will not have the opportunity to look into the code earlier than tomorrow evening/night (my kids have holidays, my wife not :-( ) but I'll catch up ASAP.

wenxin-bupt:
Have a good time!

For the question 2 above,  I've found this in `encodedText()` in `document.py`
```
def encodedText(self):
        """Return the text of the document as a bytes string encoded in the
        correct encoding.
        The line separator is '\\n' on Unix/Linux/Mac OS X, '\\r\\n' on Windows.
        Useful to save to a file.
        """
```
I think I should treat line endings depending on the OS when writing content into temp files.

wenxin-bupt:
2017/6/15

 1. I investigated how to get a file's git status. I pick two commands:
     Tell whether a file is git tracked.
     `git ls-files [FILE_PATH]`
     Get file's status: 
     `git status -p --ignored  [FILE_PATH]`
     
     With these two commands, we can tell the file's status of
     `new added`,  `staged`, `staged modified`, `committed`, `modified`, `untracked`, `ignored` 
     
    I haven't finished the function.

frescobaldi_app/vcs/gitrepoo.py
```diff
@@ -0,0 +1,416 @@
...
+       
+
+    def _is_git_path(self, path):
+        """
+        Return True if 'path' is a valid git working tree.
+        """
+        return path and os.path.exists(os.path.join(path, '.git'))
+    
+
+    def _get_work_path(self, ori_path):
```
uliska: I got stuck understanding this function. I'm sure there's something wrong, but I don't really understand *what* it should exactly do. (This is why I haven't got further in reviewing your code so far).

What is the intended return? I understand that - assuming the given path points to a file or directory in the working tree - 

* `result[0]` is the absolute path to the repository root
* `result[1]` is the relative path to the file/dir object

Is that right?

I'll add a comment below to indicate where the problem is.

frescobaldi_app/vcs/gitrepoo.py
```diff
@@ -0,0 +1,416 @@
...
+
+        Arguments:
+            ori_path (string): a absolute path.
+
+        Returns:
+            (root path of the working proj, relative path)
+            (None, None) for non-git-initilized file_path
+        """
+        if ori_path and os.path.exists(ori_path):
+            if os.path.isdir(ori_path):
```
uliska: This doesn't really work. `os.path.isdir` doesn't care about the presence or absence of a trailing (back)slash while your logic seems to rely on that difference.

If I understand better what the caller of the function actually expects I can make a better comment.

frescobaldi_app/vcs/gitrepoo.py
wenxin-bupt: Yes, and here are the examples:
1. origin_path:  `'/home/pysrc/git/frescobaldi/frescobaldi/frescobaldi_app/vcs/test.py'`
   result:  `('/home/pysrc/git/frescobaldi/frescobaldi', 'frescobaldi_app/vcs/test.py')`
   `'/home/pysrc/git/frescobaldi/frescobaldi'` is the absolute path to the repository root
  ` 'frescobaldi_app/vcs/test.py'` is the realative path to the file
   Because setting Git's working environment need the repository's root. Operating in a Git command need file's relative path to the repo root.

2.  origin_path: `'/home/pysrc/git/frescobaldi/frescobaldi/frescobaldi_app/vcs'`
     result: `('/home/pysrc/git/frescobaldi/frescobaldi', 'frescobaldi_app/vcs')`
     This time a folder path is passed in.  This is the way we are using vcs now.
      `'/home/pysrc/git/frescobaldi/frescobaldi'` is the absolute path to the repository root
      `'frescobaldi_app/vcs'` is the realative path to the folder

3.  origin_path:  `'/home/pysrc/git/frescobaldi/frescobaldi/.git/index'`
     result: `(None, None)`
     This is a subfolder in  `.git` folder. And it's not Git tracked. 
 
4.  origin_path: `'/home/pysrc/git/frescobaldi/frescobaldi'`
     result: `('/home/pysrc/git/frescobaldi/frescobaldi', '.')`
     When a repository's root path is passed, the relative_path part is a `'.'`

uliska: Thank you for the explanation, now I see. Your function would then fail on e.g. `'/home/pysrc/git/frescobaldi/frescobaldi/frescobaldi_app/vcs/'` because it would be treated as a file with empty filename.

I've added a small fix to it and will push in a minute.

frescobaldi_app/vcs/gitrepoo.py
wenxin-bupt: Thank you for this fix. I tested and found the problem that you mentioned.

wenxin-bupt:
I just added an attribute `target_is_file`. I want to make it more readable. 
E.g.  when you call it from an instance named `repo`: `repo.target_is_file` (bool) 
It seems more easy to read for me. It's also a strange name... 

But if it is named `is_file`, the code above will be `repo.is_file`. I think this might cause misunderstanding. Because this attribute is used to mark whether the given path is a file path or not. And of course the repository can't be a file.

wenxin-bupt:
Most features have been implemented so far (except _stage_ and _commit_).  I've nearly finished  the first step of finding the right git command,  shelling out and parsing the output into the format we want. Next, I want to improve these features until it can be used as a qualified module.

There are three issues that I think I need to tackle.

1. Solve Encoding, decoding problems of  the working file and git output
      I haven't investigate this issue. Currently I encode the git output into string using 'utf-8'. 
  
2. Rewrite many functions to make them can execute asynchronously
    For the convenience of test and development. All of the functions are running synchronously now.
While to provide the option of running git command asynchronously is a main design objective of this module.  

3. Provide exception handling
    Both the exception handling scheme in module level and handling normal exceptions in code details.

uliska:
Finally I managed to read all your code. I'm very happy with what I see, it is all well organized, understandable and seems to make sense. However, as -- if I'm not completely mistaken -- the `gitrepoo.py` module is still not attached to anything I have some difficulties assessing how it will all work together. So basically I just read through the code and tried to understand it. I have added a few question comments in the source and added a few smaller fixed in commits. Below you'll find the questions I noted while going through the code.

As said it's a little bit difficult without the chance to see the code "in action", so some of the questions may be pointless, while many other questions will probably only arise when we actually put it all together:


* gitrepoo.py is currently not "hooked in", isn't it?  
  If that's true, how is it currently tested?
* How will external changes to the Git repository be detected
  (so an external Git operation will affect Frescobaldi's diff)?
* There are questions which I don't really see yet, we'll have to discuss them
  when the code will actually run inside Frescobaldi
* Will the calls to git.run_blocking() be considered to be changed
  to asynchronous behaviour?
* The Git functions currently expect some string return value and presumably
  are still based on the implementation that raises an error if anything is
  written on stderr.  
  Will this be an issue if we update the behaviour of the Git job?
* Unification of line endings -  
  needs research/discussion
* I think the `_update_XXX_file()` functions should be merged into one core
  function and three wrappers. The line ending and exception handling code will
  be doubled too, so this will be useful.
* `_run_diff`
    * is this intended to compare for example the temp file to the working tree?
    * What do you mean with "Whiteblocks handling"?

* I'm not totally sure how the asynchronous Git.run() method interacts 
  with the system. 
    * What I'm missing is the emission of signals. In typical Qt coding the user
      of the Git object (i.e. Gitrepo and eventually the code in the editor) 
      will connect to the Git() object's signals and implement slots.
    * I'm not sure if the `receiver` argument is intended to provide such 
      functionality and if everything's correct here. Maybe it is, maybe not.
    * Is it really the right choice to handle the queue such that subsequent commands will start only after the previous have completed? Isn't that contradicting the asynchronous behaviour?  
    I'm not completely sure but wouldn't it be appropriate if the Git object can hold multiple running process objects and have the caller take responsibility of connecting to their signals?

      

**Note:** I'll still have to study `gitrepo.Repo.extract_linenum_diff` and `gitrepo.Repo.extract_diff_hunk` in gitrepo!

uliska:
Oh, I just saw your edits right now (was just checking to be sure before writing an email ...).

It turns out that if you *edit* a comment of mine there will be no trace, I don't get a notification *and* there isn't any entry in Github's activity log. So it would be better to add new comments and use inline quotes of the relevant parts.

I'll go through your replies in a moment.

uliska:
*Re: testing*  
OK, I see. Yes, it would be nice to share that `test.py` with me (but not committing it. You can either send it by email or write it as a comment here).

---

*Re: watching for changes*  
Doesn't using git commands change some files in the `.git` directory? If so then it would probably be more elegant to use [`QFileSystemWatcher`](http://doc.qt.io/qt-5/qfilesystemwatcher.html) like it is done in `documentwatcher.py` and `externalchanges/__init__.py`

---

*Re: git diff*  
I think we *should* work on this, but not at the moment. This is an optional feature if everything else is there, maybe for after GSoC. (#976)

---

*Re: queue etc.*  
OK, there are several aspects here.

1. I see now where the queue is necessary for. (This is something that in e.g. Node.js would be handled with chained Promise functions.)
2. There *are* functions that should *not* be chained in that way but run independently. But of course it is possible to either create multiple `Git()` objects in that case or first kill the current job before starting the new one.
3. The callbacks (receiver and _update_signal). I'm somewhat unsure if this is right. I have the feeling that using callbacks the way you do it is not really idiomatic for Python and especially Qt. But I'm not totally sure about that, and it will surely *do* work the way you plan it. However, as far as I can see in a Qt application you would - inside the `run` method - connect some member functions to the process's signals and inside these slots emit the Git job's own signals. The main advantage that I can see (apart from adhering to the framework's conventions) is that `QProcess` has several signals (http://doc.qt.io/qt-5/qprocess.html#signals) and by connecting to them you can catch any of them and not only one through the callback function

wenxin-bupt:
OK, I see.

I investigated into encoding issues yesterday and the day before.  I have to say I'm still not very clear about how Git encoding its output. On my windows 10 system, the shell is using GB2312 encoding, a encoding for Simplified Chinese.  The `git diff` command actually outputs in 'utf-8' in the shell.  It can't decode the changes written in GB2312 correctly. But it can decode 'utf-8' changes.  I think it's very weird. I don't know why till now.  Since ASCII code is supported by any encoding (I assume). And code  “@@ -3,2 +4,0 @@” and "+, -, \n, \r" are totally ASCII code. So we can extract `git diff` output with in encoding. It won't affect the diff hunk extraction. And then we should decode the extracted code into raw bytes and then encode them with the encoding Frescobaldi use for display.

wenxin-bupt:
Just a bug report:
I can't access lilypond.org.  Has the site crashed?

uliska:
It seems like it.

But this is not the place for such reports, this should go to the lilypond-user list.
As has already happened today: http://lists.gnu.org/archive/html/lilypond-user/2017-06/msg00352.html

wenxin-bupt:
> The callbacks (receiver and _update_signal). I'm somewhat unsure if this is right. I have the feeling that using callbacks the way you do it is not really idiomatic for Python and especially Qt. But I'm not totally sure about that, and it will surely do work the way you plan it. However, as far as I can see in a Qt application you would - inside the run method - connect some member functions to the process's signals and inside these slots emit the Git job's own signals. The main advantage that I can see (apart from adhering to the framework's conventions) is that QProcess has several signals (http://doc.qt.io/qt-5/qprocess.html#signals) and by connecting to them you can catch any of them and not only one through the callback function

I’obsessed with this question... And I can't find a better way to implement it myself.  In `Git.run()`, I'm using `QProcess`'s  signal: `finished()` to activate the provided `receiver()` function.  And I'm using signal `errorOccurred()` in exception handling. `QProcess` provides six signals:
- errorOccurred()
- finished()
- readyReadStandardError()
- readyReadStandardOutput()
- started()
- stateChanged()
I think we only need `finished()` and `errorOccurred()` here.  Therefore, maybe pass a function `handler()` along with `receiver()` to handle the `errorOccurred()` signal?

uliska:
It's too late in the evening now to look closely into this issue, but tomorrow I will come up with something, either a detailed description or a commit for you to review (in the form of a Pull Request against your branch).

I think we will need *all* the signals, but probably not within the scope of your GSoC project. At least when we will communicate with a remote server (for extending Git functionality to collaboration features or when using Git to auto-update Frescobaldi or openLIlyLib package) there may be the need to communicate with the process when the user has set up the system or Git server that they have to type passwords.. I think that's why Qt comes wtih this signal/slot concept (which it explicitly opposes to callback functions).

But as said, more to come tomorrow.

wenxin-bupt:
OK,  I'm going to implement `git info` today. And one more question, where should I place the `Repo()` instance? I'm testing it in `view.py` by adding a `Repo()` instance as its attribute.

uliska:
Just to be sure: your plan is to have one Repo() for each document opened in the editor?

wenxin-bupt:
Yes.

uliska:
OK.
Yes, `Repo()` should be a member of `View()`. I would put `self._repo = vcs.gitrepo.Repo(document)` in the `__init__()`  function and update `Repo`'s constructor accordingly (so it immediately sets the paths from the `document` argument).

Actually it will have to be somewhat different (see below), but I think you could start with this straightforward implementation.

Theoretically we will want to support arbitrary VCS systems, so we should take care of this at this place. So actually in `View`'s constructure we should call an `abstractrepo.Repo.instance()` class method. This should

* check which VCS's are available on the system (there should be a `vcs.availabable()` method accompanying `vcs.is_available()`)
* then check if the loaded document is controlled by one of these
* return an instance of the appropriate `Repo` subclass

What has to be done if the document isn't in a repository (or no VCS is available)?
For now this should result in a state where the Git functionality is just not available to the user. But I could imagine that (once your work-stage-commit cycle is complete) we might provide parts of the functionality even without VCS support (having the diff against the saved state of a document is useful as an improved "undo" mechanism in its own right).

uliska:
Sorry for raising yet another design question: When starting to look into the `gitrepoo.py` code I'm tempted to suggest yet another redistribution fo responsibilities:

This "Repo()" actually doesn't represent the *repository* but rather *one file* within that repository. It somehow seems strange to do everything at this level, e.g. calculating the branches.

Maybe we should have one `repositories` object, maybe as a member of the `mainwindow`. This manages a set of `repo` objects, each of which can be a `<vcstype>.Repo()` instance. This Repo will then manage the global properties of a repository, e.g. branches, root directory, remotes etc. Whenever a document is loaded into a view we would first check if it is part of an repository that is already managed and add it to that instance, or create a new Repo instance to be added to the set.

The *document*s Git object would then call Git commands for anything related to itself but would reuse the data from the parenting Repo object.

Please discuss if anything is unclear or if you have objectsions, @wenxin-bupt .

wenxin-bupt:
Yes, this structure is clear to understand.

> Our task will now be to create a new class for documents and decide which of the methods from gitrepoo.py will stay in the Repo class and which will be moved to the Document class.

As I mentioned in the [#981 Comment](https://github.com/wbsoft/frescobaldi/pull/981#issuecomment-311962562).  I'm wondering what's your opinion of the interaction pattern with `Git()`?   

Since:
> Both Repo() and Document() will make use of the gitjob.Git() object to perform their tasks.
 If we choose to use command queue, will `Repo()` and `Document()` both have its own command queue?

uliska:
I'm afraid you'll have to wait for detailed replies until (your) tomorrow. I'm out of home

uliska:
> If we choose to use command queue, will Repo() and Document() both have its own command queue?

When would the Repo() make use of a command queue for example?

If really both objects need a queue then we'll create a dedicated object for that and simply add that to both using classes.

wenxin-bupt:
All the new added commits prefixed by `Repo()` and `Document()` haven't been tested.

wenxin-bupt:
Now new `gitrepo.py` is able to support the old functionalities in Frescobaldi.

wenxin-bupt:
Now `gitdoc.py` can be tested by adding the following code in `view.py`'s `__init__` function

```python
        import os
        from vcs import gitdoc, gitrepo

        def is_git_path(path):
            return path and os.path.exists(os.path.join(path, '.git'))

        def extract_paths(ori_path):
            # ensure there is no trailing slash
            sep = os.path.altsep if os.path.altsep else ''
            sep += os.path.sep
            ori_path = ori_path.rstrip(sep)
            if ori_path and os.path.exists(ori_path):
                if os.path.isdir(ori_path):
                    _, name = os.path.split(ori_path)
                    path = ori_path
                else:
                    path, name = os.path.split(ori_path)

                # files within '.git' path are not part of a work tree
                while path and name and name != '.git':
                    if is_git_path(path):
                        return (path, os.path.relpath(
                            ori_path, path).replace('\\', '/'))
                    path, name = os.path.split(path)

            return (None, None)

        def print_status(doc):
            print(doc.status())

        def print_diff(doc):
            print(doc.diff_lines())
            # extract the diff hunk around line 10
            print(doc.diff_hunk(10))

        root_path, relative_path = extract_paths(self.document().url().path())

        if root_path is not None:
            self.repo = gitrepo.Repo(root_path)
            self.doc = gitdoc.Document(self.repo, relative_path, self)
            self.doc.diff_updated.connect(print_diff)
            self.doc.status_updated.connect(print_status)
            self.textChanged.connect(lambda: self.doc.update(fileChanged = True))
            self.repo.repoChanged.connect(lambda: self.doc.update(repoChanged = True))
```

wenxin-bupt:
@uliska  The vcs now display git status in bottom and line diff in gutter. But it still have some bugs. Could you please give me some suggestions?

uliska:
I will look into it tonight.

uliska:
I will test it (European) tonight.

wenxin-bupt:
Thank you. I'm reading the code in `widgets/folding.py.` Because it folds blocks of code in the editor when the user click on the sidebar. I assume it would help me find the way of implementing an inline editor. Could you give me a hint of how to implement it?

uliska:
Could you please clarify a bit what you mean? I.e. what kind of inline editor in which specific case?
-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

wenxin-bupt:
When we talk about the diff editor, we can pop out a widget contains the diff contents: 


But here I want to directly display (or draw ) the diff contents inside the file's window with a different background color. 

I think this picture explains my idea:
image

wenxin-bupt:
@uliska  I decide to pop out a window to display the diff contents. 
When there are many lines of code to be shown,  we can set a scrollbar in the window. If we display these code inline, it would take up a lot of space.
image

wenxin-bupt:
Hi @uliska ,
last week I investigated how to display diff contents in the popup window and I've made some progress. 

Here I have two versions of diff displaying style. 

The first one is split the popup window into two parts. Display the original contents in the left part and the changed contents in the right part. Just like in github
image

Another version is displaying compound diff contents.  For example, I changed "best_ratio" to "normal_ratio". It only displays one line, highlights the deleted part with red background color and highlights the inserted part with green background color.

image

Version 2 seem more popular among text editors and IDEs.

wenxin-bupt:
image

This is a testing popup window displaying the diff contents. You can open this window by clicking on the diff symbol in the left gutter.

However, I think it's very ugly... 

I'm implementing the split popup window like this:
image

wenxin-bupt:
image

This test-version should be better.

uliska:
Hi Xin,

finally getting to look into your code again. When I'm opening Frescobaldi what I see functional is the two-column pop-up with the diff after right-clicking on the gutter symbol shown in the screenshot above. 

Is that correct or am I missing some functionality that is already there?
I *think* at least I would expect this to pop-up after *left*-clicking on the symbol. Except if you plan to have a context menu pop up above the diff display.
This works nicely and really fast, so I'm very happy with your progress!

Before (or while) I start reading through the code, could you please tell me what you think is possible to wrap up so we have something to merge by the end of the project?

I think this diff display is already a really nice addition to Frescobaldi, and I know that there is a *lot* of code ready to be used under the hood (so it will be very nice to build upon your work, a much better situation than before the project), but it would be really good if we could have something so the user has at least one action that can be performed. 
The very least I think should be done to make it a usable function is the ability to discard the selected hunk, i.e. revert it to Index state. This would be equivalent to an improved selective undo function and already a great thing. But let me hear what you think you may complete in the remaining time.

wenxin-bupt:
I'm very happy to have you back!

I want to make clear that this popup window is just a test for the continued development. I pushed these "immature" commits so that you can check my progress. 

 Functionalities like "discard", "partial stage" both need to display the diff contents. So I implement the "displaying diff contents" first. And then there will have a button in the popup window providing the "discard" operation.  And a new popup window providing partial stage functionalities.

>  Is that correct or am I missing some functionality that is already there?

"symbols in the gutter" and "the popup window" are the functionalities can be seen in Frescobaldi.   So you are not missing functionalities.  Actually, I open the popup window by double left clicking the gutter. I was focusing on displaying the diff contents. I implemented clicking operation to help the test.

>Before (or while) I start reading through the code, could you please tell me what you think is possible to wrap up so we have something to merge by the end of the project?

I think I can finish implementing the "discard" window and "partial stage" window by the end of the GSoC.  But I think at that time I can only offer the diff of "working vs HEAD". And there are still many works of revising the current code need to be finished before the GSoC ending.

wenxin-bupt:
My thoughts of the "partial stage" window:

At first, I thought we can provide "discard" and "stage" operation for a diff hunk in one window. When we discard a changed hunk and revert it back to HEAD state, it works immediately and it won't affect other hunks. However, unlike "discard",  when a user want "partial stage", he will "partial stage" one or several hunks. Listing all the changed hunks in a window and let the user to choose the hunks he want to stage is the right solution.

uliska:
> I'm very happy to have you back!

I'm sorry I couldn't spend more time on the project recently. But I also have to say that from what I last had seen in the code and from observing the commit notifications by email I had the impression that you are able to work quite independently and don't need much hand-holding anymore.

>I want to make clear that this popup window is just a test for the continued development. I pushed these "immature" commits so that you can check my progress.

OK, acknowledged. But it already looks very nice!

> Functionalities like "discard", "partial stage" both need to display the diff contents. So I implement the "displaying diff contents" first. 

OK, I'm happy to see that it works anyway.

> And then there will have a button in the popup window providing the "discard" operation. And a new popup window providing partial stage functionalities.

> At first, I thought we can provide "discard" and "stage" operation for a diff hunk in one window. When we discard a changed hunk and revert it back to HEAD state, it works immediately and it won't affect other hunks. However, unlike "discard", when a user want "partial stage", he will "partial stage" one or several hunks. Listing all the changed hunks in a window and let the user choose the hunks he want to stage is the right solution.

I'm not sure I understand you correctly here. When you say "partial" do you mean a part of the *document* or a part of the *hunk*? I mean, it is possible to stage or discard only *some* of the changes within a hunk. Are you talking about this or about the fact that a single hunk may only be part of the changes in a document?

From what I can see it would be great if from the popup window with the diff I could either stage or discard the given hunk, leaving the document of course in an inconsistent state of partial proessing. I don't see why staging such a single hunk should not be possible from the same window as discarding this hunk.

Say you have a document with three separate hunks (I think it doesn't matter if they're additions, deletions or modifications. This means that the document differs in three places from HEAD. If I now view the diff for one hunk and discard it this place will be reverted to HEAD, and there will be two changed hunks left. If on the other hand I decide to stage this one hunk there will be on staged hunk and two hunks that are different from HEAD but not yet staged. I can then decide to  stage or discard the hunks one by one or also commit at any time with some staged and some unstaged changes.

Or am I missing something?

> Actually, I open the popup window by double left clicking the gutter. I was focusing on displaying the diff contents. I implemented clicking operation to help the test.

OK, I see. In the end it should be accessible either by single-left-click or maybe even by hovering, I think.

> I  think I can finish implementing the "discard" window and "partial  stage" window by the end of the GSoC.  But I think at that time I can  only offer the diff of "working vs HEAD". And there are still many works  of revising the current code need to be finished before the GSoC  ending.

I think if we can see the changes of the current document vs. HEAD and decide to either discard or stage them this would be a satisfying outcome for the GSoC project and a good foundation to extend the functionality afterwards.

Probably there should one additional pair of actions be implemented and accessible for example through a context menu in the VCSDiffArea: `discard (all)` and `stage (all)`. This should be quite easy but well worth doing.

And there is one final aspect I'm not sure I have read you have a good idea about already:  Suppose I stage a hunk, then these lines are "staged". What happens if the user again changes these lines? They are in "staged modified" stage. What will happen and how will this be visualized?

wenxin-bupt:
> Say you have a document with three separate hunks (I think it doesn't matter if they're additions, deletions or modifications. This means that the document differs in three places from HEAD. If I now view the diff for one hunk and discard it this place will be reverted to HEAD, and there will be two changed hunks left. If on the other hand I decide to stage this one hunk there will be on staged hunk and two hunks that are different from HEAD but not yet staged. I can then decide to stage or discard the hunks one by one or also commit at any time with some staged and some unstaged changes.

We have the same understanding of “partial stage”. If we have three changed hunks in a file and we stage two hunks this time, we are "partial staging" the file. 

At first, I thought we could put the "stage" and "discard" functions in same popup window. But I found it would lead to sophisticated operations and introduce much more work in implementation.

Q1: When we "stage" a hunk, are we staging it directly into Index?  

That is to say, when we stage several hunks, we will add it one by one into Index. If we've added the first hunk and then made some modifications. And these modifications merge the first and the second hunk. 
How should we deal with the "stage" operation of this new hunk? I think this fashion would introduce many problems in coding. 

Q2: What if we change the "stage" operation in popup window into the "select" operation? We select the hunks that we want to stage. And then stage them by clicking "stage" in context menu.

We will face the same problem: what if user changed the hunks we've already selected? 


To simplify this, we should stop the user from making any modifications during the partial stage. I think SmartGit's Index editor is a good model. So like the Index editor, we should provide a specialized partial stage window.

wenxin-bupt:
Can we assume that Frescobaldi is always opened by an account who has the permission of reading and writing the files in repo?

If so, I don't need to take effort handling the file-write permission problems.

uliska:
If Frescobaldi can't open a file your code won't be reached in the first place.
And if it can't write there won't be any changes.

However, to be sure you might want to check, and if you can't write you should consider the document as not-versioned.

Does that help?

wenxin-bupt:
I see, thank you.

wenxin-bupt:
Hi @uliska ,

I'm considering disable all `vcs` functionalities when we meet any error or crash during running. But I find the `vcs` has greatly coupled with `view` , `sidebar`, `statusbar`. It's not difficult to disconnect the signals. But removing the widgets `diffarea`, `statusbar`,  options in context menu seems troublesome for me. Could you please give me some advice?

wenxin-bupt:
Hi @uliska ,
When Frescobaldi catch an error in `vcs` (either exitcode == 1 or git crashed), I'm thinking that we don't  raise an exception but log the error message and then disable the `gitdoc` or `gitrepo` instance or the entire `vcs` module.  In this way, the error will be handled silently. While I haven't found the same mechanism in `Frescobaldi`.  There are no `logging` or `print` related to error handling. So, am I thinking in a wrong way on this topic?

uliska:
You should disable the Git functionality for the current document or could disable the whole repository. But you should not disable the whole module as that affects Frescobaldi itself.

I suggest you simple print() a message for now.  There's no logging yet, and it would be out of scope to find a solution right now.
-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

wenxin-bupt:
This mechanism still has a problem. It can't handle the errors occur during the instance's initialization. More works need to be done.

wenxin-bupt:
In above two commits. I fix the error caused by ignoring the last-line-ending problem. 

I was stuck on this for two days. There are many special cases need to be handled respectively. I write some comments to make function `revert()` more clear. But I think it's still hard to read.

uliska:
See #1005

PR #987, opened 2017-07-20
 Some minor questions

uliska:
This is not intended to be merged but only to give convenient access to a few question comments I added to the code.

### DIFF
--- a/frescobaldi_app/vcs/gitdoc.py
+++ b/frescobaldi_app/vcs/gitdoc.py
@@ -298,6 +298,7 @@ class Document(abstractdoc.Document):
 
         # arguments to get file status
         args = [
+            # TODO: really? including ignored files?
             'status', '--porcelain', '--ignored', self._relative_path
         ]
 
--- a/frescobaldi_app/vcs/gitrepo.py
+++ b/frescobaldi_app/vcs/gitrepo.py
@@ -40,6 +40,7 @@ class Repo(abstractrepo.Repo):
     or a document's project.
     """
 
+    # Don't think we need this. gitrepo.Repo should only be instantiated if git is available.
     _git_available = None
 
     repoChanged = pyqtSignal()
@@ -218,6 +219,8 @@ class Repo(abstractrepo.Repo):
         """
         Returns True if the respective VCS is installed
         """
+        #TODO: This has to be implemented, of course
+        # because it will also tested upon application startup
         return True
 
     def track_document(self, relative_path, view):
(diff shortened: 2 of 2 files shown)


frescobaldi_app/vcs/gitdoc.py
```diff
@@ -298,6 +298,7 @@ def set_status(gitprocess, exitcode):
 
         # arguments to get file status
         args = [
+            # TODO: really? including ignored files?
```
wenxin-bupt: At that time I was think that ignored files are a part of the repository and users may open them in Frescobaldi.    

This will also help us deal with the ignored files. Since we just judge a file through its location. If we check this file's status at first, then we don't need to handle the possible errors brought by tracking a git ignored file's diff result.

uliska: > At that time I was think that ignored files are a part of the repository and users may open them in Frescobaldi.

Well, that's true (although unlikely). But if a file is ignored we don't need to bother with its diffs, as they won't go into a commit anyway.

> This will also help us deal with the ignored files. 
Do you mean you collect this list in order to *not* check files that are in it? That makes sense, of course

frescobaldi_app/vcs/gitrepo.py
```diff
@@ -238,6 +241,8 @@ def branches(self, local=True):
         If local is False also return 'remote' branches.
         """
         res = []
+        #TODO: Are you sure this works asynchronously?
+        # What happpens if this is called while the job is working?
```
wenxin-bupt: Right now, `self._update_all_attributes(blocking=True)` is called  synchronously  in `__init__` function.  So `branches()` should be available once the object is constructed. 

`self._local_branches` and `self._remote_branches` should be updated asynchronously when the `repoChanged` emits (Have not implemented).

frescobaldi_app/vcs/gitdoc.py
wenxin-bupt: > Do you mean you collect this list in order to not check files that are in it? That makes sense, of course

`_update_status()` only checks the `Document()` object's corresponding file. This object will be added into the `Repo()`. But we don't track its git diff.  We can display a status line in the bottom E.g. `In Frescobaldi on rewrite-vcs, File is ignored.` That's my current thought.

uliska: OK, good.

wenxin-bupt:
>So if you give me some hints on how to set up a way to test the code in a running Frescobaldi I'll have another look including the core diff functions.

I'm afraid you have to wait for some time. I haven't tested `gitdoc.py` and I need to take some time revise it until it can work properly.

uliska:
This isn't needed anymore

PR #1049, opened 2018-01-02
Add -p/--python-ly command line option

uliska:
With the -p or --python-ly option the path to python-ly can be specified.
This maybe makes it cleaner to use python-ly from the Git repository
(rather than wrapping the call to Python with a PYTHONPATH= clause).

This is inspired by the recent discussion about installation instructions.

### DIFF
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -192,6 +192,10 @@ def main():
         sys.exit(0)
 
     if args.python_ly:
+        # The python-ly path has to be inserted at the *second* position
+        # because the first element in sys.path is the directory of the invoked
+        # script (an information we need in determining if Frescobaldi is run
+        # from its Git repository)
         sys.path.insert(1, args.python_ly)
 
     check_ly()
@@ -280,5 +284,3 @@ def main():
         cursor.setPosition(pos)
         win.currentView().setTextCursor(cursor)
         win.currentView().centerCursor()
-
-
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -85,7 +85,6 @@ def parse_commandline():
         parser.add_argument('-p', '--port')
         parser.add_argument('-f', '--file')
         parser.add_argument('-o', '--output')
-        parser.add_argument('--python-ly')
 
 
 
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -72,7 +72,7 @@ def parse_commandline():
         help=_("List the session names and exit"))
     parser.add_argument('-n', '--new', action="store_true", default=False,
         help=_("Always start a new instance"))
-    parser.add_argument('-p', '--python-ly', type=str, metavar=_("STR"), default="",
+    parser.add_argument('--python-ly', type=str, metavar=_("STR"), default="",
         help=_("Path to python-ly"))
     parser.add_argument('files', metavar=_("file"), nargs='*',
         help=_("File to be opened"))
@@ -85,7 +85,7 @@ def parse_commandline():
         parser.add_argument('-p', '--port')
         parser.add_argument('-f', '--file')
         parser.add_argument('-o', '--output')
-        parser.add_argument('-y', '--python-ly')
+        parser.add_argument('--python-ly')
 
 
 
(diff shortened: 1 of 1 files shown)


uliska:
To clarify what I mean:  I could change my `frescobaldi` wrapper script from

```
#!/bin/bash                                                                     
PYTHONPATH=$PYTHONPATH:~/git/lilypond/python-ly python3 ~/git/lilypond/frescoba\
ldi/frescobaldi $@
```
to

```
#!/bin/bash                                                                     
python3 ~/git/lilypond/frescobaldi/frescobaldi -p ~/git/lilypond/python-ly $@ 
```

through this patch

frescobaldi_app/main.py
```diff
@@ -83,6 +85,7 @@ def parse_commandline():
         parser.add_argument('-p', '--port')
         parser.add_argument('-f', '--file')
         parser.add_argument('-o', '--output')
+        parser.add_argument('-y', '--python-ly')
```
PeterBjuhr: Will there not be a conflict for the `-p` argument when pydevd is used, and do you need to add the python ly argument again?

frescobaldi_app/main.py
```diff
@@ -189,6 +192,9 @@ def main():
         sys.stdout.write(debuginfo.version_info_string() + '\n')
         sys.exit(0)
 
+    if args.python_ly:
+        sys.path = [args.python_ly] + sys.path
```
PeterBjuhr: Another option is to write `sys.path.insert(0, args.python_ly)`

frescobaldi_app/main.py
uliska: I'm sorry but I don't really understand:
* I have named the short argument `-y` in order to avoid that conflict
* Why "again"? I don't see where it has already been used

frescobaldi_app/main.py
uliska: I'll have to investigate this anyway. I noticed that when invoking Frescobaldi with that `-y` option it obviously found the `ly` module - but the `Git` menu was missing! And I have no idea why yet ...

PeterBjuhr: After some investigation it does seem to be a direct consequence of this code!

❝the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter.❞

This is also what it is used to check if git is present which in turn is a prerequisite for the git menu:
https://github.com/wbsoft/frescobaldi/blob/master/frescobaldi_app/vcs/__init__.py#L73

❝Notice that the script directory is inserted before the entries inserted as a result of PYTHONPATH.❞
https://docs.python.org/3/library/sys.html

So it seems you should be doing:
`sys.path.insert(1, args.python_ly)`
Please confirm this!

frescobaldi_app/main.py
uliska: OK, only now I see what you meant. I have clarified in cc40d5b5

frescobaldi_app/main.py
uliska: Yes, that's correct. Thank you for investigating.
95b19711

frescobaldi_app/main.py
PeterBjuhr: Maybe it's me that don't understand! :slightly_smiling_face: As I read it you still add it twice! 

From the code and from your comments I assume you see it as two different sets of arguments and that you need to add it to both. Why is that? :thinking: 

I have searched for a relevant reference about this in the pydevd docs but haven't found any!

uliska:
I think the latest commit should remove @PeterBjuhr 's doubts. Can it be merged that way?

frescobaldi_app/main.py
```diff
@@ -189,6 +191,9 @@ def main():
         sys.stdout.write(debuginfo.version_info_string() + '\n')
         sys.exit(0)
 
+    if args.python_ly:
+        sys.path.insert(1, args.python_ly)
```
dliessi: @uliska Can you add a comment here? I would copy the corresponding commit message:
"The path to python-ly has to be inserted at the *second* position
because the first element in sys.path is the directory of the invoked
script (an information we need in determining if Frescobaldi is run from
its Git repository)"

fedelibre:
@uliska remember to update the wiki pages

I've just tested the new option and everything seems to work fine.

fedelibre:
@uliska Friendly reminder about documenting `--python-ly` option in the wiki.

PR #1057, opened 2018-02-09
Add command rename

### LINKED ISSUE #980: Wish: Either command "rename file" or option to delete upon "save as"
When I use the command "Save as..." I think in 50% of the cases the intention is to *rename* the file, while in the other half I want to make a *copy* from it.

It would be nice to have the option of renaming a file directly.
This could either be implemented as an additional command "rename" (with a dialog where I can either simply enter a new name to rename in the same directory or use Qt's "save as" dialog) or by providing an option in the existing "save as" dialog, e.g. a check box "rename (delete current file)".

Maybe these two functions could be merged with "save copy or selection as..." in *one* dialog?

@wbsoft Opinions?

wyj1026:
enable user to rename a file directly
hope to close #980


uliska:
Thank you for the contribution!

I don't see anything in the code that looks problematic, and testing didn't reveal any issues either. I just have a few comments/suggestions regarding the behaviour. These are meant for discussion, so you can voice your opinion on each of them, "defend" your decisions, make other suggestions, or simply add more commits if you agree with my ideas.

* When the current document doesn't have a name yet `renameDocument()` should not pass on to `saveDocument()`, I think. Instead the action should be disabled in the first place. So when the document is *created* the action should be deactivated, and after it is saved it should be activated. Note that a check should be triggererd when the document changes, i.e. another tab becomes active.
* I find it a bit surprising that the action doesn't open/use the default "Save as" dialog but only offers the option to rename the file in the same directory. Admittedly this will be the most common use case, but it should not be too much work to simply use the same dialog as "Save As".
* I would prefer the menu item to appear after "Save Copy or Selection As".
* The menu item doesn't have an icon. This is because there *is* no icon `document-rename`. We can only resort to either adding a custom icon (I don't think that's a good idea) or use `document-save-as` too here.

uliska:
I just realized that the "real" File Save dialog would seem somewhat odd with a function "rename", but still I think the use case would often involve saving to another directory (like "Save As" allows).

Probably the action should then be name "Rename/Move File" (just like on UNIX `mv` is used for renaming a file).

wyj1026:
Thanks for your suggestions! 


This is a  good idea, I used the default "Save as" dialog at first and then I realized it is a little like you said odd. So I changed it to a input dialog. It will be much bettter if we call it "Rename/Move File" and use the "Save as" dialog. 
> Probably the action should then be name "Rename/Move File" (just like on UNIX mv is used for renaming a file).

I'll make some commits later.

By the way, do you think the function name should be changed from `rename` to `move` like 

`renameDocument()` -> `moveDocument()` since `mv` can be used both for renaming a file and moving a file.

uliska:
Yes, that would be more consistent.
-- 

uliska:
Thank you, this is more or less fine by now, and I'd be ready to merge it!

However, I now found one additional inconsistency which would be nice to fix: after renaming/moving a file the `Open Recent` menu will only be updated after restarting Frescobaldi, so initially it will contain both the old and the new filename, where clicking on the old name will cause an error because that file is obviously not present anymore.

If you'd like to find the solution for that too it would be great, but I think this is an issue that could also be neglected.

wyj1026:
Thank you for pointing it out and thanks for your patience, I've made another commit to fix it. :smile:

uliska:
Thank you for the fix, I'll now merge it.

Don't worry about "patience". Sometimes it is simpler to just complete things oneself, but as I somehow suspected this was at least related to GSoC I thought to let you do the stuff and see how you respond to review. Apart from that now you get the credit as documented in the Git history.

PR #1063, opened 2018-03-15
Submenus for grouping sessions in Sessions menu

uliska:
This commit parses session names for (one) slash in it and treats such
sessions as grouped sessions and creates submenus for each.

image

The image also shows that the current session is within the `Group` submenu

### DIFF
--- a/frescobaldi_app/sessions/menu.py
+++ b/frescobaldi_app/sessions/menu.py
@@ -61,8 +61,6 @@ class SessionMenu(QMenu):
                 add_session(self, name)
 
         def add_groups():
-            g_keys = [k for k in groups.keys()]
-            g_keys.sort()
             for k in sorted(groups.keys()):
                 m = self.addMenu(k)
                 for name in groups[k]:
--- a/frescobaldi_app/sessions/menu.py
+++ b/frescobaldi_app/sessions/menu.py
@@ -35,43 +35,71 @@ class SessionMenu(QMenu):
     def __init__(self, mainwindow):
         super(SessionMenu, self).__init__(mainwindow)
         app.translateUI(self)
-        mgr = manager.get(mainwindow)
-        ac = mgr.actionCollection
-        ag = self._actionGroup = QActionGroup(self)
-        ag.setExclusive(True)
-        ag.addAction(ac.session_none)
-        ag.triggered.connect(self.slotSessionsAction)
-        self.addAction(ac.session_new)
-        self.addAction(ac.session_save)
-        self.addSeparator()
-        self.addAction(ac.session_manage)
-        self.addSeparator()
-        self.addAction(ac.session_none)
-        self.addSeparator()
         self.aboutToShow.connect(self.populate)
 
     def translateUI(self):
         self.setTitle(_('menu title', '&Session'))
 
     def populate(self):
-        ac = manager.get(self.parentWidget()).actionCollection
-        ag = self._actionGroup
-        for a in ag.actions():
-            if a is not ac.session_none:
-                self.removeAction(a)
-                ag.removeAction(a)
-        ac.session_none.setChecked(not sessions.currentSession())
-        for name in sessions.sessionNames():
-            a = self.addAction(name.replace('&', '&&'))
+        groups = {}
+        tl_sessions = []
+
+        def add_session(menu, name, group=None):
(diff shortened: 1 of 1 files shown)


PeterBjuhr:
To be able to group the session is a nice idea! I'll take a look as soon as I can find some time!

frescobaldi_app/sessions/menu.py
```diff
@@ -35,43 +35,71 @@ class SessionMenu(QMenu):
...
+            a.setObjectName(fullname)
+            self._actionGroup.addAction(a)
+
+        def add_sessions():
+            for name in tl_sessions:
+                add_session(self, name)
+
+        def add_groups():
+            g_keys = [k for k in groups.keys()]
+            g_keys.sort()
```
PeterBjuhr: Will you explain these two lines? It seems to me to belong to a first sketch of the code and should be removed!?

frescobaldi_app/sessions/menu.py
```diff
@@ -35,43 +35,71 @@ class SessionMenu(QMenu):
...
+
+        def reset_menu():
+            self.clear()
+            ac = manager.get(self.parentWidget()).actionCollection
+            ac.session_none.setChecked(not sessions.currentSession())
+            ag = self._actionGroup = QActionGroup(self)
+            ag.setExclusive(True)
+            ag.addAction(ac.session_none)
+            ag.triggered.connect(self.slotSessionsAction)
+            self.addAction(ac.session_new)
```
PeterBjuhr: Do we need to add these each time now? Previously they were just added on init.

frescobaldi_app/sessions/menu.py
```diff
@@ -35,43 +35,71 @@ class SessionMenu(QMenu):
...
+            self.addAction(ac.session_manage)
+            self.addSeparator()
+            self.addAction(ac.session_none)
+            self.addSeparator()
+
+        reset_menu()
+        for name in sessions.sessionNames():
+            if '/' in name:
+                group, name = name.split('/')
+                g = groups.get(group, [])
```
PeterBjuhr: Why not just `if group in groups:`?

frescobaldi_app/sessions/menu.py
uliska: https://github.com/wbsoft/frescobaldi/pull/1063/commits/2274adc5a662812deb556739102949a82f9eda22

Yes, obviously when testing the code for the *subsequent* line I got distracted :-(

frescobaldi_app/sessions/menu.py
uliska: I may be wrong (would be happy) but after quite some searching I didn't find a way to *remove* a submenu from a `QMenu`. But it is important to remove the submenus each time the menu is populated, so I found it's the easiest way to just clear the whole menu. After all these actions aren't *created* but only *added*. 
But I'd be happy to learn about a cleaner way to achieve it.

frescobaldi_app/sessions/menu.py
uliska: OK, yes, that's cleaner.
https://github.com/wbsoft/frescobaldi/pull/1063/commits/2077b4fc99169a710744bb56f909e395d644acd0

frescobaldi_app/sessions/menu.py
PeterBjuhr: I can dig deeper into this if you like. But as I understand it you remove (also) submenus with `removeAction`. It's in the original code.

uliska:
ok, I'll try that.
-- 

frescobaldi_app/sessions/menu.py
uliska: Unfortunately it doesn't work like that:

image

I'll ask on stackoverflow if it's possible to remove submenus.

uliska: Sorry, should have looked in *that* direction before sending the previous comment:

It's possible to remove the submenu with `<submenu>.deleteLater()`. I'll update the code and push something soon.

PeterBjuhr: Ok, I guess that resolves the issues I had initially!

uliska:
Thanks.
I don't expect Wilbert to weigh in on that, but who knows ...

I'll add some documentation before merging.
-- 

PR #1074, opened 2018-07-01
Add file path as tooltip in Documents menu

uliska:
Adds the tooltip to the Documents menu that is also found in the Documents Tool and the editor's tabs.

Merges duplicate implementation to a common function in the `util` module.


uliska:
I think this is uncontroversial, so I'll wait one more day and then merge this myself.

PR #1078, opened 2018-07-19
Add Tools=>Directions submenu with three commands

### LINKED ISSUE #1044: Quick remove, add action "directions"
Add an entry to the "quick remove" menu that changes all directional operators (`^` and `_`) to the neutral `-`.

Maybe the function could also replace `\nnUp` and `\nnDown` with `\nnNeutral`.

This is useful for cleaning up input that has been produced by OMR.

uliska:
Fixes #1044 

This is a tool similar to the Quick Remove tools used to force directions to be up, down or neutral.

I need help with correctly replacing words in the document,
and there would be the need for an icon


uliska:
My problem where I need help is in this segment: https://github.com/wbsoft/frescobaldi/pull/1078/files#diff-5f0b2d6d5566d16905f105868e6e1361R208.

It seems the latest commit addresses that, but I'm wary if I haven't overlooked something about how text insertions are handled (I think my code is standard PyQt, but there are specific things for maintaining point-and-click information, IIRC).

uliska:
I think the latest commit makes it more in line with the coding style. Probably it will be slightly difficult to merge into the tools menu ...

Issue #1076, opened 2018-07-18
Questions about the Show Available Fonts dialog


uliska:
This is related to #1075.

I just moved the "show available fonts" dialog to its own module because I think it's too heavy to be squeezed into the `engrave.lytools` module.

As a consequence I would also move its actions out of the `engrave` module completely, because apart from the fact that it starts LilyPond it is completely unrelated to "engraving".

In fact I would even suggest to move the menu entry from `LilyPond` to `Tools`, maybe at the end of the group with Pitch, Rhythm etc. The idea behind this is that I thought that once the display situation is improved this dialog could receive some additional functionality, for example generating font selection code, either "locally" or as a general document-wide font setting.

What do you think about these ideas?

uliska:
Well, I can do that quickly and still wait for Wilbert and/or Peter to comment on. But I won't go for the code generation right now.

dliessi:
I agree with @fedelibre.

uliska:
Obsolete

PR #1075, opened 2018-07-17
Improve "Show available fonts" dialog

uliska:
Closes #420 
Closes #1077

This PR is not ready to be merged but I would like to get some feedback. I want more functionality, but actually the code would be a significant improvement already now.

**[EDIT:]** I've removed the "WIP" from the title and consider the PR ready by now.

The output of the `LilyPond=>Show Available Fonts...` command is practically unusable when more than a few fonts are available on a system. It is totally overwhelming, and the structure is hard to understand. Therefore I wanted to have an improved presentation of the data.

In its initial state the PR doesn't simply print out the LilyPond output but parses it to a number of dictionaries. The most important one is the `families` dictionary with entries for each font family. For now an ordered and structured list is printed to the same log window as was previously in that dialog. The "filter" line edit is still non-functional.

image

## To be done:

* Replace the log widget with a custom widget. What I think would be appropriate is a list-style widget where initially only the font families are listed. Only upon "opening" an entry will additional information be loaded and displayed (and these can then be pretty detailed).
* Implement the filter edit. When some text is typed into that line edit all fonts not matching the pattern should be suppressed by making their parent widget invisible.
* Add an export/print/save function 

The code *could* be merged without these additional features, after removing the non-functional filter edit.

PR #1080, opened 2018-07-20
Restructure Tools Menu

uliska:
The Tools menu is becoming too long IMO, so I think we should introduce additional submenus, even if that requires one more level to access some options. This PR introduces the following changes:

* New "Code Formatting" submenu with reformat, indent etc.  
  I have some ideas to add more "linting" commands that would further extend that list of commands, which is an additional incentive to move them in a submenu
* Submenus for Tools Panels

To be discussed:

* Naming of the panels submenus
* organization of panels in these submenus
* should the remaining tools also be moved to submenus, if so, what should they be called like
* If that is accepted the mnemonics for the panels can be reconsidered because they have to be unique only in their submenus
* Should the section from "Pitch" to "Quick Remove" also be moved to a submenu? That would be consistent but may move some regularly used commands one level deeper
* Should "Update with convert-ly" be moved to either the new "Code Formatting" submenu or to the LilyPond menu?

### DIFF
--- a/frescobaldi_app/panelmanager.py
+++ b/frescobaldi_app/panelmanager.py
@@ -60,6 +60,10 @@ class PanelManager(plugin.MainWindowPlugin):
                 'menu': QMenu(_('&Structure')),
                 'panels': []
             }),
+            ('midi', {
+                'menu': QMenu(_('&MIDI')),
+                'panels': []
+            }),
             ])
 
         # add the the panel stubs here
@@ -68,14 +72,14 @@ class PanelManager(plugin.MainWindowPlugin):
         self.loadPanel("viewers.manuscript.ManuscriptViewPanel", "viewers")
         self.loadPanel("docbrowser.HelpBrowser", "viewers")
         self.loadPanel("logtool.LogTool", "viewers")
+        self.loadPanel("layoutcontrol.LayoutControlOptions", "viewers")
         self.loadPanel("quickinsert.QuickInsertPanel", "coding")
         self.loadPanel("charmap.CharMap", "coding")
         self.loadPanel("snippet.tool.SnippetTool", "coding")
         self.loadPanel("doclist.DocumentList", "structure")
         self.loadPanel("outline.OutlinePanel", "structure")
-        self.loadPanel("miditool.MidiTool")
-        self.loadPanel("midiinput.tool.MidiInputTool")
-        self.loadPanel("layoutcontrol.LayoutControlOptions")
+        self.loadPanel("miditool.MidiTool", "midi")
+        self.loadPanel("midiinput.tool.MidiInputTool", "midi")
 
         # The Object editor is highly experimental and should be
         # commented out for stable releases.
--- a/frescobaldi_app/panelmanager.py
+++ b/frescobaldi_app/panelmanager.py
@@ -91,7 +91,6 @@ class PanelManager(plugin.MainWindowPlugin):
         mainwindow.tabifyDockWidget(self.musicview, self.docbrowser)
         mainwindow.tabifyDockWidget(self.musicview, self.svgview)
 
-        print(self._submenus)
 
     def loadPanel(self, name, submenu=None):
         """Loads the named Panel.
--- a/frescobaldi_app/panelmanager.py
+++ b/frescobaldi_app/panelmanager.py
@@ -23,8 +23,10 @@ Manages the Panels (Tools).
 
 
 import sys
+import collections
 
 from PyQt5.QtCore import QSettings
+from PyQt5.QtWidgets import QMenu
 
 import actioncollection
 import actioncollectionmanager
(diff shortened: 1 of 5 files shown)


uliska:
image

uliska:
I have now also moved the musical transformations to a new submenu. I had hesitated to do so because it will add one layer to access the functions, *but* I had the idea to make that whole new submenu also available in the editor's and all viewers' context menus, so when you right-click on an item in the score the context menu will have an item `Musical Transformations=>` available (note: this is not included in this PR).

image

fedelibre:
I thought just few days ago that the Tools menu was too long. I like the new restructure.

Just a few comments for now:

- what about adding "Layout Control Options" to the Viewers submenu? It's basically an advanced music  viewer.
-  then I would group the two MIDI tools in a MIDI submenu

fedelibre:
> Should "Update with convert-ly" be moved to either the new "Code Formatting" submenu or to the LilyPond menu?

I think it should be moved to the LilyPond menu, as it's about LilyPond syntax version.

fedelibre:
This comment is not much related to this PR.
IMO the Documents menu is not that useful and it's stealing space which might be assigned to something else.

I usually use Alt+arrows to move from one document to another.
If I have lots of files open, then I'd rather open the Documents pane, which opens on the left by default. This is what you expect from most editors; you don't have this list in a dedicated menu.

uliska:
Am 22. Juli 2018 12:59:57 MESZ schrieb Federico Bruni <notifications@github.com>:
>This comment is not much related to this PR.
>IMO the Documents menu is not that useful and it's stealing space which
>might be assigned to something else.
>
>I usually use Alt+arrows to move from one document to another.
>If I have lots of files open, then I'd rather open the Documents pane,
>which opens on the left by default. This is what you expect from most
>editors; you don't have this list in a dedicated menu.

I know that some people including Wilbert feel pretty strong about that menu,and while the documents *panel* may be the more traditional place I think the problem is that it *permanently* uses space. Alt+D is a very fast way of accessing the list of documents.

-- 

fedelibre:
> I've done so, but I'm not 100% convinced. convert-ly operates on the currently opened document and not on the LilyPond installation (like e.g. the available fonts). This would make it more likely to be placed next to the new "Code formatting" and "Musical transformations" submenus in Tools?

Well, also the Engrave commands in that menu operate on the currently opened document.

Updating the syntax of a file is not exactly a "code formatting" operation.
But I think I see a point in keeping it in the Tools menu:

- LilyPond menu is mainly about producing output. You may even call it Engrave...
- Tools menu is mainly about changing the input file, so it's where you'd expect to find a tool like convert-ly

I think I changed my mind and I'd rather place it in Code formatting submenu.

uliska:
I'd rather place it below the two submenus but within the same group of separators. The commands in Code Formatting have in common that they don't change the content in any way.

uliska:
... which would be the following (uncommitted yet):

image

frescobaldi_app/panelmanager.py
```diff
@@ -45,33 +47,49 @@ def __init__(self, mainwindow):
...
-            self.loadPanel("objecteditor.ObjectEditor")
+            self.loadPanel("objecteditor.ObjectEditor", "coding")
         self.createActions()
 
         # make some default arrangements
         mainwindow.tabifyDockWidget(self.musicview, self.docbrowser)
         mainwindow.tabifyDockWidget(self.musicview, self.svgview)
 
-    def loadPanel(self, name):
+        print(self._submenus)
```
dliessi: @uliska Please, remember to delete debugging lines like this one before committing.

dliessi:
I like this change very much and I also like the idea of having the "musical transformation" submenu in the context menu.
Regarding the position of "update with convert-ly", I agree with
> I'd rather place it below the two submenus but within the same group of separators.

dliessi:
Shouldn't also "show available fonts" be moved to the "tools" menu?

frescobaldi_app/panelmanager.py
uliska: Arggh

uliska:
> Shouldn't also "show available fonts" be moved to the "tools" menu?

I will do so once #1075 is merged.

uliska:
This is the current state:

image

uliska:
Yet another modification: I've merged a few commands in a new `Tools=>Directories` submenu.

For the reasoning I copy the commit message:

> This warrants another submenu because (when we may finally merge
the Git code from GSoC) I want to add the commands to open the current
document's repository root in the terminal or file explorer too.
>
> Moving the LilyPond data directory here doesn't hurt IMO. When the
"show available fonts" command has been moved to Tools too the "LilyPond"
menu will be exclusively for engraving-related matters and could even
be renamed (if people don't object too much to that ...)

image

---

When the "show available fonts" PR has been merged I want to add a "Text/Markup" submenu below "Musical Transformations". This will initially hold only that single action, but others should be added:

* choose font (for a single in-place command)
* choose document fonts
* Maybe even (feature request) a text editing dialog that makes LilyPond's markup commands available to a GUI text editor.

uliska:
For completeness: After merging the `force-directions` branch there is an additional submenu Tools=>Musical Transformations=>Directions.

When the musical transformations are made available through context menus there should also be actions to *flip* directions.

image

dliessi:
>> Shouldn't also "show available fonts" be moved to the "tools" menu?

> I will do so once #1075 is merged.

Ok, I have just seen (and commented on) #1076.

dliessi:
Two comments on the icons.

1. I think that the Audio-x-generic icon would look better in the menu without the shadow.
2. The music-directions icon is stretched horizontally to a square: it would be better to prepare it as a square from the beginning.

(Both tasks are easily done with Inkscape.)

uliska:
> 1.   I think that the Audio-x-generic icon would look better in the menu without the shadow.

Actually I'm totally unsure how to best proceed with this issue. I did a web search for "Pango Icons Music" and ended up with a list claiming to be Pango icons (which is where Frescobaldi mainly gets its icons from).

> 2.   The music-directions icon is stretched horizontally to a square: it would be better to prepare it as a square from the beginning.

OK, I'*ll have a look.
I'll see how it works out in Inkscape. I have experienced that SVG files modified/saved in Inkscape were not visible in Frescobaldi.

dliessi:
> I have experienced that SVG files modified/saved in Inkscape were not visible in Frescobaldi.

Maybe you only need to ensure they are saved as "pure SVG" instead of "Inkscape SVG" (but I'm not sure).

Issue #1090, opened 2018-08-08
How can I programmatically set the page in MusicView or Manuscript View?


uliska:
I'm getting more and more confused, so I have to ask rather than continue digging through the sources ...

I have an action that needs to open a given document in the manuscript viewer (if it isn't already opened) and scroll to a specific page. I'm totally confused by now discerning view, widget, popplerview, surface etc.

So how do I get hold of an object where I can open a document in the viewer and call something like `setCurrentPage()` on it?

uliska:
Never mind, it's actually the panel itself.
For better access to this I've added a function to `PanelManager`, which I'll upload in a minute as PR.

PR #1091, opened 2018-08-08
PanelManager: Add panel_by_name() function

uliska:
This function is helpful to access any panels from anywhere else.

I think it is pretty lightweight - except if there's another way already present that I didn't see.

(relates to #1090)

Usage example:

```python
import panelmanager
manuscript_viewer = panelmanager.manager().panel_by_name('manuscript')
```


uliska:
I will merge this tomorrow. There's absolutely nothing specific about it, I think.

PR #1095, opened 2018-08-14
Ensure includepaths have trailing slashes

### LINKED ISSUE #1094: Include-path not stored correctly on Windows
I tried to run the Edition Engraver today with Frescobaldi. It didn't work on Windows (though I could get it running on Linux.)
I pointed Frescobaldi's includepath to the folder containing OpenLilyLib (on my machine: "D:/OpenLilyLib"). Nevertheless I got the error "cannot find '/oll-core/package.ily' ". It turns out Frescobaldi saves the includepath without a trailing slash ("D:/OpenLilyLib"). I had to manually append the slash in Windows registry so it read "D:/OpenLilyLib/". After that, Frescobaldi/Lilypond could find the correct packages.

So it seems Frescobaldi should be updated to append a slash to a Windows-folder in the includepath settings panel.

uliska:
Closes #1094

It seems LilyPond on Windows requires a trailing slash while Linux
doesn't care about this.

I don't know about Mac, though, so the PR should be tested on Mac
before that.

### DIFF
--- a/frescobaldi_app/engrave/command.py
+++ b/frescobaldi_app/engrave/command.py
@@ -39,6 +39,13 @@ def info(document):
     return lilypondinfo.preferred()
 
 
+def paths(includepath):
+    """Ensure paths have trailing slashes for Windows compatibility."""
+    result = []
+    for path in includepath:
+        result.append('-I' + path.rstrip('/') + '/')
+
+    
 def defaultJob(document, args=None):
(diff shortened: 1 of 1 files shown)


dliessi:
@uliska Please squash the commits into a single one and rebase on master before merging: there is no need to preserve the history of mistakes and I would prefer avoiding merge commits unless necessary.

PR #1098, opened 2018-08-27
update italian translation

fedelibre:
I did not include the file it.mo in this PR.


dliessi:
I would be in favor of removing the mo files from the repository, given that they are generated programmatically, provided that we either make sure that we generate and include them in source releases, or include gettext in the build requirements and adapt setup.py to automatically generate mo files before installing.
Until that time, I think that we should keep including and updating the mo files in the repository.

Anyway, I would like to hear from @wbsoft on this topic and also on the other point of [your recent message to the Frescobaldi list](https://groups.google.com/d/msg/frescobaldi/ROVJg4mRS3M/D8t_wxjdCAAJ) (i.e., using the standard Python gettetxt library).
Can you please open a new issue to discuss these? Thanks.

fedelibre:
Thanks for the review: I've applied your suggestions in a new commit (which can be squashed when merging).

I chose to use a single term to translate "directory", for consistency.

fedelibre:
@uliska Can you merge it?

uliska:
Are you in agreement about the .mo/.po files questions?

fedelibre:
I agree with what @dliessi wrote above. IIUC it implies that in order to run localized Frescobaldi from source I would have to run `python3 setup.py build`, correct? (once the setup is changed, of course)

However, right now I think you should run make and commit it.mo file as well.

dliessi:
> However, right now I think you should run make and commit it.mo file as well.

I agree.

uliska:
@fedelibre could you please check if that latest commit was right?

fedelibre:
Yes, the  latest commit is correct: I can see the latest translation in the GUI

uliska:
I was surprised to see that after running `make` a *lot* of `.mo` files had been updated. Since they are "binary" files I don't really know what is going on here, and I don't have experience with the translation process.

What's happening here? can it be that these are `.mo` files that haven't been committed after earlier translations? Should we commit these too?

fedelibre:
I've just  run make on current master (including this merged commit).
All .mo files have a new modification time (today), *but* only it.mo has changed. I guess that the change might depend on the difference of versions of gettext you and I are  running, which might explain why my msgfmt  changes it.mo.

On Fedora 28 I have:

```
$ dnf info gettext | grep Version
Versione     : 0.19.8.1

$ msgfmt --version
msgfmt (GNU gettext-tools) 0.19.8.1
```

fedelibre:
Ok, I've followed these instructions on [how to diff git mo files](https://gist.github.com/meddulla/d7596d81d4df923d3864) and I see that the only difference is the POT creation date merged in the it.mo file:

```diff
$ git diff .
diff --git a/frescobaldi_app/po/it.mo b/frescobaldi_app/po/it.mo
index cea6ec76..c72f9ff4 100644
--- a/frescobaldi_app/po/it.mo
+++ b/frescobaldi_app/po/it.mo
@@ -2,6 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Frescobaldi 3.0.0\n"
 "Report-Msgid-Bugs-To: info@frescobaldi.org\n"
+"POT-Creation-Date: 2018-08-23 15:52+0200\n"
 "PO-Revision-Date: 2018-08-28 16:51+0200\n"
 "Last-Translator: Federico Bruni <fede@inventati.org>\n"
 "Language-Team: Italiano <>\n"
```

uliska:
That means I'll simply discard these updated .mo files.

PR #1100, opened 2018-09-02
Move "job" related code to "job" module

uliska:
In the context of https://github.com/wbsoft/frescobaldi/projects/4
This is the first step of encapsulating all job related code into a module,
which will make extending the functionality substantially easier:
https://github.com/wbsoft/frescobaldi/projects/4#card-12637526

The next steps in this specific sub-topic of the project will be
a)
making clearer that a "job" can be anything, not only a LilyPond job
(admittedly I'm right now not sure if that's not the case anyway)
b)
implementing a multithreaded job queue (which I actually have done
in a private fork for a project and will make generically available)


uliska:
This PR should actually not change *anything* usage-wise. I just moved the files into a `job` subdirectory and adjusted the `import` statements. But I'd like to have this looked at by someone before actually merging myself.

uliska:
OK, now I think this PR is fairly complete, with a complete overhaul of the `Job()` infrastructure (implying lots of opportunities for bugs ...).

I've restructured everything relating to `job.Job()`, that is code that so far used `job.Job()` or `process.Process()` - which had been pretty redundant. Now everything is in a single `job` module, there's the generic `job.Job()` class and the more specific `job.lilypond.LilyPondJob()` class. The latter prepares the job for a typical LilyPond compilation (i.e. handles the differences between a generic and a LilyPond job), and I think it's an improvement that the user (using code) doesn't have to manually compose the command line for LilyPond anymore but that it is done implicitly by adding options and arguments. Most of the manual specification that had been done in each module *using* jobs is now donw by using an appropriate subclass of `LilyPondJob`.

uliska:
... which means I would like to have it reviewed and merged now, before proceeding further.

uliska:
One thing I *know* I've managed to break is the decoding of log output in non-English languages.

dliessi:
Using custom engraving I get a bunch of
```
2018-09-05 08:53:03.626 Python[3124:111661] Path  given to -[NSWorkspace iconForFile:] is not a full path.
```

If I then engrave _without_ selecting to run LilyPond with English messages, I get
```
  File "/Users/davide/src/frescobaldi/frescobaldi_app/engrave/__init__.py", line 168, in engraveCustom
    self.runJob(dlg.getJob(doc), doc)
  File "/Users/davide/src/frescobaldi/frescobaldi_app/engrave/custom.py", line 216, in getJob
    del j.environment['LANG']
KeyError: 'LANG'
```

dliessi:
Apparently the preview in the score wizard is not working.
Each time I click the preview button a new empty document is opened in Frescobaldi and the preview in the score wizard does not change.
The first problem (new empty document each time preview is clicked) is due to this pull request, the second (preview does not change) apparently is not.

dliessi:
> the second (preview does not change) apparently is not.

Indeed also v2.x has this problem.

Before I open a new issue, can you confirm that on Linux clicking the preview button actually updates the preview in the score wizard dialog?

uliska:
>> If I then engrave without selecting to run LilyPond with English messages, I get
>>
> I think I know what the (coding) error is here. I assume it doesn't happen when you have activated "run LilyPond with English messages" in the Preferences.

Should be fixed by https://github.com/wbsoft/frescobaldi/pull/1100/commits/a10736afe2b0c31f87e82505772221d796542530

uliska:
> ... and the preview in the score wizard does not change.

@dliessi please insert something like `print("Show preview")` after line 122 of `scorewiz/dialog.py` (the beginning of the `showPreview()` method) to test whether the button initially triggers the right action.

uliska:
One issue I notice now:

When the "sticky document" is not the one currently open the LilyPond log isn't connected to the job.

uliska:
It seems I have now fixed all the issues listed in the original description *except* those that were reported on Mac and Windows.

I can't do anything about them before getting feedback on the "tasks" I have given the reporters. Please test with the latest commits, it is quite possible that my fixes also affected other problems.

@dliessi

dliessi:
@uliska 
> Clicking the preview button should instead open a pop-up dialog where you'll first see a LilyPond log

This happens...

> and upon completion the PDF:

... and this does not, even in branch `v2.x`.

> Is that a completely empty document that is opened, and does it make a difference what is set in the dialog as score content?

Yes, a completely empty document, regardless of selections in the dialog.
Also I just noticed that when closing those empty documents there are several graphical glitches with the document tabs (e.g. closed tabs don't disappear completely).

>>> If I then engrave without selecting to run LilyPond with English messages, I get

>> I think I know what the (coding) error is here. I assume it doesn't happen when you have activated "run LilyPond with English messages" in the Preferences.

> Should be fixed by a10736a

Confirmed.

I also confirm that the log in non-English languages appears now to be readable.

I'm now going to test the score wizard as you suggested in https://github.com/wbsoft/frescobaldi/pull/1100#issuecomment-419003445.

dliessi:
> Also I just noticed that when closing those empty documents there are several graphical glitches with the document tabs (e.g. closed tabs don't disappear completely).

Ok, that's not correct: this problem does not depend on closing those empty documents but on closing any document by clicking on the cross.
Closing using the keyboard shortcut does not exhibit the problem.
I'll open a new issue for this.

dliessi:
> I'm now going to test the score wizard as you suggested in #1100 (comment).

The `showPreview()` method is correctly triggered.

dliessi:
(See #1101 for the score wizard preview and #1102 for the glitches in document tabs.)

dliessi:
> Just to be clear: you are sure that you configured at least one part in the Parts tab of the dialog?

Ok, I'm sorry for the confusion: I never use the score wizard so I didn't know what to expect, but I should have imagined that with no parts there would be no score block and thus no graphic output (normal LilyPond behavior and also suggested by the log).
Indeed after adding parts in the Parts tab the music preview is correctly shown.
I'll close the corresponding issue.

> I could reproduce the following:
>
> * Open the Score Wizard
> * Do _not_ configure a part
> * Click `Preview`
> * Preview dialog opens with the LilyPond log running
> * _At the same time_ an empty document `document.ly` is opened in the editor but not made active.
> * If I then click `Cancel` the `document.ly` stays open
> * If I click OK instead a new document "Untitled" with just a version string is created and opened.

That's what I see, but in the last point if I click OK the empty `document.ly` stays open, too.

> Is the issue with closing documents related to this PR or not?

I think it is not: as I wrote in #1102 it happens with _all_ documents (not only the spurious `document.ly`) also with 77e2c0d, which dates back to December 2017, so it is not related to recent changes or this PR.

uliska:
Ok, thanks for the clarification.

I have by now identified the *trigger* for the problem, but will have to take a closer look to find a solution.
But I'm glad I can check (or rather ignore) the remaining problems. I'd assume the problem with closing tabs is somehow related to the other issues between MacOS and PyQt5.

uliska:
As I haven't heard anything for five days, I'll probably merge the PR tomorrow, leaving out the job queue code for a new branch from master.

dliessi:
@uliska I'm sorry, I haven't yet tested the latest changes. I'll try to test them in the evening.

uliska:
@dliessi OK, I'll wait for your comments.

dliessi:
I tested the current state of the branch.
Everything seems to work correctly.
No more spurious empty documents!

I still get the `...NSWorkspace iconForFile:...` warnings mentioned in https://github.com/wbsoft/frescobaldi/pull/1100#issuecomment-418619607: I believe that they are Mac-specific and that they are related to the icons for file output formats.
I also still have the issue with document labels, but we already concluded that it is not related to this PR.

I think this can be merged (didn't look at the code, though).

dliessi:
@uliska Please, consider cleaning the commit history a little bit before merging (e.g. squashing commits that only fix mistakes introduced in other commits), possibly doing a fast-forward merge (i.e. rebasing on top of master).
Of course this is just a personal preference, not a requirement (and of course I'm in no position to impose requirements 😃).

uliska:
I will have a look how the history can be tidied up a bit. Indeed there's no need to document the meandering ways this has taken ;-)

But I will do a merge commit, as has been the common understanding in this project. I can see your point but I think it's important to have on single commit moving master forward - also got the case if anything turns out to be problematic later.

PR #1103, opened 2018-09-19
A global job queue

uliska:
A job queue is a queue from which a given number of "runners" can pick jobs to process. This allows a specified number of processes to be run in parallel, allowing Frescobaldi to make use of multiple CPU cores, even when LilyPond itself can't run in parallel. There are multiple use cases that will be possible with this job queue:

* launching a background process so crawling for autocomplete/highlighting data won't freeze Frescobaldi with large documents
* Starting to compile a document without having to abort the compilation of another document
* batch compiling large numbers of files (successfully tested in a private fork with an earlier version of the code. This *significantly* reduces waiting times.
* etc.

This PR provides mainly a generic implementation of a (multithreaded) job queue and integrates all "external" processes (i.e. jobs except LilyPond engraving) with the global `app.job_queue()` object. This isn't terribly useful yet but will serve as the basis for substantial improvements, and I want to have them reviewed and merged before going forward.

This is related to https://github.com/wbsoft/frescobaldi/projects/4 (not fully populated yet).


fedelibre:
I'm not sure what I should test as user. Or is it just a code review?

uliska:
As a user you can simply test if you notice any problems when using Frescobaldi. Most  changes could affect the "external" commands like convert-ly etc. (note that show available fonts is broken but I didn't want to touch it because of my open work on that).

PR #1109, opened 2018-09-24
#1105: Fix external file import

### LINKED ISSUE #1105: MusicXML import - OSError: Process didn't complete in time
I'm trying to import a MusicXML file (extension changed from .mxl to .txt to be able to attach it here):
[debussy-deux-arabesques.txt](https://github.com/wbsoft/frescobaldi/files/2397742/debussy-deux-arabesques.txt)

The .ly file is generated but it's not displayed on Fresco editor pane, while the following error appears:

```
  File "/home/fede/src/frescobaldi/frescobaldi_app/file_import/__init__.py", line 99, in importMusicXML
    self.openMusicxmlDialog()
  File "/home/fede/src/frescobaldi/frescobaldi_app/file_import/__init__.py", line 109, in openMusicxmlDialog
    self.runImport()
  File "/home/fede/src/frescobaldi/frescobaldi_app/file_import/__init__.py", line 158, in runImport
    j = dlg.run_command()
  File "/home/fede/src/frescobaldi/frescobaldi_app/file_import/toly_dialog.py", line 155, in run_command
    raise OSError("Process didn't complete in time")
OSError: Process didn't complete in time
```

uliska:
This closes #1105, but not only by removing the limitation to the import timeout but by a complete review of the way file imports are structured.

In terms of implementation much redundancy has been removed and the code streamlined.

For the user interface external imports are now displayed with the help of a (new) dialog running external jobs and displaying the logs.

Additionally the import now fails properly when the underlying import script fails (previously a truncated `.ly` file had been produced without giving any warning).

### DIFF
--- a/frescobaldi_app/engrave/custom.py
+++ b/frescobaldi_app/engrave/custom.py
@@ -206,7 +206,7 @@ class Dialog(QDialog):
                 k, v = job.lilypond.parse_d_option(t)
                 j.set_d_option(k, v)
             else:
-                j.add_additional_arg(t)
+                j.add_argument(t)
 
         # Set environment variables for the job
         if self.englishCheck.isChecked():
--- a/frescobaldi_app/file_import/__init__.py
+++ b/frescobaldi_app/file_import/__init__.py
@@ -23,9 +23,11 @@ Import non-lilypond file types.
 
 
 import os
+import importlib
+import shutil
 
-from PyQt5.QtCore import Qt, QUrl
-from PyQt5.QtGui import QKeySequence, QTextCursor
+from PyQt5.QtCore import QUrl
+from PyQt5.QtGui import QTextCursor
 from PyQt5.QtWidgets import QAction, QFileDialog, QMessageBox
 
 import app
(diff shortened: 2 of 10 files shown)


uliska:
Nice!

However, there are two interface/behaviour things that may be discussed (although I'd prefer merging immediately ;-) )

1)
Now we have two dialogs after another instead of only one: First the configuration, then the log. In order not to waste users' time I decided to close the log dialog immediately upon successful completion.  But while most of the time one will *not* be interested in staring at the successful log of an import there *may* be reasons one would want so anyway. This could be made configurable, but I'm not exactly looking forward doing so ...

2)
I find the behaviour for handling multiple imports less than ideal. So far (I've just taken the existing implementation) `File=>Import=>Import...` produces an arbitrarily sorted list of input files that is handled one by one, with both dialogs opening for every single input file. I think this could be made more elegant by grouping the files by file type, then provide the configuration dialog only once per file type. Together with the automatic closing of the log dialog this will make the process of importing multiple files more straightforward.

3)
When the job queue will be available we could have the imports run in parallel. This raises the question of how the GUI (especially the logs) can be handled. My first thought is to give the (general) job dialog the ability to display multiple jobs in tabs. Successful jobs' tabs can be closed automatically while failed jobs remain.

---

What do you think?

fedelibre:
A very minor "issue" I noticed is that when you import a big MusicXML file (I tested a 1,7M xml file), there's a short moment (about 2 seconds), between the moment the log window is closed and the moment the lilypond code appears in the editor pane, when you  don't see anything happening on screen and you might wonder if the job crashed. It's short in my test, but it may grow depeding on how big is the xml file and how fast a computer is... So this might be an argument in favour of keeping the log window open. Or, probably better, you may add a sort of "spinning weel" (as in web) to let the user know that the job is still going on.

In general, I guess that if the import works correctly, the log window should be closed.

I like the idea of a real batch import, leveraging the job queue to run several processes in parallel.

uliska:
> A very minor "issue" I noticed is that when you import a big MusicXML file (I tested a 1,7M xml file), there's a short moment (about 2 seconds), between the moment the log window is closed and the moment the lilypond code appears in the editor pane, when you don't see anything happening on screen and you might wonder if the job crashed. I

What happens when you independently open the generated .ly file from the File Open command? I'd assume the delay would be exactly the same. In that case it might be a good idea to modify the general document-opening code to display a "waiting" mouse pointer. I'd start with explicitly changing the pointer, and when that looks flickering in general cases I'd insert a delaying timer, so the pointer is changed after, say 0.1 seconds if the document hasn't been loaded by then.

uliska:
> I like the idea of a real batch import, leveraging the job queue to run several processes in parallel.

I will merge the current PR and then continue on that track in the job queue branch. This batch import may be a better candidate to explore the job queue approach than the regular document compilation. While I have the strong impression such a multi-tab log widget might be a widget that can be shared between the two tasks in the import we don't have the overhead/complication of synchronizing the log tabs with the open editor tabs.

dliessi:
@uliska I haven't reviewed this PR, but I noticed that some comments in `frescobaldi_app/job/mozart_process.py` are in German. Can you please translate them in English? Thanks.

uliska:
I don't know why this file got added during a merge commit, sorry. I've removed it in 90d789f6

fedelibre:
> What happens when you independently open the generated .ly file from the File Open command? I'd assume the delay would be exactly the same. In that case it might be a good idea to modify the general document-opening code to display a "waiting" mouse pointer. I'd start with explicitly changing the pointer, and when that looks flickering in general cases I'd insert a delaying timer, so the pointer is changed after, say 0.1 seconds if the document hasn't been loaded by then.

Yes, the delay is the same, but there's a key difference: when I open the generated .ly file, **the file open dialog remains visible until the file content is loaded** in the editor pane. So I understands I should wait.
OTOH when I import the XML file, after the log dialog closes, nothing happens for about 3 seconds.

PR #1111, opened 2018-09-25
Change mouse pointer to "waiting" while opening documents

uliska:
Relates to #1109 

When loading large documents there may be a significant moment of
seeming inactivity. This commit changes the mouse pointer to "wait"
during that time.

**NOTE:** This only works when opening a document into the *first* tab.
Obviously *some* code inside AbstractDocument.load() must take
care of this (i.e. returning only when the document is *fully* loaded).
Therefore this pull request should *not* be merged yet.

@fedelibre please test this both with large original files and in conjunction with file import.

### DIFF
--- a/frescobaldi_app/app.py
+++ b/frescobaldi_app/app.py
@@ -26,7 +26,7 @@ The global things in Frescobaldi.
 import os
 import sys
 
-from PyQt5.QtCore import QSettings, QThread
+from PyQt5.QtCore import QSettings, QThread, Qt
 from PyQt5.QtWidgets import QApplication
 
 import appinfo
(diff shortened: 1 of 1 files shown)


fedelibre:
I've tested it and it works as expected. It helps.
Thanks

uliska:
Maybe, but I'm of course not satisfied by the fact that it only works for the first tab ...

I have the suspicion that this delay is related to the known problem of too-long-running tasks to update the autocompletion and highlighting data. So maybe when *that* will have been fixed the situation might become different: Files would be opened quickly, with the highlighting kicking in delayed.

PR #1115, opened 2018-10-09
#1066: Respect Logtool font size setting

### LINKED ISSUE #1066: log fontsize change has no visible effect
The log font and fontsize can be changed at Edit→Preferences→Tools. These settings can be found in the file `frescobaldi.conf`:

```
[log]
fontfamily=Inconsolata
fontsize=16
```

But the fontsize setting seems to have no effect.

uliska:
Closes #1066

There seems to be a conflict between letting the log font be zoomed
and respecting the Preference for font size.

So this makes the Log Tool consistent with how font sizes are managed
in the document editor. But the ability to *zoom* the log with the
mouse or keyboard shortcuts (?) is removed by that commit.

@mmeyn

### DIFF
--- a/frescobaldi_app/log.py
+++ b/frescobaldi_app/log.py
@@ -120,8 +120,6 @@ class Log(QTextBrowser):
 
         output = QTextCharFormat()
         output.setFont(outputFont)
-        # enable zooming the log font size
-        output.setProperty(QTextFormat.FontSizeAdjustment, 0)
 
         stdout = QTextCharFormat(output)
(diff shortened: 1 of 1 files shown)


fedelibre:
It works fine.

I think that we can lose the ability to zoom the log (hidden and not much useful feature IMO) if that means let the font size set in the Preferences work as expected.

uliska:
That's what I think too.

But I'd appreciate one or more addition comments before merging ...

uliska:
Well, if no objections arise ...

PR #1126, opened 2018-11-12
Prepare to upload version 3 on PyPI (pypi.org)

fedelibre:
Set the minimum version requirements for Python and python-ly.
Fix Classifier name as per https://pypi.org/classifiers/

PR #1127, opened 2018-11-14
Require Python 3.3

uliska:
I stumbled over [`main.patch_pyqt`](https://github.com/wbsoft/frescobaldi/blob/master/frescobaldi_app/main.py#L115-L147) and think this can/should be dropped. Essentially this was necessary for Python2 and should *not* be necessary for Python3 anymore.

Based on the code in the function this "patch" was necessary until Python 3.2, but given the fact that Python 3.3 was released in 2013 and has "reached end-of-life" (see https://www.python.org/download/releases/3.3.3/) I think it's safe to drop it altogether and bump the requirement to Python 3.3.

This is also related to #1126.

I'm pretty sure there's no problem with this but wanted to ask for feedback anyway.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,8 @@ ChangeLog of Frescobaldi, http://www.frescobaldi.org/
 
 Changes in 3.0.1 --
 
+* Requirement changes:
+  - Frescobaldi now requires Python3.3+
 * New features:
   - New "First System Only" option in Custom Engrave
   - Goto Line command (#927, feature request #676)
--- a/frescobaldi
+++ b/frescobaldi
@@ -1,6 +1,23 @@
 #!/usr/bin/env python3
 import sys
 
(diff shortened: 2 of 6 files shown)


uliska:
Ok, good point. I'll do so when I"m back home. Tuesday if I remember picking up ...
-- 

uliska:
@proski I have added a check that will prevent any attempt to load the application when the Python version is too low. Could you please have another look at the other commits I've added? I have changed all cases where `sys.version_info` was checking for Python2, which is not an issue anymore. But in the last case (snippet) I'm not 100% sure I fully understand the code I've changed. (The conditional is not the issue but I'm not fully clear about the `exec` here).

uliska:
Since I was 99% sure about it I'm merging this. It can still be reverted or fixed otherwise if someone should find a problem.

Issue #1069, opened 2018-06-11
automation (Make, Meson/ninja, etc.)


fedelibre:
It would be nice if Frescobaldi could be able to run external commands.
This is probably something that could be developed as plugin/extension, as discussed in issue #599.

Some ideas from other editors:

- [Atom build package](https://atom.io/packages/build)
- [Visual Studio Code tasks](https://code.visualstudio.com/docs/editor/tasks#vscode)

While running an external command might be "easy", I'm not sure how to handle the PDF preview in case the output files are generated in directories other than that of the respective source file.

uliska:
How do you think such external commands should be configured? Do you think on running external commands on the current document? Or rather completely independently like `make`?

I'm thinking if this can't be achieved with snippets? I mean, it should be possible to start external commands from a snippet, and would providing some example snippets already satisfy your needs?

fedelibre:
@marnen Article is delayed because I started another article which should be published first. And I'm on holidays now... I'll share it as soon as it's published, but it will take some time.

There are several reasons why a build tool might be handy. Let me highlight only those related to Frescobaldi:

1. **Reliability** - Frescobaldi doesn't make any difference between PDF files containing point-and-clicks and PDF files ready for publication or sharing. Both are located in the same directory and have the same name. When I use a build tool, I know that PDF files in the same directory as the source file are generated by Fresco and contain point-and-click, while PDF files in the build directory are the final files I can share or print.

2. **"Mnemonics" and collaboration** - Yes, Fresco can use custom commands, but you might need different complex commands for different projects. You might also want to share these commands with other collaborators, so better to have them in the build tool file.

3. **Efficiency** - Fresco can engrave at once all the files opened in current session. This is handy, but a) it still requires you to open all the files and b) it might lose time and resources to generate files that might not need to be recompiled. A build tool integrated in Fresco would solve these two limitations.

Jim-Duke:
Lurker here jumping in.  I can add several other reasons for automation:

1. **Post Processing** - I often want to do things to the products produced by Lilypond to transform them into formats not supported by Lilypond.  For example, most of my scores also produce a number of "Practice Tracks" for choral singers - and I use a command line MIDI to MP3 converter for those.

2. **Complex Projects** - I have a library of hymns, and choral works.  Many hymns and choral works are assembled into different packages for different performances / purposes.  I like to define those and use build tools - which specialized in such things - rather than do it manually using Fresco.

[...]

Issue #599, opened 2015-01-28
Supporting External Libraries


uliska:
This issue was partially discussed in #592 but it was (rightly so) suggested to open a dedicated issue for it.

There are things an editor/IDE could do with LilyPond scores that is not in the Python domain but that works on the LilyPond side. The basic approach is somehow injecting LilyPond code in the to-be-compiled score.

As an example we have the Layout Control Modes that work that way. For this we have some LilyPond files in the `layoutcontrol` directory. But this approach works only because we don't require any code to be persistent in the LilyPond files.

However, there are things that only work together with code inside the LilyPond files. One example (from my current interest) are annotations that can be entered in LilyPond files using my new [ScholarLY](https://github.com/openlilylib/scholarly) library. I would like to add an GUI for these annotations (annotation editor, annotation browser, context menu integration etc.) because it would be absolutely terrific to have such annotation editing integrated in Frescobaldi
Of course we could include the relevant library code into Frescobaldi like we did with the Layout Control Modes. But that would mean that the files can only be compiled with Frescobaldi - which is of course an inacceptable situation.

So the idea for a solution is to "support" external libraries. That means Frescobaldi "knows" whether a user has a certain library installed and can accordingly provide functionality. So if Frescobaldi knows that, say, ScholarLY is available it can enable related functionality such as the annotation editor and otherwise hide it in some way (by removing menu items or panels, or by greying them out).

There are three approaches we could consider:
- Create a dedicated "Frescobaldi library" and make it publicly available separately from Frescobaldi (like python-ly). This may be the cleanest solution.
- Support _one_ existing library, presumably openLilyLib. The problem with this is that openLilyLib is by now rather shapeless, and through its quite open attitude it's also comparably volatile.
- Support arbitrary libraries. This would mean that any given functionality would depend on a given library. The responsibility for such support would be on the side of the Frescobaldi developers, not on that of the users.

It may be of interest in this context that currently we also have discussions about creating specifications for writing libraries in a consistent way. The hope is that this encourages people to upload their personal solutions and make them more generally available. The overall hope is that one day we'll have an ecosystem of extension packages similar to that of LaTeX (although much smaller of course).

PaulMorris:
[...]

PR #1129, opened 2018-11-27
Extensions infrastructure

uliska:
I have finally started to work on an extension infrastructure allowing to integrate external code in Frescobaldi (which has been discussed in #599, #592, #1069 and probably elsewhere too). This start was motivated by personal need to be able to add custom (partially private) functionality without having to actually fork the repository.

This PR is as small as it could be. I intended to propose the changes as early as possible, but it turned out that it was necessary to create an essentially working version for a first step to make any sense. But I hope it goes in the right direction.

In order to test this PR you have to do the following (in addition to checking out the branch, of course):

* Create one arbitrary directory that will hold all future extensions (e.g. `~/git/frescobaldi/extensions`)
* Clone https://github.com/frescobaldi-extensions/sample-extension.git into a subdirectory of that folder
* Go to the new "Extensions" page in Frescobaldi's Preferences dialog and store the above extensions directory.

Then there should be an `Tools=>Extensions` submenu available where all extensions will expose their (optional) Tool panels and additional actions:

grafik

Custom selections from the available actions are made available in the context menus of the editor and the music view (so far, other hooks to expose extension functionality in the GUI should be discussed and implemented later).

Both the files in the `extensions` module and the sample extension are heavily commented to help understand what I did.

Basically an extension is a Python package that implements an `Extension` class in its `__init__.py`. Optionally it can provide an action collection and a tool panel (from which at least one must be present). For all required classes appropriate base classes are available that bridge between the usual Frescobaldi classes and the extension, in addition a number of properties are passed through for the extension to directly access (e.g. the current document or its cursor).
The basic tasks of creating an extension are very small and simple, and then the extension has full access to Frescobaldi's modules and functionality (I assume we should place an appropriate warning somewhere to only load extensions from reliable sources because they will have unrestricted (non-root) access to the computer, but I don't think we should actually try to implement security like a sandbox). 

I hope this approach goes into the right direction and we can rather quickly iron out any issues to be able to merge this as a first iteration of an extension infrastructure. So far I have the impression it is already very powerful and runs smoothly.

---

Some issues I'm already aware of or that I haven't tackled yet:

The first group should probably be fixed before merging this Pull Request:

* [x] I want the Preference page to display information about the extensions (that has to be provided by some config/metadata file), at least the loaded extensions should be listed by name. Preliminary implementation in 195641407a54a2908f0e049a739082265fabda03 (should be sufficient for now)
* [x] extensions should be able to load icons from their directory. An extension icon (like a favicon) should be used, in the Preferences page and maybe at some point in the extension Panel.  
Implemented in 7c88b1f249cae6ab3df873c2791c5fad2af6780f and a6402b4dbe8bdb1378f8d4fcdcdae154a8e02431
* [x] There are glitches in the handling of settingsChanged when the root path is changed from a valid to an invalid path and back. In this case the menus have to be reloaded, and it seems that isn't done perfectly yet.  
The same is true (actually worse) when de/activating extensions altogether. Therefore this functionality is currently disabled.  
I've decided to require a restart in these cases. We don't know how (3rd-party) extensions will behave, and it is impossible to "unload" a module from the running application. edf93d82bc18d3bbac8f895ad9308a3144d3ad49
* [ ] I'm not sure if I'm correctly handling the reference to the main window. There are provisions in the code to manage more than one mainwindow instance, and I have never really understood them. IIUC the Extensions object (`app.extensions()`) should be a global one, so I'm not sure what would happen if someone actually has multiple MainWindow instances.

The others may be deferred to later development (and should be added as issues):

* [ ] It would be good to extend the localization support to extensions.
* [ ] It should be possible to assign keyboard shortcuts to extensions' actions. This must gracefully handle the case when extensions are *not* available anymore.
* [x] The preferences should allow the deactivation of extensions in general 
* [x] It should be possible to deactivate individual extensions (c840b4f24c9962c6f3a5012174ec9c2296f7d4fd)
* [ ] Also there should be a means for custom sorting of extensions (how they appear in the menus)
* [x] Probably a mechanism is needed to ensure version compatibility, say, an API version number extensions can reliably refer to.
* [ ] Integrate documentation. E.g. force an extension to provide a What's This item. Or automatically integrate extension documentation in the User guide


uliska:
Hey, as you can see from the commit list I have continued working on this branch.
I've made the code more straightforward (I hope), and especially made it easier to create extensions. Also I've implemented a number of the issues I listed when originally opening the Pull Request, specifically extensions can now provide a configuration widget that is implicitly shown in the Preferences.

In parallel I've converted a custom Tool panel that I use for a private project to an extension, and it turns out that the system seems to work pretty smoothly already.

In addition to the heavily commented sample extension I intend to create another one called "tutorial". This will be much less commented (the comments actually hide the fact how simple an extension can be) but accompanied by a proper written introduction.

What I'm not completely sure about is how extensively the extension mechanism should be reflrected in the user guide.

But in general I'd say this could be ready to be merged now.

*Preferences showing the configuration widget of the sample extension:*

grafik

*Action reflecting the option set in the Preferences:*

grafik

uliska:
I've now added handling of extensions's meta information, proper (I hope) handling of faulty extensions, and the userguide pages. The user guide is *exclusively* from the perspective of the user, a guide how to *write* extensions should not be added to the userguide at all. I'll write a separate document that can be either added to the (new) website or provided as a PDF resource in the `sample-extension` repository.

I think this is now ready for a final review, any further wishes could be added later IMO. The aspect I'm not 100% comfortable with is (as noted in the bullet list in the original description) the handling of the `mainwindow`.

uliska:
To test the error handling you can do the following:

* Add a directory with arbitrary files to the extension directory  
`=>` should be ignored
* Add an empty `extension.cnf` to that directory  
`=>` A warning is displayed, information is present in the Preference page  
`=>` Additionally errors about not loaded extension
* Change a mandatory field in the `extension.cnf` of an existing extension (comment out `extension-name`)  
`=>` similar behaviour, but extension loads
* Cause an error in an extension (e.g. `import whatever` in its `__init__py`)  
`=>`extension doesn't load

uliska:
Further improvement: significantly simplify handling of settings (and preventing extensions from messing around with the gobal settings/preferences).

uliska:
Minor polishing and small additions while re-creating a fairly complex private extension. I really think this is good to go now.

uliska:
After quite some pause I returned to this branch and made some final (?) improvements and polishing.

I will now also complete the mentioned custom extension and see if this indicates the need for more changes, but since there seems noone ready to test this I'll take the liberty to eventually merge this branch without explicit review.

PeterBjuhr:
I'll try to manage to review it this weekend! But it's pretty comprehensive... :slightly_smiling_face: 


.

uliska:
Indeed. I'm sorry about that but it wouldn't have been possible to do it differently. Originally I had intended to *only* prepare some very first steps, but it turned out to become a pretty comprehensive solution that I have just bumped to a `0.9.0` API version statement ;-)

It would have probably been better if this had been developed by a team/pair, but I somehow feel pretty confident that I didn't produce too much nonsense ...

frescobaldi_app/app.py
```diff
@@ -228,6 +228,21 @@ def displayhook(obj):
     if obj is not None:
         print(repr(obj))
 
+
+_extensions = None
+
+def load_extensions(mainwindow):
+    """Called from MainWindow.__init__()"""
+    import extensions
+    global _extensions
```
PeterBjuhr: The use of global variables gives me a warning sign. But in this case it's following the rest of the code and could perhaps be seen as a constant object?

frescobaldi_app/extensions/__init__.py
```diff
@@ -0,0 +1,914 @@
...
+        """Returns a reference to the main window."""
+        return self.parent().mainwindow()
+
+    def text_cursor(self):
+        """The text cursor of the current document."""
+        return self.mainwindow().textCursor()
+
+
+class ExtensionMixin(QObject):
+    """Convenience mixin class for widgets in an extension.
```
PeterBjuhr: @uliska Sorry I don't have the time to dig into the details of the code! Could you reply more about why you are choosing to use both mixins and inheritance?

frescobaldi_app/app.py
uliska: Yes, that was the rationale: implement the caching like with other comparable objects.

frescobaldi_app/extensions/__init__.py
uliska: I came to the conclusion that the functionality provided by this class had to be available *anywhere*.  What this approach solves is the problem to have it available in classes that inherit from arbitrary widgets or non-gui classes.

There are some preconfigured classes that can be used in various parts of an extension, but *any* class, be it a widget or any other class can additionally inherit from `ExtensionMixin` to "attach" itself to the extension (usually for direct access to the `settings()` and other properties.

Does that help?

PeterBjuhr: Thanks! I have not a clear overview of this (yet). It just seemed to me to be an unusual design choice.

uliska:
@wbsoft I think this PR should be considered beta quality code, but I suggest to merge it now. I think it will only be possible to iron out remaining flaws or determine remaining features that are necessary by creating some real-world extensions. And this will only be possible when the API is at least in the master branch.

What do you think? I would prefer having some feedback on this understanding before I merge the branch.

uliska:
OK, after two days I finally merged this, knowing that there will probably be more work down the road...

Issue #1088, opened 2018-08-05
Auto-complete hinders tweaking `Beam.thickness`


sincere-music:
When I type

    \tweak Beam

the auto-completion suggests a list of words, first `beam-thickness` and last `Beam`. I don’t need auto-complete at this moment, so I just type in the dot, but at that point it’s auto-completed to

    \tweak beam-thickness.

and I have to work around by typing on, then returning to change `beam-thickness` into `Beam` again.

pedro-psb:
I tried to reproduce it here. (Windows 7 / Fresco 3.0.1)
If I type "\tweak Be" just "Beam" and "BendAfter" pops out.
The "beam-thickness" just pops if I start with lowercase "be".

uliska:
This is because the behaviour has changed since the last release.

When you want to participate in the development (even by "only" suggesting and testing) you should consider running Frescobaldi from its Git repository as described in the Wiki: https://github.com/wbsoft/frescobaldi/wiki/Run-Frescobaldi-3-from-Git-on-Windows
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

uliska:
@sincere-music I have added this to the "project" of the next release. As the autocomplete has been significantly modified (and in general I'd say improved) we should iron out this kind of glitch before the next release.

uliska:
@sincere-music I just stumbled over this issue, and I'm unsure what to make of the various commits I have added in the meantime.

Could you please have a look at how the autocompleter behaves right now and give me some feedback if there are more possible issue with it?

PR #1132, opened 2018-12-16
#1088 Autocompleter: Don't treat dot like ENTER

uliska:
#1088 describes how the fact that entering a dot makes editing harder
in cases when there is no completion intended (`\tweak Beam.` should
*not* complete to `beam-thickness`)

This commit removes the link from Enter/Return to `.`. That means that
entering the dot will *not* automatically complete to the first match
in the completer popup. If that is wanted one would now have to enter
ENTER plus the dot. This need for one more keystroke seems more than
worth it, considering less unintended completions.

### DIFF
--- a/frescobaldi_app/widgets/completer.py
+++ b/frescobaldi_app/widgets/completer.py
@@ -56,7 +56,7 @@ class Completer(QCompleter):
             # a key was pressed while the popup is visible
             cur = self.textCursor()
             modifier = QApplication.keyboardModifiers()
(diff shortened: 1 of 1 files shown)


uliska:
@sincere-music what do you think, is this the "correct" solution or only a patch for one specific issue that might produce other problems down the road?

sincere-music:
I really don’t have much of a larger perspective on this, it was just one case I stumbled upon. But aside from that, it seems a reasonable change. Zealous autocomplete mechanisms can quickly become unhandy, so the simpler the better IMO (simpler meaning fewer automatic reactions that go on to make behaviour less intuitive).

Issue #1147, opened 2019-02-07
button to open LilyPond documentation in the browser


fedelibre:
I suggest adding a *browser icon* in the Documentation Browser's toolbar to allow to open the LilyPond documentation in the default browser (Firefox, Chrome, etc.).

In fact browsing the LilyPond documentation within Frescobaldi is useful sometimes but can be limited.

@uliska what do you think? I've marked it as beginner issue.

uliska:
Yes, nice idea, and I agree that this usxa good "beginner" issue.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

urielo515:
could i try and develop this?

uliska:
Of course, great!

Do you need pointers to get started at all or would you like to look for yourself first?
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

urielo515:
nice. i'm checking the code right now, but it's really large. if you could just point me to where the documentation browser toolbar is written, i guess i can work from there.

uliska:
Ok, I"LL have a look when I'm back at the computer.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

uliska:
The documentation browser is (AFAICT) completely within the directory `frescobaldi_app/docbrowser`.

`__init__.py` provides the integration in the Tools system, the `HelpBrowser` class is not the browser itself but the "tool panel" (inheriting `panel.Panel`). The `createWidget()` method of such a panel descendant tells you where the actual visible widget is defined, namely in `browser.Browser`. So this `Browser` class is what you'll have to start looking at.

You will have to add an action in `__init__`, and you can access that in `browser.Browser` through the `ac = dockwidget.actionCollection`.

Don't hesitate to ask more ...

listentolist:
When I right-click inside the Documentation Browser and choose to "Open Document in New Window" it already opens the current page inside the default browser. So do we really need a toolbar button for that? Or should the toolbar button link to the start page of the documentation instead of the current page? Anyway, if no one is working on it at the moment, I'd be interested in it.

PR #1153, opened 2019-05-29
1147 button browser

listentolist:
#1147 Added a button to the Documentation Browser's toolbar for opening it in a web browser and an icon from the Tango Icon Library.


uliska:
I have two questions about this:

1. It is possible (with reasonable effort) to open the currently viewed page in the external browser?
2. It would be nice (and more consistent) to also have that Action in the context menu

To achieve this one should a) change the Action to open the current page, and b) replace the Viewer's builtin (I assume that is made available thorugh Qt) with a new one with that Action.

@listentolist Do you think that's possible or would you like us to look into it and give you a pointer?

listentolist:
@uliska Thank you for your fast reply. It should be possible. But don't we have that feature already in the context menu? When I choose "Open Document in New Window" or "Open Link in New Window" from the context menu, it opens the link in the the web browser, that is specified in the "helper applications". Maybe that should be renamed in "Open in Web Browser".
So the button in the toolbar should link to the currently viewed page rather than the start page?

uliska:
I'm not 100% sure right now what would be the proper solution. One thing though is that I think that the button and the context menu should correspond, so both should either open the current document or the home page. And it would be preferable if they had the same icon.

One alternative suggestion would be to add *two* actions to the button (like we have with the "save" or the "compile" buttons in the toolbar), one opening the current and one opening the home page.

uliska:
I'm sorry I haven't looked into this earlier. But can you confirm it works properly on your side? When trying it out I always get "not found" errors, both inside the documentation browser and outside. This *may* be related to something already present earlier.

The problem I see is that "Documentation" is added to every URL in `lilydoc.documentation.Documentation`, but the documentation *home page* does *not* have this path segment. So the actual *home page* is found at `http://lilypond.org/manuals.html` or `http://lilypond.org/development.html`. So IISC one should differentiate between getting the URL for the actual home page and providing the "root" for arbitrary other URLs with the `Documentation` injected.

Would you like to update your patch? Or maybe I'm misunderstanding something (I must admit I have never actually used the documentation browser.

uliska:
See #1163 1163

listentolist:
I'm sorry, that my answer is so late. It actually works on my side. @uliska I failed getting your error... 
Do you have any entries in the Preferences->Lilypond Documentation (I have got none)?
Which version is chosen in the chooser inside the documentation browser (I have only one possibility: "2.18.2 (lilypond.org)")?
Does the documentation browser work normally (so does this patch break it, then I know where to look)?

fedelibre:
@listentolist Could you please rebase your branch on current master? I'd like to test your PR.

fedelibre:
I've rebased on current master and pushed. I hope you don't mind @listentolist 

I like your patch and it would be nice to have it in the next release.
As far as I can see, it works and I see that you've addressed the requests made by Urs in the first comment.

+1 for me
Waiting for the approval from @PeterBjuhr and @Jean-Abou-Samra

listentolist:
@fedelibre Thank you!

fedelibre:
@jeanas @PeterBjuhr Can you please comment on the code? I like this PR, as it makes more clear what the button actually does. I'd like to have it in 3.3.

PeterBjuhr (approved):

Issue #1151, opened 2019-04-19
Add open documents to tab context menu


uliska:
I've come to enjoy a feature of TeXstudio: right clicking on a document's tab provides access to all open documents. I think this would be nice to have in Frescobaldi too because I find it very natural and straightforward to look for open documents (at least more natural than finding the "Documents" menu.

What I would do is (differently than in TeXstudio) create a submenu in the context menu containing the open documents.

What do you think? (@wbsoft)

uliska:
Closed through #1154

PR #1154, opened 2019-05-30
Added a submenu for open documents in tab context menu

listentolist:
Issue #1151


wbsoft:
@listentolist and Urs, thanks! Very useful addition.

PR #1166, opened 2019-09-05
Fix python-ly error message bug

techdude1996:
When running without python-ly installed, python throws an error: NameError: name 'ly_required' is not defined in main on line 145. Changing ly_required to appinfo.required_python_ly_version fixes this error.


uliska:
Thanks!

Issue #1164, opened 2019-09-01
"Show Available Fonts" features


uliska:
This is a follow-up and check issue for #1075 . I want to keep track of features that should be fixed before that PR can be merged and features that may be added later (preferrably before a 3.1 release)

#### Before merging

* [ ] More music examples (classical, score)
  *(Check mailing lists, I think there have already been contributions)*
* [ ] "Glyph sample list" (https://github.com/frescobaldi/frescobaldi/pull/1075#issuecomment-459260351 and https://lists.gnu.org/archive/html/lilypond-user/2019-09/msg00004.html)
* [ ] copy the resulting LilyPond code to the clipboard or include at the current position in the current document.
* [ ] Check implementation of "guard against < 2.19.12."

#### Maybe included in current PR

* [ ] Make it possible to select text fonts "as" roman|sans|typewriter
* [ ] Use the text fonts in the preview
* [ ] Do something about the issue of paper blocks in "custom"/"current" choice (a paper block in the document will nullify the font selection).

#### Later

* [ ] Caching/Precompilation of examples (https://github.com/frescobaldi/frescobaldi/pull/1075#pullrequestreview-180879129) and https://github.com/frescobaldi/frescobaldi/pull/1075#issuecomment-443907555)

fedelibre:
@uliska I'm in a hurry so I'll write quickly a few comments without checking too much.

I've tested your branch now but I had to reset hard to commit 71b8413f33dd6dfe90bcf02c3a5cea9e8bcc1b85 because otherwise I could not start Frescobaldi. Yesterday's commits may have broken something.

I see the examples already built are now cached, wonderful! I still see a blue "progression bar" at the bottom when I come back to an example but it's not a big deal.

Installing the fonts in LilyPond is very easy. What about offering also a way to download the music fonts? I guess you've already thought about it... Is [OpenLilyPondFonts](https://github.com/OpenLilyPondFonts) a good resource?

uliska:
> I've tested your branch now but I had to reset hard to commit 71b8413 because otherwise I could not start Frescobaldi. Yesterday's commits may have broken something.

OK, don't bother to do further testing before I push the work I've done in the meantime but not committed yet. It "works for me" so I'll be curious about any issue reports.

> I see the examples already built are now cached, wonderful! I still see a blue "progression bar" at the bottom when I come back to an example but it's not a big deal.

[...]

PR #1173, opened 2019-09-13
Document Fonts dialog

### LINKED ISSUE #420: add search to "Show available fonts"
The new "Show available fonts" command is very useful,  but it's hard to read and find what you are looking for. As I use it mostly to search something, currently I prefer opening a terminal and using:

```
fc-list | grep -i keyword
```

### LINKED ISSUE #1077: Available Fonts: Show samples
When we have the list of font families and styles in a QTreeView since #1075 it would be really nice to display a font sample in a second column.

I think this shouldn't be too hard, but I suggest we don't block the Pull Request on that additional feature.

One thing I think we should make sure is that the styled font samples are only generated when an entry is opened. This may mean that the first entry will *always* be only the family name.

uliska:
Closes #420 
Closes #1077 

This Pull Request replaces #1075 by squashing the long and inconsistent development history to a few commits. Please refer to that closed PR for discussion of features and issues (mainly from the bottom up because the initial comments refer to a state that doesn't even closely resemble the current results.

This provides a new dialog `Tools=>Document Fonts...` that allows 

* browsing available text and music fonts
* managing music font installation for a given LilyPond instance
* Generating and using font setting commands with various strategies
* Previewing fonts in (cached) scores

Note that besides the dialog there are also options on the `Preferences=>Paths` page, and that there's rather extensive documentation in the user guide.

Although there are more wishes open (#1164) I'm convinced this is now ready to be merged.


uliska:
If noone raises objections I'll merge this tomorrow.

uliska:
Frescobaldi-font-schenker

wbsoft:
Thanks for this great work!

PR #1185, opened 2019-09-21
Dev/urs/jobqueue (WIP)

uliska:
This is not intended for merging since the functionality is only halfway implemented so far. But I'd like to have it tested to see whether that half has been done properly, and maybe for some discussion about the way forward.

This builds on the multi-process job queue I had already implemented last year but starts making it properly available within Frescobaldi.
A global job queue manages a fixed number of runners that can execute external jobs in parallel. The point is not to *enable* parallel processing (which was possible since the beginning) but rather to *limit* it and control the use of resources. With the job queue it's possible to start literally thousands of jobs without either overwhelming the OS in managing them or having to execute them one by one. In a project I've successfully used that approach to have six cores permanently busy compiling hundreds of snippets.

There are three built-in queues, "engrave", "generic" and "crawl" that will be used for LilyPond jobs, general ones (like importing or convert-ly, or other jobs that extensions might launch), and I want to have a one-runner queue for "crawler" jobs to permanently do stuff like parsing autocomplete information etc.

Queues can request dedicated runners for exclusive access, and all remaining runners are shared between all queues. The three built-in queues are "continuous", i.e. running for the whole Frescobaldi session while custom queues can be set up as "single" to enqueue a fixed number of jobs and "finish". Such single queues will have the option (not implemented yet) to request more dedicated runners which are automatically released when the queue is finished. This is a way to temporarily give a batch process more resources at the cost of the overall operation.

Currently this is only partially implemented in so far as all three queues only use "shared" runners, and I haven't really thought about how the distribution of resources will be managed, and how the user is educated when making invalid requests (will there simply be a message saying "I don't have 27 cores at my disposal" or will be attempt a dynamic assignment etc.).

For testing please try out what happens when multiple jobs are started (including compiling multiple documents from tabs and starting processes like the score wizard preview, convert-ly, imports, or the document fonts dialog).
I suggest starting with Preferences=>Multicore=>Cores set to 1, which will force Frescobaldi to only use one external process at a time. So when you start more than one job consecutive jobs will only start when the previous job has finished. OTOH when you have more than one core the corresponding number of jobs will run in parallel.
*NOTE: While this may seem limiting it is an effort to prevent a loss of responsiveness due to overwhelming the OS with LilyPond processes.*

Note the small enhancments of the Log messages ;-)

See also https://github.com/frescobaldi/frescobaldi/issues/1176 for some things I have in mind for this functionality. Two really important use cases are:

* reserving one permanent runner for the background tasks that currently make Frescobaldi unresponsive with large and/or complex documents (as was confirmed to me a few days ago this can mean waiting minutes after each keystroke, which is totally unacceptable)
* providing support for batch processing, e.g. (one feature that may be added to Frescobaldi itself, everything else will be in extensions) "compile all `.ly` documents in the current directory.

### DIFF
--- a/frescobaldi_app/app.py
+++ b/frescobaldi_app/app.py
@@ -26,7 +26,10 @@ The global things in Frescobaldi.
 import os
 import sys
 
-from PyQt5.QtCore import QSettings, QThread
+from PyQt5.QtCore import (
+    QSettings,
+    QThread
+)
 from PyQt5.QtWidgets import QApplication
 
 import appinfo
@@ -58,6 +61,39 @@ sessionChanged = Signal()       # (name)
 saveSessionData = Signal()      # (name)
 jobStarted = Signal()           # (Document, Job)
 jobFinished = Signal()          # (Document, Job, bool success)
+
+
+# Number of phyiscally available CPU cores (if determined)
+_cores = QThread().idealThreadCount()
+if _cores == -1:
+    _cores = None
+
+def max_cores(fallback=True):
+    """
+    Return the number of physically available CPU cores.
+    If fallback=True (and the number can't be determined)
+    return a fallback of 16.
+    """
+    return _cores or (fallback and 16)
+
+def default_cores():
+    """
+    Return a default value for available cores
+    that is used in several places as long as the
+    user hasn't stored the max cores explictly.
+    """
+    return max(max_cores() - 1, 1)
+
+_available_cores = None
+
+def available_cores():
+    """
+    Return the number of requested runners for Frescobaldi.
+    """
+    if not _available_cores:
+        s = settings('multicore')
+        return s.value('num-cores', default_cores(), int)
+    return _available_cores
 
 
 def activeWindow():
--- a/frescobaldi_app/convert_ly.py
+++ b/frescobaldi_app/convert_ly.py
@@ -230,11 +230,11 @@ class Dialog(QDialog):
 
     def slotJobDone(self):
         j = self.job
-        if not j.success and j.failed_to_start():
+        if not j.success() and j.failed_to_start():
             self.messages.setPlainText(_(
                 "Could not start {convert_ly}:\n\n"
-                "{message}\n").format(convert_ly = j.command[0],
-                    message = j.error))
+                "{message}\n").format(convert_ly = j.command()[0],
+                    message = j.error()))
             return
         out = j.stdout()
         err = j.stderr()
--- a/frescobaldi_app/convert_ly.py
+++ b/frescobaldi_app/convert_ly.py
@@ -217,11 +217,11 @@ class Dialog(QDialog):
 
         self.job = j = job.Job(command, encoding='utf-8')
         if QSettings().value("lilypond_settings/no_translation", False, bool):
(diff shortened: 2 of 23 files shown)


uliska:
@wbsoft a more specific question. `app.job_queue()` emit a number of signals:

https://github.com/frescobaldi/frescobaldi/blob/678c17ed5de41bb53f50edba5aa8399302239099/frescobaldi_app/job/queue.py#L786-L812

where I feel the latter are somewhat parallel (and redundant?) to the `app.jobStarted` and `app.jobFinished` signals. The main difference is that the previously existing signals are specific to *editor document* jobs while the new signals are emitted for *every* job.

There are only a few places where the job signals are used:

https://github.com/frescobaldi/frescobaldi/search?q=app.jobStarted&unscoped_q=app.jobStarted and https://github.com/frescobaldi/frescobaldi/search?q=app.jobFinished&unscoped_q=app.jobFinished

Do you think there's anything that speaks against adjusting these places to use the new job queue signals? It's easy to "filter" the signals to only respond to document jobs if that should be necessary.

wbsoft:
> Do you think there's anything that speaks against adjusting these places to use the new job queue signals? It's easy to "filter" the signals to only respond to document jobs if that should be necessary.

I had not the time to look thoroughly in the new job queue code, but I think it is no problem. The app.jobStarted and jobFinished signals are quite old, and I want to innovate stuff anyway: as soon as LilyPond writes a file, I want to know it, instead of waiting till LilyPond actually finishes the process. Sometimes a process writes many small PDF documents, and I like to get the updated view as soon as an output file has been written.

uliska:
Am 22. September 2019 20:28:08 MESZ schrieb Wilbert Berendsen <notifications@github.com>:
>> Do you think there's anything that speaks against adjusting these
>places to use the new job queue signals? It's easy to "filter" the
>signals to only respond to document jobs if that should be necessary.
>
>I had not the time to look thoroughly in the new job queue code, but I
>think it is no problem. The app.jobStarted and jobFinished signals are
>quite old, and I want to innovate stuff anyway: as soon as LilyPond
>writes a file, I want to know it, instead of waiting till LilyPond
>actually finishes the process. Sometimes a process writes many small
>PDF documents, and I like to get the updated view as soon as an output
>file has been written.

I think that would require adding a new signal within job.lilypond.LilyPondJob that parses the output for file related messages. However, thst seems to be unreliable as it would have to take the language into accoung and know all languages.

-- 

wbsoft:
> I think that would require adding a new signal within job.lilypond.LilyPondJob that parses the output for file related messages. However, that seems to be unreliable as it would have to take the language into accoung and know all languages.

Another possibility could be to watch the folder for newly created files while a job is running.

uliska:
Would that increase the chance to catch *all* output files, even with non-standard names? Wasn't that a cause of complaints by some users?
-- 

wbsoft:
> Would that increase the chance to catch *all* output files, even with non-standard names? Wasn't that a cause of complaints by some users?

Indeed. In #1186 I try to work this out further :-)

uliska:
I will reopen this (or a new one) when I feel it's ready.

PR #1193, opened 2019-09-26
File new/close/save actions

uliska:
* Add a new action "close all files" (but not session)
* Add additional actions to the file toolbuttons

Two comments for further consideration:

* I would suggest to remove the Score Wizard action from the toolbar, but I can imagine that people don't want to do that.
* Adding the same list of actions in several places (menu and toolbutton) seems redundant and prone to errors. Maybe there should be groups holding, say, all "file save" actions, which then are added "all" to a given menu or toolbutton.


uliska:
> I like the distinction between close all and close all and session.

Good, this is something I have long been missing.

> But I'm reluctant to add a menu to the New document toolbar button. I think in most cases a user wants an empty document when hitting New

But that's what would happen when simply hitting "New" (rather than waiting for the pull-down)?
I was thinking that *if* we start providing additional functions in the pull-downs it should be rather comprehensive than random.

> Also, the Save button should not get too many options, only the most used.

I will not force my point-of-view, but I find it confusing if the functionality in the pull-down is not consistent with the - corresponding - functionality from the menu.

While we're at that, I have the issue that the "File" menu has become so long that it barely fits on the screen.
I will make another commit suggesting a more hierarchical layout, for now without making changes to the previous commit. From there we can reconsider the issue: I think when functions are arranged in submenus to the File menu it may be more natural (both implementation and UI) to have the *same* submenus attached to the toolbuttons.

wbsoft:
Currently, the New button is just that, New; the Open button has recent documents in the pulldown, and the Save button Save as and Save all. Maybe it's better than to remove the pulldown of Save. Keeps it simple and not too redundant.

Indeed the File menu grows too large....🙂

uliska:
I've made the addition of pull-down menus optional.
While at it I combined the three file/sessions-related sections in `Preferences=>General` into one TabWidget, which saves much space on that page.

wbsoft:
Nice! (Side note: I think we must wake for too many config options, the toolbar pulldowns is a small issue. Maybe this is overkill. But I think this is a nice preference.)

Regarding the pulldowns change on settings change: it should be possible to move the settings-specific modifications to the actions in the readSettings() and writeSettings() in MainWindow, without altering the toolbars.

The tabwidget in the prefspage is very nice.

uliska:
Can this now be merged?

wbsoft:
Yes. Thanks!

uliska:
I have squashed the branch to two commits 7e5e51c46735da96d73790384403724e912c36b1 and 9b85d18110fd74c32ac2225c5d5451d5d9bf745e without a merge commit.

wbsoft:
Now that there is a submenu File→New (advanced), the templates in the old New from templace submenu could be added directly to that submenu, instead of having a second submenu.
Or just rename "New (advanced)" to "New from template", and have the scorewizard on top of that submenu, then a horizontal line, the templates, another horizontal line and the "Manage templates..." option.

uliska:
Ok

PR #1205, opened 2019-10-17
Handle availability of audio export (WIP)

### LINKED ISSUE #1175: Provide other tools than "timidity" for midi2wav
Currently "timidity" is hard-coded in the code for exporting midi to wav:
https://github.com/frescobaldi/frescobaldi/blob/4e42feebe3cc9f0aec396538ad6f26f79f4dd80f/frescobaldi_app/file_export/__init__.py#L115
This should be made configurable in `Preferences=>Helper Applications`.

uliska:
Closes #1175 (when complete, this will add support for multiple audio converters)

So far:

* check availability of supported audio converters
  (prepare support for additional converters)
* check presence of MIDI files *before* triggering the action

TODO: also call the check when the current document has changed (@wbsoft sorry for the silly question, but what signal should I connect to when I want to check any time the current document has changed?)

### DIFF
--- a/frescobaldi_app/file_export/__init__.py
+++ b/frescobaldi_app/file_export/__init__.py
@@ -49,6 +49,9 @@ class FileExport(plugin.MainWindowPlugin):
         actioncollectionmanager.manager(mainwindow).addActionCollection(ac)
         ac.export_musicxml.triggered.connect(self.exportMusicXML)
         ac.export_audio.triggered.connect(self.exportAudio)
+        app.jobFinished.connect(self.set_export_audio)
+        # TODO: Also connect to a signal that is emitted
+        # whenever the current document has changed.
 
     def exportMusicXML(self):
         """ Convert the current document to MusicXML """
@@ -78,13 +81,8 @@ class FileExport(plugin.MainWindowPlugin):
         """ Convert the current document to Audio """
         mainwin = self.mainwindow()
         doc = mainwin.currentDocument()
-        midfiles = resultfiles.results(doc).files('.mid*')
-        if not midfiles:
-            QMessageBox.critical(None, _("Error"),
-                    _("The audio file couldn't be created. Please create midi file first"))
-            return False
         orgname = doc.url().toLocalFile()
-        midifile = midfiles[0]
+        midifile = resultfiles.results(doc).files('.mid*')[0]
         wavfile = os.path.splitext(orgname)[0] + '.wav'
         caption = app.caption(_("dialog title", "Export Audio File"))
         filetypes = '{0} (*.wav);;{1} (*)'.format(_("WAV Files"), _("All Files"))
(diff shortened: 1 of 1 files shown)


uliska:
My last comment means: In order to test you (for now) have to load a document that produces a MIDI file, then *compile* it (even if it is already there and displayed in the music view, and only *then* go to File=>Import/Export/Export Audio. (I've added a much more informative ToolTip to the menu item)

uliska:
The second commit provides initial (non-functional) support for alternative audio file converters in so far as it detects if `timidity`, `fluidsynth` or `lame` are installed (*TODO:* Add entries for non-standard invocations in PReferences=>Helper Applications) and populates the filters in the file dialog according to the available converters.

I added `lame` to the list because MP3 export will always be high on people's wish list. To achieve that we will have to use timidity to first create the audio and then pipe it through Lame.

grafik

This shows an initial selection of file formats (duplicate formats available to both converters are automatically filtered) (note that there is no functionality available yet, right now clicking on Save will just trigger the existing Timidity export).

@fedelibre would you mind providing a list of file formats we should support for timidity and for fluidsynth? And options that can be useful for a LilyPond=>Audio export (this also for Lame)?

Note-to-self: I will move the actual code out of `__init__.py` and add separate submodules for the different tools, like in `file_import`.

fedelibre:
> @fedelibre would you mind providing a list of file formats we should support for timidity and for fluidsynth? And options that can be useful for a LilyPond=>Audio export (this also for Lame)?

I guess that WAV/Mp3 and Flac/Ogg are the most requested music audio formats.

To be honest, I've never used the Audio export feature and I commented on issue 1175 just because of a recent discussion on lilypond-user list. I suggest you write there to get good feedback from people who actually uses it.

wbsoft:
Then `MainWindow.currentDocumentChanged()` is what you're looking for :-) It gives the current document and the one that was the current document before. (A Document is only current in the context of a MainWindow :-) )

uliska:
> Then MainWindow.currentDocumentChanged() is what you're looking for :-) 

Not completely, but thanks anyway. This will not be emitted when an initial empty document is created in the first tab. See https://github.com/frescobaldi/frescobaldi/pull/1205/commits/eb8d82fc7d79d2590af1dddf0ecc051c479cf69d

uliska:
That means the "enabled" state of the export audio menu item should now be correct right from the startup of Frescobaldi.

wbsoft:
> Not completely, but thanks anyway. This will not be emitted when an initial empty document is created in the first tab. See [eb8d82f](https://github.com/frescobaldi/frescobaldi/commit/eb8d82fc7d79d2590af1dddf0ecc051c479cf69d)

Ah, I should change it that MainWindow also emits that signal on opening a new empty document. Will try that after 26 okt, because I'm a week with my wife and children on holiday :-)

uliska:
That might also help with #1111
-- 

uliska:
BTW, there should be an option to play the audio after exporting.

PR #1215, opened 2019-11-18
Changes for using QtWebengine instead of QtwebKit in SVG view

19joho66:
This is a first implementation of the SVG view using QtWebEngine. So far displaying of the note sheet in the SVG view works. Also pointandclick from notesheet to source works but not the other way around. This is work in progress.

### DIFF
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -26,6 +26,7 @@ import os
 import re
 import sys
 
+from PyQt5.QtWebEngineWidgets import QWebEngineView
 from PyQt5.QtCore import QSettings, QTimer, QUrl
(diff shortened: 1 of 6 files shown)


wbsoft:
Thanks! A good first step I think. The coming days I can look at it, I hope to be able to simplify things a bit, like the concatenation of the js scripts with the webchannel js script (is it needed to prepend that to each javascript?); also from a first glance, `editsvg.js` is not used. Importing QtWebEngineWidgets in `main.py` is not needed anymore, that already happens in `app.py`.

wbsoft:
Thanks for this work! I looked closely and implemented this PR in master, simplifying some things on the go, and removing some code that was run unneeded multiple times.

PR #1227, opened 2019-12-27
Display appinfo.version as ..-git

uliska:
When the application is run from Git show that in the
appinfo.version string (through the new
appinfo.running-version() function)

---

I'm not sure whether *all* of the changes are correct, namely in the files relating to translations and releases

### DIFF
--- a/frescobaldi_app/appinfo.py
+++ b/frescobaldi_app/appinfo.py
@@ -41,3 +41,11 @@ required_python_ly_version = (0, 9, 4)
 
 # this one is used everywhere in the application
 appname = "Frescobaldi"
+
+def running_version():
+    import app
+    if app.is_git_controlled():
+        return "{}-git".format(version)
+    else:
+        return version
+    
--- a/frescobaldi_app/userguide/resolve.py
+++ b/frescobaldi_app/userguide/resolve.py
@@ -34,7 +34,7 @@ def appname():
     return appinfo.appname
 
 def version():
-    return appinfo.version
+    return appinfo.running_version()
 
 def author():
     return appinfo.maintainer
--- a/macosx/build-dmg.sh
+++ b/macosx/build-dmg.sh
@@ -77,7 +77,7 @@ echo
 VERSION=$(${PYTHON} -c 'import os
 os.chdir("..")
 from frescobaldi_app import appinfo
-print(appinfo.version)')
+print(appinfo.running_version())')
 
 if git rev-parse --git-dir > /dev/null 2>&1
 then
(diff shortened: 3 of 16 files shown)


macosx/build-dmg.sh
```diff
@@ -77,7 +77,7 @@ echo
 VERSION=$(${PYTHON} -c 'import os
 os.chdir("..")
 from frescobaldi_app import appinfo
-print(appinfo.version)')
+print(appinfo.running_version())')
```
dliessi: Please don't change this: here I am retrieving the version of Frescobaldi that will appear in the compiled application for the release, but I will always build the release from the Git repository, so `appinfo.running_version()` will return the version with the `-git` suffix, which I do not want.

macosx/mac-app.py
```diff
@@ -45,7 +45,7 @@
   help = 'force execution even if SCRIPT does not exist')
 parser.add_argument('-v', '--version', \
   help = 'version string for the application bundle, \
-  visible e.g. in \'Get Info\' and in \'Open with...\'', default = appinfo.version)
+  visible e.g. in \'Get Info\' and in \'Open with...\'', default = appinfo.running_version())
```
dliessi: Please don't change this: see the comment about macosx/build-dmg.sh.
Here it is only the default value, and when build-dmg.sh passes the version explicitly to mac-app.py, but I don't see any advantage in changing this default.

windows/freeze.py
```diff
@@ -174,7 +174,7 @@ def copy_plugins(name):
  Flags: postinstall nowait skipifsilent;
 
 '''.format(
-    version=appinfo.version,
+    version=appinfo.running_version(),
```
dliessi: I think that the same comments as for macosx/... files hold also here: this version is used in the compiled Windows installer for the release, which is probably built from the Git repository.

windows/frescobaldi-wininst.py
```diff
@@ -111,7 +111,7 @@ def shortcut(directory):
 
 def welcome():
     """Prints a nice welcome message in the installer."""
-    print("\nWelcome to {0} {1}!".format(appinfo.appname, appinfo.version))
+    print("\nWelcome to {0} {1}!".format(appinfo.appname, appinfo.running_version()))
```
dliessi: Same comment as for windows/freeze.py.

frescobaldi_app/userguide/resolve.py
```diff
@@ -34,7 +34,7 @@ def appname():
     return appinfo.appname
 
 def version():
-    return appinfo.version
+    return appinfo.running_version()
```
dliessi: This appears to be ok.

frescobaldi_app/appinfo.py
```diff
@@ -41,3 +41,11 @@
 
 # this one is used everywhere in the application
 appname = "Frescobaldi"
+
+def running_version():
+    import app
+    if app.is_git_controlled():
+        return "{}-git".format(version)
```
dliessi: Adding also the abbreviated commit hash can be useful especially in bug reports, but the resulting string would be too long, probably.
What do you think?

Issue #879, opened 2016-12-23
add AppStream metainfo file


fedelibre:
December 2019 update: the term appdata has been replaced by metainfo in AppStream spec, so I'm changing the title accordingly. More importantly, the expected file is `$PREFIX/share/metainfo/%{id}.metainfo.xml`.

[AppStream spec](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps) allows to define some information on the project that can be read and used by different software centers (GNOME Software, KDE Discover, AppCenter).

As Frescobaldi has no AppData file, when I find Frescobaldi in GNOME Software it appears without screenshots nor any other information:

![schermata da 2016-12-23 17-27-15](https://cloud.githubusercontent.com/assets/1644647/21458233/18dbd7b6-c935-11e6-8754-1d0bc127ccbe.png)

This screenshot was made 3 years ago. As of 2019 Frescobaldi is not visible anymore on GNOME Software.

fedelibre:
I'm trying to make a patch, just to see how it works (I'll probably have to leave it to some real developer). I cannot understand how to copy the `frescobaldi.appdata.xml` file to the proper install location. I've added this line to setup.py:

```
$ git diff
diff --git a/setup.py b/setup.py
index 9af67d7..bf9e834 100644
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ else:
         ('share/icons/hicolor/scalable/apps', ['frescobaldi_app/icons/frescobaldi.svg']),
         ('share/applications', ['frescobaldi.desktop']),
         ('share/man/man1', ['frescobaldi.1']),
+        ('share/metainfo/', ['frescobaldi.appdata.xml']),
     ]
 
 setup_extra_args = {}
```

but then the file is copied to the python install directory, e.g. `/usr/lib/python3.5/site-packages/frescobaldi-3.0.0-py3.5.egg/share/metainfo/frescobaldi.appdata.xml`. But it should be relative to the install path, that is in /usr or /usr/local (or whatever is chosen by the user).

fedelibre:
Before I forget about this, for those who want to test it, this is a basic file:

```
<component type="desktop-application">
​  <id>frescobaldi</id>
​  <metadata_license>CC0</metadata_license>
​  <project_license>GPL-2.0+</project_license>
​  <name>Frescobaldi</name>
​  <summary>LilyPond Music Editor</summary>
​
​  <description>
[...]

PR #1249, opened 2020-01-13
Follow freedesktop.org standards for Linux installation

fedelibre:
Add AppStream metainfo file (fix issue #879).

Some changes in internationalization:

- Let xgettext extract also the contents in the Desktop and AppStream files.
The language is guessed by the file name extension, so we can safely remove
the --language=python option.

- The -d option of msgfmt (used to generate desktop and metainfo files) requires
all the PO files to be in the same directory.  Thus the directory layout
within i18n/ is now frescobaldi/LL.po and userguide/LL.po.
The LINGUAS file is required by msgfmt.


fedelibre:
@wbsoft I did not touch some files because I wanted to hear your feedback before proceeding. If you agree with this PR, I will add:

- a commit to move the other translation files to the new layout (there's only italian as example right now)
- a commit to update the instructions for translators

Finally, the screenshot images in the metainfo.xml file should point to a proper location. This may be discussed and fixed later.

wbsoft:
Looks great! Ok to go ahead :-)

fedelibre:
@wbsoft I've added the other translation files. Then I tested `make pot`, which worked. And `make` or `make frescobaldi/fr.po`, which did not work.
I cannot figure out why it's not working, as the pot files are more recent than the po files. And I'm sure it did work when I tested the italian po files.

Can you test it and see if you can spot the error?
Thanks

fedelibre:
Oh, silly me! The problem is in the languages variable, which used to pick each language subdir. Subdirs per languages are gone, so using the wildcard function to retrieve the list is not optimal.

I'll come up with a solution, maybe checking the LINGUAS files in each frescobaldi and userguide subdirs.

fedelibre:
@wbsoft With the latest commits, this is ready to be reviewed.

Let me know if you have any concern.

For example, the setup.py script needs `make -C linux` to be run in order to create the final desktop and metainfo files. Is it Ok? I don't think we want to track in Git both the .in templates and also the generated files.

One thing we did not discuss is how to handle the release notes.
Right now for each new release we have to copy them from the Changelog (I can volunteer for this task). I think that some GNOME projects automate this through `appstream-util news-to-appdata ...`, but I haven't investigated, also because we should change the "format" of the Changelog file.

fedelibre:
Again on release notes.
I think it's actually good to have two places, maintained separately, because they serve different purposes.

Changelog:

- is meant to be a file where each developer should drop any new feature or fix as they happen
- can contain technical details and references to issues/PRs

OTOH the release notes in the metainfo file:

- are mainly addressed to the final users (in fact software centers can show them when a new version is available or just installed)
- so we should try to adapt the language and restrict the list to the most important news (AppStream spec allows to set a `url` tag where you can link to a webpage with a more detailed list)
- the file should be updated right before a new release. As said before, I can volunteer for this.

Quoting from [AppStream spec author](https://blog.tenstral.net/2020/01/a-big-appstream-status-update.html):

> The release descriptions are short, text-only summaries of a release, usually only consisting of a few bullet points. They are intended to give users a fast, quick to read overview of a new release that can be displayed directly in the software updater. But sometimes you want more than that. Maybe you are an application like Blender or Krita and have prepared an extensive website with an in-depth overview, images and videos describing the new release. For these cases, AppStream now permits an url tag in a release tag pointing to a website that contains more information about a particular release.

In conclusion, I don't want to use converters from Changelog to metainfo.
I want to simplify the release notes I copied&pasted from the Changelog to the metainfo.xml file.
Let me add one more commit. Eventually we'll squash them :-)

wbsoft:
LGTM. I write most times an announce email, whose text can be seen as release notes.

fedelibre:
I'm happy with this PR now. Let me know if I should squash the commits or if you'll do it when merging.

PR #1258, opened 2020-02-01
Set the desktopFileName property

### LINKED ISSUE #1248: Wrong app id and missing icon in the dash (GNOME on Wayland)
Since some version of GNOME (certainly from 3.34), a symbolic icon is needed for the top bar.
Otherwise, no icon appears:

fresco-gnome-no-symbolic

I may look into it next week.

fedelibre:
Fix #1248: wrong app id and missing icon in GNOME on
Wayland.

I found the solution [here](https://github.com/IJHack/QtPass/pull/468).
I started working on this just for fun but I don't know what I'm really doing :-)

Now the icon is appearing in the left dash on GNOME.
In the topbar the application name is correct, but there's no icon.. only a kind of stylised F.
Unless I miss something, this patch finally allows to find Frescobaldi in GNOME Software,  even though the icon is not found. See below screenshots:

fresco-top-icon

fresco-gnome-dash

### DIFF
--- a/frescobaldi_app/app.py
+++ b/frescobaldi_app/app.py
@@ -125,8 +125,9 @@ def instantiate():
         args.append("-platform") 
         args.append("windows:fontengine=freetype") 
     qApp = QApplication(args) 
-    QApplication.setApplicationName(appinfo.name)
+    QApplication.setApplicationName(appinfo.rdns_name)
     QApplication.setApplicationVersion(appinfo.version)
+    QApplication.setDesktopFileName(appinfo.rdns_name)
     QApplication.setOrganizationName(appinfo.name)
     QApplication.setOrganizationDomain(appinfo.domain)
     if sys.platform.startswith('darwin'):
--- a/frescobaldi_app/appinfo.py
+++ b/frescobaldi_app/appinfo.py
@@ -25,6 +25,7 @@ Information about the Frescobaldi application.
 
 # these variables are also used by the distutils setup
 name = "frescobaldi"
+rdns_name = "org.frescobaldi.Frescobaldi"
 version = "3.2"
 extension_api = "0.9.0"
 description = "LilyPond Music Editor"
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -27,7 +27,7 @@ import re
 import sys
 
 from PyQt5.QtCore import QSettings, QTimer, QUrl
-from PyQt5.QtGui import QTextCursor
+from PyQt5.QtGui import QTextCursor, QGuiApplication
 from PyQt5.QtWidgets import QApplication
 
 import appinfo          # Information about our application
@@ -36,6 +36,9 @@ import install          # Update QSettings structure etc. if needed
 import guistyle         # Setup GUI style
 import i18n.setup       # Setup language
 import remote           # IPC with other Frescobaldi instances
+
+# It's needed by Qt Wayland backend. Requires Qt>=5.7.0
+QGuiApplication.setDesktopFileName("org.frescobaldi.Frescobaldi.desktop")
 
 
 def parse_commandline():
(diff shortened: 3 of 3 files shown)


wbsoft:
Lgtm, although you can probably set the property on QApplication instead of QGuiApplication. Both are the same, QApplication inherits from QGuiApplication and is already imported.

fedelibre:
@wbsoft I noticed that these kind of properties were already set in app.py so I moved it there and made it more pythonic :-)

I've verified that it's now working fine both on X11 and on Wayland.
The only issue I see is the **missing icon in the top bar of GNOME shell**. If I find some time I'll try to investigate. If you don't see new commits in this PR in the next two days, you can merge it and we'll take care of the icon later (what we've already fixed is more important IMO).

fedelibre:
I've verified this branch in another computer and also with another user in the same laptop. It works fine in both cases, either running fresco from the repository or from the installation directory.
So my guess is that I have some stale files or other installations which are interfering (I did several local installations in my home directory while working on the flatpak for Frescobaldi).

For me this is Ok to be merged.

fresco-topbar-ok

fedelibre:
>  In the topbar the application name is correct, but there's no icon.. only a kind of stylised F.

Ok, this is embarassing. I forgot I played with "symbolic icon discovering" some time ago.

Schermata da 2020-02-05 00-08-03

fedelibre:
A drawback of this PR - or maybe #1249 - is that Frescobaldi will create and use a new configuration file `~/.config/org.frescobaldi.Frescobaldi.conf` without copying the contents of previous file `~/.config/frescobaldi.conf`. I had to do it manually.

fedelibre:
I think it would be desirable to keep `~/.config/frescobaldi/frescobaldi.conf`.
But if I set back the application name to `frescobaldi`, then in X11 the window manager will fail to find the desktop file.

This is what I get when application name is rdns_name:

```
$ xprop | grep WM_CLASS
WM_CLASS(STRING) = "frescobaldi", "org.frescobaldi.Frescobaldi"
```

and this is what I get when application name is name:

```
$ xprop | grep WM_CLASS
WM_CLASS(STRING) = "frescobaldi", "frescobaldi"
```

I wonder if there are other ways to change the WM_CLASS. Other than changing QApplication.setApplicationName.
Same problem is discussed for example [here](https://github.com/bitcoin/bitcoin/issues/10064#issuecomment-289019240).

I'm afraid this is going to be beyond my limited skills... let me know your thoughts.

wbsoft:
Settings really should be taken over. On windows the registry is used, no frescobaldi.conf is used there. Maybe there is a away to have QSettings just use the old application name etc. Will look into it.

We could change the settings location, install.py could then take care of that. But it's hairy, especially when users run multiple versions of Frescobaldi in their home directory.

wbsoft:
It is possible to construct QSettings with its own application and organisation name.
https://doc.qt.io/qt-5/qsettings.html
I will devise a way to do that throughout Frescobaldi.

PeterBjuhr:
@fedelibre what is the status of this?

fedelibre:
I'm stuck, because this patch is also changing the name of the configuration file (at least in Linux) and we don't want this. Ideally, I would like to use the reverse DNS name for the application name and the normal name for the configuration file. But I couldn't find an easy way to do it.

fedelibre:
After checking what others did, I think that the desktop file name should include the .desktop extension so I should probably change a line to:

    QApplication.setDesktopFileName('org.frescobaldi.Frescobaldi.desktop')

I don't have much time and energy to try to understand this issue.

fedelibre:
<Face palm> It seems it's easier than expected. Setting the desktop file seems to be enough, no need to change the application name, since Wayland finds the right app ID in the desktop file.

I'm going to open a new PR and eventually we'll close this one if the fix is confirmed.

fedelibre:
replaced by #1478

PR #1262, opened 2020-02-15
fix pinch gesture zoom behaviour

dliessi:
Previously, only very quick pinches changed the scale substantially,
because scaleFactor concerns only the last input on the touch device
and long (in time) pinches most likely are seen as a sequence of
short pinches.  Instead, totalScaleFactor concerns the scale change
from the beginning of the pinch.

I don't know if this is relevant to other OS besides macOS.
My Linux machine does not support pinch gestures and I cannot test on other machines today.
I opened a pull request to give a chance for testing on Linux and Windows.


wbsoft:
It looks good to me. I can't test either. When there are no problems I'll merge it, with thanks!

fedelibre:
I think my trackpad does not support  pinch gesture (while the software should support it).
So I cannot test it, sorry.

Here's my device:

```
Device:           CyPS/2 Cypress Trackpad
Kernel:           /dev/input/event5
Group:            10
Seat:             seat0, default
Size:             100x60mm
Capabilities:     pointer 
Tap-to-click:     disabled
Tap-and-drag:     enabled
Tap drag lock:    disabled
Left-handed:      disabled
Nat.scrolling:    disabled
Middle emulation: n/a
Calibration:      n/a
Scroll methods:   *two-finger edge 
Click methods:    none
Disable-w-typing: enabled
Accel profiles:   none
Rotation:         n/a
```

dliessi:
Reading [Qt's documentation on the pinch gesture](https://doc.qt.io/qt-5/qpinchgesture.html) and the [example code](https://doc.qt.io/qt-5/qtwidgets-gestures-imagegestures-example.html) (where they indeed use totalScaleFactor), I am convinced that my interpretation is correct.
Anyway, I'll wait for @19joho66's input regarding Windows and then merge it.

dliessi:
> Anyway, I'll wait for @19joho66's input regarding Windows and then merge it.

I'm inclined to merge this tomorrow even without further review.

19joho66:
Hmm, somehow I missed this one. Sorry.
I've tested this with the touchpad on Win10 and couldn't see much difference. I think the zooming is ok in both cases. But anyway, it doesn't seem to have any negative effect on windows.

dliessi:
@19joho66: thanks for testing!

I updated the ChangeLog as well.
I'll squash and merge now.

PR #1263, opened 2020-02-16
Fix Python selection on Mac

dliessi:
I prefer this to be reviewed, since it is not trivial and does not touch only Mac-specific parts.

To test the new behaviour, you can artificially make `lilypondinfo.toolcommand` return `None` for some combinations of tool command and LilyPond version: the corresponding LilyPond entries should disappear from the list of the tool command.
You can also test with `lilypondinfo.toolcommand` always returning `None`, to see that it does not crash.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,7 @@ Changes in 3.1.2 -- ?
     - end of line: cmd+right
   - fixed #1087, File names in document tabs are not shown correctly on Mac
   - fixed #1272, Global menu with no windows on Mac is not working
+  - fixed #1232, Error using convert-ly in Frescobaldi 3.1 Mac app
 
 
 Changes in 3.1.1 -- January 4th, 2020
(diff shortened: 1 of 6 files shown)


dliessi:
@wbsoft I'd like to merge this: is it ok for you?

dliessi:
I squashed the two commits changing `toolcommand` and added a line in ChangeLog.

@wbsoft: I would really like to push this to master.
If you don't have time, maybe @uliska can have a look at this and check that I haven't messed up anything else.

dliessi:
Apparently the forced push to this branch made the exchange with @wbsoft disappear...

(For the record, the exchange was about the previous version of my changes to `lilychooser.py`, which added `toolcommand` as an argument to `lilychooser.load`, breaking the signals involving it.)

dliessi:
> Apparently the forced push to this branch made the exchange with @wbsoft disappear...

Found it: https://github.com/frescobaldi/frescobaldi/commit/a0c925d58c6a91faf96162b67f4029cbfb532566#r37433383

PR #1268, opened 2020-03-09
move arthurbackend_print preference to printing group

dliessi:
While testing this change, I had to run Frescobaldi twice for the install module to do its job.
Maybe before merging we should find out why that's happening.
How does it behave on Linux and Windows?


dliessi:
I split the change in two commits: one moves the checkbox to the "Printing of Music" group only in the UI, the other moves the option `musicview/arthurbackend_print` to `printing/arthurbackend_print` in QSettings.
I pushed the former directly to master and the latter to this pull request.

dliessi:
@wbsoft did the install module work for you the first time?

PR #1289, opened 2020-04-23
Add the NEWS file (replacing ChangeLog)

fedelibre:
In order to keep the AppStream metainfo file up-to-date, we should find a way to copy the release notes from a simple text file to the metainfo XML file. After reading [this blog post by the AppStream author](https://blog.tenstral.net/2020/03/maintain-release-info-easily-in-metainfo-files.html/comment-page-1#comment-341381), I decided to make a quick try and send a PR to initiate a discussion.

NEWS is a GNU "standard" and it's used by several Free Software projects:
https://www.gnu.org/prep/standards/html_node/NEWS-File.html

Even if a spec doesn't exist, it has a simple enough structure that allows for parsing and scripting.

***

You may notice that I've changed also the number and text of the 3.1.2 release notes.
Current ChangeLog seems sometimes a kind of memo for developers, with "weird numbers", technical language, acronyms, unclear changes. Not much different from git history.
I believe that release notes should be written to communicate to users: list only changes relevant to the users, avoid technical language if possible, try to be concise.

Let me have your opinions and feedback.
Then, *if you agree with this approach*, I will:

- add the current changes for 3.2 to NEWS
- add some instructions on README-development.
- use NEWS in the manifest and macosx/ directory (but I will keep ChangeLog in git, of course)

EDIT: Fix issue #1296


wbsoft:
I think it's a perfect idea!

fedelibre:
>  I will include the NEWS file instead of ChangeLog in the next release for Mac.

I've added a new commit to make this change now

fedelibre:
@wbsoft Can this be squashed and merged?

wbsoft:
Yes!

fedelibre:
I've merged on the command line and pushed to master, as I thought it was cleaner than using Github buttons. But I see that  it "breaks" the PR. Sorry for this.

Commit 00e5d22cb0309d10f8657ac78a577909a897c8c3

Issue #1321, opened 2020-07-07
Multiple window layout presets


uliska:
I'm right not in need for an solution to work on very different computer environments, mainly:

* My standard Laptop with a regular FullHD resolution
* A set-up with three monitors, the central one being rather large (and 4K).

The latter may also be used quite differently, say in settings where Frescobaldi is my main task or in others where writing music is just secondary (e.g. when writing texts with music examples, or when developing for Frescobaldi.

So I'd be glad about an option to save different window layouts and switch between them. Right now, when switching workplaces, I have to completely reconfigure all the stuff manually each time.

--- 

Where is this information stored (storage location and code)? I see a section `[mainwindow]` in the configuration file, and a `state` setting in it. Is this where the layout information is stored?

---

If we should implement this we should think about making it possible to save *session*-related window layouts and others that are independent and rather reflect the hardware environment.

uliska:
I now realize that this isn't only relevant for multiple screens but should be available for general use: to change between different layouts for different tasks. For example I often have work to do where I don't really need to see more than one system in the music view (or where my "scores" only consist of single-staff examples. For some projects I need the manuscript view next to the music view. Sometimes I'm focused on Scheme infrastructure development, sometimes for note entry you'll want to have the Quick Insert and MIDI input/playback tools available etc.

So there could by a submenu in the `View` menu giving access to an arbitrary number of custom layouts.

How would one want to name that feature?

* Workspaces
* Workspace configurations
* Application layouts
* Window layouts
* Desktops
?

PR #1323, opened 2020-07-09
Allow nested docks

uliska:
The option of nesting dock areas is useful on large screens.

However, this would be good to be accompanied by a solution to #1321. When switching back to my significantly smaller laptop screen the layout is totally messed up.

This screenshot is not what I really want it to look like but demonstrates what is possible with this PR:

grafik

The editor still is the main widget of the window. There's nothing in the top dock area but the left and right columns are heavily nested. Note in the music/SVG view that combining docks in tabs is still possible.

### DIFF
--- a/frescobaldi_app/mainwindow.py
+++ b/frescobaldi_app/mainwindow.py
@@ -93,6 +93,13 @@ class MainWindow(QMainWindow):
         QMainWindow.__init__(self)
         self.setAttribute(Qt.WA_DeleteOnClose)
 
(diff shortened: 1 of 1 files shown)


uliska:
@wbsoft do you think this could have any unpleasant side effects?

I don't see any but don't want to simply merge this because if we should revert this at some later point it might leave undesirable changes in users' preference files.

wbsoft:
I don't see unpleasant side effects, great!

PR #1327, opened 2020-07-26
Let FilePathEdit choose not only directories

uliska:
This is a convenience function to set the file mode of the included `QFileDialog`.
FilePathEdits may not only be useful for choosing directories.

I chose to add the option to pass in the value as a simple string so the caller doesn't have to include `QFileDialog`


uliska:
As nobody has objected for a long time, and I need this, I'll merge.

Issue #888, opened 2017-02-14
Edit existing score wizard without creating new score


bachstudies:
It would be fantastic if, after creating a score with the wizard, you could edit the information in the wizard to, say, add another voice part to a staff, add or remove an instrument, change the language, paper size etc. without creating a brand new score. Either that, or have the ability to easily add a new voice part or instrument via a menu option without needing to enter the Lilypond code which might be daunting for a novice.

And finally, as a way of uncluttering the code window, perhaps there could be a view option for simplified code so that each note or lyric entry was a textbox with the usual "enter notes/lyrics here" text.In other words, code that was not necessary for the user to see in order to produce a score could be hidden.

PR #1343, opened 2020-11-28
Improvements to the Score Wizard

bmjcode:
Hi there! I've made some improvements to the Score Wizard, including the ability to create a new score based on an existing document. I also did some work on the part types, adding several new instruments and doing some clean-up on existing ones.

### Creating scores from existing documents
I added a new command under File -> New -> From Current Document... that pre-populates the Score Wizard's fields based on an existing score. At the moment it handles headers and score properties well, and has partial support for recognizing existing parts -- at least those named following the Score Wizard's own convention. This is a first step towards using the Wizard to edit an existing score in-place, as requested in issue #888.

A few caveats:
* I have not yet implemented support for identifying parts nested in containers (but hope to soon if I have the time).
* I have not yet tested how this handles headers containing markup expressions.
* I have not attempted to preserve non-musical score properties (input language, layout/typography settings, etc.). Would it be useful to copy these from an current score? Or would it be better to leave those to the user's individual preference, as when generating a new score from scratch?

### Changes to instrument/part types
I've also added a few new instrument types to the Score Wizard that I think will be especially helpful for writing rock and popular music. Specifically, these are:
* Plucked strings
  * Acoustic guitar
  * Electric guitar
* Keyboards
  * Electric piano
  * Synth lead
  * Synth pad
  * Synth bass
  * Synth strings
  * Synth brass

The new guitar types are based on the existing classical guitar and jazz guitar, respectively, but with instrument names that don't tie them to a particular style or genre. They also default to different MIDI sounds, for example steel-string rather than nylon-string for the acoustic guitar. The electric piano is similarly a cosmetic rename of the existing piano type. Listing it separately allows distinguishing that instrument from the acoustic variety in styles (or individual scores) where both are used.

The synth types allow omitting the left- or right-hand parts by setting the corresponding number of voices to zero, to avoid generating extraneous staffs when writing monophonic lines. "Synth lead" and "synth bass" respectively default to a single treble- or bass-clef staff. The other synth types default to two staves, as these are intended for writing polyphonic parts.

Finally, I renamed the plucked string instrument "Bass" to "Acoustic bass" (abbreviated "A.Bs."), to differentiate it from the bass voice or other stringed bass instruments.

### DIFF
--- a/frescobaldi_app/mainwindow.py
+++ b/frescobaldi_app/mainwindow.py
@@ -1148,7 +1148,7 @@ class MainWindow(QMainWindow):
         t.addAction(ac.edit_undo)
         t.addAction(ac.edit_redo)
         t.addSeparator()
-        t.addAction(scorewiz.ScoreWizard.instance(self).actionCollection.scorewizNew)
+        t.addAction(scorewiz.ScoreWizard.instance(self).actionCollection.scorewiz)
         t.addAction(engrave.engraver(self).actionCollection.engrave_runner)
         w = t.widgetForAction(engrave.engraver(self).actionCollection.engrave_runner)
         w.addAction(engrave.engraver(self).actionCollection.engrave_publish)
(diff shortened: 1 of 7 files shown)


wbsoft:
Looks nice, thanks!!

PR #1350, opened 2021-01-06
fixed typing mistakes

freddii:


wbsoft:
Merged, thank you! Tip: it is not needed to modify pot or po files, those are automatically updated from the source files. But thanks to the typos you found,I fixed the remaining ones in the source files.

PR #1351, opened 2021-01-07
fixed misspelled word

freddii:


wbsoft:
Please don't fix the misspellings in the pot files, but fix them in the source code. I fixed it in 2cf7dc8df3d896634428f5511594a25f8517208b, so this PR is not needed anymore. Also pot and po files are auto-updated. But thanks!!

PR #1382, opened 2021-07-24
Fixes #1379: Add explicit decoding of arguments.

### LINKED ISSUE #1379: Launching Frescobaldi from File Explorer with multi-character path
When launching a LilyPond file from a Windows File Explorer, if the path to the file has multi-characters in it, Frescobaldi will open with two blank tabs.

Steps to recreate:

1. On a Windows 10 machine, create a folder structure with a multi-character folder in the path, like this:
`c:\documents\CHOIR_2021\勉勵\`

2. In that folder place a LilyPond file.  It can be anything.

3. Make sure Frescobaldi is associated withe LilyPond files.

4. From the Windows Explorer, double-click the LilyPond file.

5. Frecsobaldi will open with two blank tabs. 
frescobaldi-two-tabs

seraku24:
While this is addressing a Windows-specific issue, the added logic should work on all platforms. As such, I have omitted an OS check in the code. Unfortunately, I am unable to test on Mac or Linux, so we would need to confirm this does not introduce any new problems before merging the PR.

The underlying issue is that `QApplication` unintentionally mangles the argument list that it gets passed due to differences in string handling. There was logic added to use `os.fsencode` but no corresponding call to `os.fsdecode` was present to reconstitute the arguments.

### DIFF
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -86,6 +86,9 @@ def parse_commandline():
 
     args = QApplication.arguments()
 
+    # Decode arguments to properly handle Unicode
+    args = [os.fsdecode(bytes(arg, 'mbcs')) for arg in args]
+
     # Strip interpreter name and its command line options on Windows
     if os.name == 'nt':
         while args:
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -86,8 +86,9 @@ def parse_commandline():
 
     args = QApplication.arguments()
 
-    # Decode arguments to properly handle Unicode
-    args = [os.fsdecode(bytes(arg, 'mbcs')) for arg in args]
+    # Decode arguments to properly handle Unicode on Windows
+    if os.name == 'nt':
+        args = [os.fsdecode(bytes(arg, 'mbcs')) for arg in args]
 
     # Strip interpreter name and its command line options on Windows
     if os.name == 'nt':


VladimirSlavik:
Hi @seraku24 ! Thank you for trying to fix this, I'm one of the users affected by #1379. The `mbcs` encoding is a [Windows-only](https://docs.python.org/3/library/codecs.html#module-encodings.mbcs) thing, so the patch needs to detect the platform somehow, or try to see if `encodings.mbcs` exists. I tried this patch on Linux and got a traceback, as expected.

On a side note, I think rewriting `args` is bad form, but that's better left to the maintainers. As an user I'll take the fix and be happy regardless :-)

```
$ PYTHONPATH=$PYTHONPATH:/home/vslavik/Development/qpageview frescobaldi/frescobaldi komunitni-noty/hymnus-svatého-kříže/hymnus\ svatého\ kříže.ly 
  File "frescobaldi/frescobaldi", line 40, in <module>
    main.main()                     # Parse command line, create windows etc
  File "/home/vslavik/Development/frescobaldi/frescobaldi_app/main.py", line 154, in main
    args = parse_commandline()
  File "/home/vslavik/Development/frescobaldi/frescobaldi_app/main.py", line 90, in parse_commandline
    args = [os.fsdecode(bytes(arg, 'mbcs')) for arg in args]
  File "/home/vslavik/Development/frescobaldi/frescobaldi_app/main.py", line 90, in <listcomp>
    args = [os.fsdecode(bytes(arg, 'mbcs')) for arg in args]
LookupError: unknown encoding: mbcs
```
(The same command did succeed without the patch.)

seraku24:
Thanks, @VladimirSlavik, for the confirmation that a platform check is required. I have scoped the fix for `os.name == 'nt'`.

wbsoft:
Thanks!

PR #1390, opened 2021-09-05
Fix engraving of unsaved files using relative pathnames without \include

lfm83:
Unconditionally add directory of current LilyPond file to LilyPond
include directory list (-I ...) if engraving an unsaved file.
This enables use of relative paths in commands like \markup\epsfile ...
even if no actual \includes are being used in the .ly file.

### DIFF
--- a/frescobaldi_app/documentinfo.py
+++ b/frescobaldi_app/documentinfo.py
@@ -218,7 +218,7 @@ class DocumentInfo(plugin.DocumentPlugin):
             scratch = scratchdir.scratchdir(self.document())
             if create:
                 scratch.saveDocument()
-            if filename and self.lydocinfo().include_args():
+            if filename:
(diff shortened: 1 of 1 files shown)


PeterBjuhr (commented):
Looks ok to me. (But that condition has been there for along time so it's a little bit scary to remove)

dliessi:
`self.lydocinfo().include_args()` being true means that the document contains `\include` commands. Unconditionally adding the current directory to the include path is the default behavior of the `lilypond` command: indeed it makes no difference if the file is saved (and `lilypond` is run on the original file) and results in the behavior expected by the user if the file is not saved (and `lilypond` is run on a temporary file outside of the current directory).

dliessi:
I'm sorry, I didn't check the commit messages, which would have needed a bit of editing. I'll be more careful next time.

PR #1397, opened 2021-10-08
Fix app for MacOS with Homebrew

stweil:


frescobaldi_app/job/lilypond.py
```diff
@@ -173,8 +173,18 @@ def configure_command(self):
         # just before starting the process.
         if sys.platform.startswith('darwin'):
             import macosx
-            if macosx.inside_app_bundle() and self.lilypond_info.frommacports():
-                self.environment['PATH'] = self.lilypond_info.bindir()
+            if macosx.inside_app_bundle():
+                bindir = self.lilypond_info.bindir()
+                if self.lilypond_info.frommacports():
+                    self.environment['PATH'] = bindir
+                else:
```
dliessi, line 180: MacPorts and Homebrew are not the only two possibilities: there is also the app bundle we provide. Please add a `fromhomebrew` method to `lilypondinfo.py` and use it here for an `elif`.

frescobaldi_app/job/lilypond.py
```diff
@@ -173,8 +173,18 @@ def configure_command(self):
...
+                    # Add PATH for Homebrew.
+                    import re
+                    regexp = re.compile('(.*)/Cellar/.*')
+                    prefix = regexp.match(bindir)
+                    if prefix:
+                        bindir = prefix[1] + '/bin'
```
dliessi, line 186: Even better, can you directly fix `lilypondinfo.bindir` (using `fromhomebrew` and these lines therein) and here simply do
```
if macosx.inside_app_bundle() and (self.lilypond_info.frommacports() or self.lilypond_info.fromhomebrew()):
                self.environment['PATH'] = self.lilypond_info.bindir()
```
?

dliessi (commented):
Thanks for your work. Can you please consider the changes I propose? I would do those myself as follow-up work, but I do not use Homebrew and could not fully test them.

frescobaldi_app/job/lilypond.py
stweil: I had initially tried to get Frescobaldi working by adding `/usr/local/bin/lilypond` in the settings. That's a symbolic link to `../Cellar/lilypond/2.22.1/bin/lilypond`, and Frescobaldi used `/usr/local/Cellar/lilypond/2.22.1/bin` for `self.lilypond_info.bindir()` which is not the right path to find `gs`. So that case must be handled correctly as well.

My code should work fine with the app bundle for lilypond, too, and fix also manual installations which often install `lilypond` and `gs` in `/usr/local/bin`, so it is not restricted to Homebrew.

stweil: That's a good idea and might work, but then we have to make sure that symbolic links in the `lilypond` path are not resolved, because the resolved path is no longer valid for finding `gs`.

frescobaldi_app/job/lilypond.py
stweil: Maybe it would be better to remove `frommacports` instead of adding a `fromhomebrew`. Adding the path from `lilypond` as a search path for other executables like `gs` seams to be a general solution which is not specific to MacPorts or Homebrew.

jsfelix:
It works here with frescobaldi and homebrew lilypond. I am the author of the lilypond formula for homebrew and I thought I did something wrong!

Thanks!

dliessi:
@stweil Thanks for your contribution!

mykhaylo-:
Hi guys,
I'm about to extensively use Frescobaldi for my work and I really need to get this fix ready. I'm happy that it is merged to master. When do you think you'll be able to make a new release with this fix? 
Thanks, Mykhailo

matatata:
Thanks for the fix and I'm looking forward a release as well, as I was not able to build the current master myself.

@mykhaylo- In the meantime I was able to get my frescobaldi working fine with a homebrew installation on a MacBookAir M1 by taking the modified files from the master branch and putting them into the Frescobaldi.app version 3.1.3 application bundle:

take 
https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi_app/lilypondinfo.py and
https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi_app/job/lilypond.py

and replace the existing files in
/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app
and /Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app respectively

That works for me.

Best regards

mykhaylo-:
@matatata your fix worked for me. thank you so much

fer-gil:
Hello, I'm running on macOs 12.2.1 Monterey, installed frescobaldi and lilypond from brew, patched the files but I still get the same error:  `g_spawn_sync ha fallado (0): gs: Falló al ejecutar el proceso hijo «gs»`

PR #1402, opened 2021-11-10
Fix more errors under Python 3.10

dexgs:
This commit addresses 2 more instances of an implicit cast to int which
causes an error under python 3.10.

The features that work properly after these changes are:
- Engraving a score
- Splitting the source view


dliessi:
Thanks for the PR.
One of the changes was not actually needed: I changed your commit and pushed it to master as c0d68b4e1a6c8f6ed4e4f5d0147dc0d83f5c889c.

PR #1439, opened 2022-05-30
Remove NEWS

fedelibre:
NEWS is unmaintained and it's causing more problems than
solutions.  I'd rather keep the maintenance of the metainfo
file simple and use one line release tags with no
detailed descriptions.

Partially reverts 00e5d22cb0309d10f8657ac78a577909a897c8c3

PS
This will make Frescobaldi flatpak releases a bit easier. In the worst case scenario I'll have to add just a single line in the metainfo file, if Wilbert forgot it :-)


fedelibre:
@wbsoft It's an easy PR, can you please review it? :pray:

PR #1473, opened 2022-11-22
Add option to format on save

PeterBjuhr:


PeterBjuhr:
Yes, it seems I have. I was quite active before. But admittingly that is a couple of years back.

PeterBjuhr:
Great, I probably need an update!

fedelibre:
I've tested your branch and it works when I save the document.
As I have "Save document if possible" under *Running LilyPond* preferences, I would have expected that the format would have been performed also after a lilypond engrave, but it didn't. Did you exclude it on purpose or it's an oversight?

Finally - a comment not specific of this PR, but it might be the chance to discuss it - I think it might not be evident to the user what format is. Under *Tools>Code formatting* there are Re-indent, Format and Remove trailing whitespaces. I think that Format is Reindent+Remove trailing whitespaces. How can this be made more explicit? Maybe a tooltip? This affects also the preferences modified in this PR. Some ideas:

- Use an order with a rationale, for example from the less invasive to the most invasive. For Code formatting: Remove whitespaces, Reindent, Format. For Saving options: Strip trailing whitespaces, Format document, Keep backup copy, etc.
- Saving options: add a tooltip for Format document, e.g. "Strip trailing whitespaces and re-indent".
- Code formatting: what about adding a tooltip, at least for Format? (the other two are quite obvious)

PeterBjuhr:
It was an oversight. But looking at the code it seems Strip trailing whitespaces is also done only on document save.

If we add it to save after engrave how will that affect auto-engrave? I don't think it's a good to format on auto-save.

PeterBjuhr:
I agree that is not clear exactly what format does, but I think the concept is common enough for editors. 

We can look closer into this in a future PR.

PeterBjuhr:
@fedelibre I've added a commit for save on run. It didn't seem to affect the auto-engrave. So it seems to work fine.

If you have the time please check it aswell.

fedelibre:
Yes, it works as expected. Thanks

PR #1475, opened 2022-11-26
Enable Python warnings

Jean-Abou-Samra:
Most users won't see them due to Frescobaldi's GUI nature, but Frescobaldi contributors now will. These warnings would have provided an advance notice for things like the float conversion issues that plagued Frescobaldi in Python 3.10.

### DIFF
--- a/frescobaldi
+++ b/frescobaldi
@@ -28,6 +28,12 @@ frescobaldi_app contains the rest of the application.
 
 import sys
 
+# Enable all Python warnings, unless overridden (with
+# "python -W<something> frescobaldi").
+if not sys.warnoptions:
+    import warnings
+    warnings.simplefilter("default")
+
 from frescobaldi_app import toplevel
 toplevel.install()              # Add the path to frescobaldi_app to sys.path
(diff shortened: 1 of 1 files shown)


PeterBjuhr:
After some reading it seems this is a little dangerous. If I understand it correctly there is a growing support for the `s` argument (Coreutils>=8.30) but if it's not supported an error will occur.

Could you settle my concern about this?

PeterBjuhr:
Great, sounds like a safer approach!

frescobaldi
```diff
@@ -26,6 +26,10 @@ This script should be placed anywhere in the path, and the Python module
 frescobaldi_app contains the rest of the application.
 """
 
+# Enable all Python warnings
+import warnings
+warnings.simplefilter("default")
```
PeterBjuhr, line 31: suggestion: How about?
```
if not sys.warnoptions:
    import warnings
    warnings.simplefilter("default")
```
    
Found here: https://docs.python.org/3/library/warnings.html

PR #1476, opened 2022-11-26
Fix invalid escape sequence

Jean-Abou-Samra:
Fixes the warning

home/jean/repos/frescobaldi/frescobaldi_app/open_file_at_cursor.py:36: DeprecationWarning: invalid escape sequence '\s'
  incl_regex = re.compile('(\\\\include\s*\")(.*)(\")')

### DIFF
--- a/frescobaldi_app/open_file_at_cursor.py
+++ b/frescobaldi_app/open_file_at_cursor.py
@@ -33,7 +33,7 @@ import documentinfo
 import browseriface
 
 # regular expression for finding \include expressions
-incl_regex = re.compile('(\\\\include\s*\")(.*)(\")')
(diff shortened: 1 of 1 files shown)


frescobaldi_app/open_file_at_cursor.py
```diff
@@ -33,7 +33,7 @@
 import browseriface
 
 # regular expression for finding \include expressions
-incl_regex = re.compile('(\\\\include\s*\")(.*)(\")')
+incl_regex = re.compile(r'(\\include\s*")(.*)(")')
```
PeterBjuhr, line 36: suggestion: When you are changing the regex you might aswell change it to `r'(\\include\s*")([^"]*)(")'`. Which is more precise.

PR #1478, opened 2022-11-29
Specify the desktop file name

### LINKED ISSUE #1248: Wrong app id and missing icon in the dash (GNOME on Wayland)

fedelibre:
so Wayland (on Linux) will find the right application ID and display the icons correctly.

Fixes #1248.

<hr>

This replaces #1258 where I tried to also change the application name in order to fix also X11; but it broke the configuration file name.
Setting the desktop file name only at least fixes the problem in Wayland and doesn't change the configuration file name.

Wayland is [the default in most used Linux distros](https://en.wikipedia.org/wiki/Wayland_(display_server_protocol)#Adoption).


PeterBjuhr (approved):

dliessi:
@fedelibre before we merge this, can you please define the desktop file name in `appinfo.py` and use that variable in `app.py`? Thanks.

fedelibre:
> @fedelibre before we merge this, can you please define the desktop file name in `appinfo.py` and use that variable in `app.py`?

done

PR #1479, opened 2022-11-29
Deal with imp deprecation

### LINKED ISSUE #1457: Deprecation warning for imp module
Running `python -Wd frescobaldi`, I get

```
/home/jean/repos/frescobaldi/frescobaldi_app/portmidi/__init__.py:285: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
  import imp
```

Jean-Abou-Samra:
imp is planned to be removed in Python 3.12.

Fixes #1457

### DIFF
--- a/frescobaldi_app/portmidi/__init__.py
+++ b/frescobaldi_app/portmidi/__init__.py
@@ -282,11 +282,19 @@ def _load_module(name):
     Raises ImportError when the module can't be found.
 
     """
-    import imp
-    path = None
+    from importlib.machinery import PathFinder
+    from importlib.util import module_from_spec
+
+    spec = None
     for n in name.split('.'):
-        file_handle, path, desc = imp.find_module(n, path and [path])
-    return imp.load_module(n, file_handle, path, desc)
+        if spec is not None:
+            path = spec.submodule_search_locations
+        else:
+            path = None
+        spec = PathFinder.find_spec(n, path)
+        if spec is None:
+            raise ImportError
+    return module_from_spec(spec)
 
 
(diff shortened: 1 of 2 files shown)


dliessi:
I cannot test the change on Fedora since PyPortMidi in both `python3-portmidi` and `python3-pygame` lacks `Initialize()` (and I cannot test the Windows packaging since I don't have a Windows machine).
I believe there is a plan to upgrade `portmidi` and `python3-portmidi` to a recent version in Fedora, but even before that happens, MIDI is still working in Frescobaldi via `ctypes` (not affected by your change).

The change in `frescobaldi_app/portmidi/__init__.py` seems correct.
Do I understand correctly that you coded it the way you did instead of using `importlib.import_module` because the latter would import also the parent modules?

Do we have an idea of the state of PyPortMidi on various Linux distributions?
On Mac via MacPorts (and the prebuilt application bundle if I manage to resurrect it) the `ctypes` option is used.
We should probably advise packagers in the release notes to patch https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi_app/portmidi/__init__.py#L86 bringing `ctypes` in first position if the available PyPortMidi is incompatible with `importlib`.

dliessi:
> I am also using Fedora. For testing, I installed `python3-pygame` without `python3-portmidi`,

I must have done something wrong because now using pygame works for me too.

> Checking the state of portmidi in distros sounds like a good idea, but I would recommend doing it in a new issue;

I agree, I didn't mean to block this PR on that question.
Thank you!

PR #1480, opened 2022-11-29
Avoid <event> == <QKeySequence member> comparisons

### LINKED ISSUE #1453: 3.2 fails to launch on Python 3.11
File "/usr/lib/python3.11/site-packages/frescobaldi_app/view.py", line 107, in event
    if ev.type() == QEvent.KeyPress:
       ^^^^^^^
TypeError: a member of enum 'StandardKey' is expected not 'QEvent'

Fedora 37/38, Python 3.11

jeanas:
Under some versions of PyQt, these seem to "work", returning True if the event is a key event and matches the key sequence, but with other versions of PyQt, they cause crashes (#1453).

Fixes #1453

### DIFF
--- a/frescobaldi_app/view.py
+++ b/frescobaldi_app/view.py
@@ -103,7 +103,7 @@ class View(QPlainTextEdit):
             return False
 
         if ev.type() == QEvent.KeyPress:
-            if ev.type() == QEvent.KeyPress and ev.matches(QKeySequence.InsertLineSeparator):
+            if ev.matches(QKeySequence.InsertLineSeparator):
                 return False
 
             cursor = self.textCursor()


fedelibre:
I've tested your branch and I confirm it works on Fedora 37. Finally no crash.
I can't help with the code review.

madrisan:
I also tested this patch and everything works for me: startup, the find function (Control-F), and Control-Z too :)
When I type Control-Enter, the line is reformatted by removing the heading spaces. I do not know if this was the previous behaviour.
Thanks a lot!

fedelibre:
> When I type Control-Enter, the line is reformatted by removing the heading spaces. I do not know if this was the previous behaviour.

Do you mean the *trailing* spaces maybe? (as opposite to leading spaces, which should not be removed)

Recently a new option to reformat on save has been added.

PeterBjuhr:
> Recently a new option to reformat on save has been added.

Yes, but it should be off by default so you need to turn it on to use it. The setting for trailing spaces has been there for a while.

madrisan:
> Do you mean the _trailing_ spaces maybe? (as opposite to leading spaces, which should not be removed)

No, the leading ones. The text is moved to the first column of the line where the cursor is located.

fedelibre:
Can you paste a snippet so we can try to reproduce this bug?
Also, you wrote that you pressed Ctrl+Enter: what is doing this shortcut?

madrisan:
In fact most of the time I press Ctrl+Enter the IDE does nothing.
The leading spaces are just removed if an empty line exist *before* the line where the cursor is located.
```
      \once\shape #'((0 . 0) (0 . 1) (0 . 1) (1 . 0)) PhrasingSlur
      \ctuplet e'32\sf\( dis\)

      \omit TupletNumber

      \once\shape #'((0 . 0) (0 . 0) (0 . -1) (1.5 . 0)) PhrasingSlur
      \stemUp \ctuplet fis,32\sf_\( eis\)
```
So in the previous snippet, this is the case for the line `\omit TupletNumber`.
An empty line *after* the current line is irrelevant.

fedelibre:
It's still not clear to me. Does it happen on save (Ctrl+S)? Which saving preferences you have?

Maybe you are just formatting the input (which includes re-indenting)?

madrisan:
> Does it happen on save (Ctrl+S)? 

Nope. While editing a file.

>  Which saving preferences you have?

Should be the default ones. I just modified the UI colors.

> Maybe you are just formatting the input (which includes re-indenting)?

I can't understand this part. I just put the cursor on a line and press Ctrl+Enter

dliessi:
I can reproduce it.
It actually does not delete the leading space, it rather aligns the line to the current level of indentation: try it on
```
{
c
}
```
with the cursor on line 2.

madrisan:
> I can reproduce it.

Good. I think I'm using an unusual indentation scheme [(example)](https://github.com/madrisan/open-scores/blob/main/johann-sebastian-bach/chromatische-fantasie-und-fugue-BWV_903/parts/bach-chromatische-fantasie-und-fugue-2-fuga.ly#L10),  which I find useful to better organize the musical material. This may perhaps play a role.

dliessi:
Well, actually the behavior @madrisan is seeing seems to be independent of this PR: I can reproduce it with current master.

fedelibre:
I don't want to hijack this PR. Please open a new issue if you think there's a bug.

PS
I don't think it's a bug. It seems that Ctrl+Enter simply indents the current line to what you defined in your preferences. If you want a different indentation just change it in Preferences>Editor preferences.
Is Ctrl+Enter shortcut documented anywhere?

dliessi:
> I don't think it's a bug.

I agree.

> Is Ctrl+Enter shortcut documented anywhere?

I cannot find it.

madrisan:
I just start writing about Ctrl-Enter because Jean-Abou-Samra mentioned it [here](https://github.com/frescobaldi/frescobaldi/pull/1480#issuecomment-1330717729). But yes, it's unrelated to the issue fixed by this PR.

dliessi:
Moving back to this PR: on my installation (Fedora 36) current master seems to behave as declared in the comments and this PR changes that behaviour.
I would prefer to separate the correct handling of key presses and shortcuts and the possible change of behaviour.
That would mean finding a way to ignore `QKeySequence.InsertLineSeparator`, `QKeySequence.Undo` and `QKeySequence.Redo` in `view.py` and `QKeySequence.InsertLineSeparator` in `editinplace.py`.

Regarding the undo/redo actions, I have no idea what changes between handling them via `QPlainTextEdit` and via the menu, but @wbsoft added them to the list of ignored events with the comment "good view behaviour" (2dd0c0873e72a083ab082f5a71f59e627b5bd434).
Maybe after nine years of updates to Qt this is now outdated, but I do not know if that is the case.

Regarding the "insert line separator": the current behaviour is indeed that pressing shift+enter in the view is ignored, while in the new behaviour newlines are produced.
However, the new lines have no line numbers (I guess because `QPlainTextEdit` interprets them as new lines within the same paragraph, speaking in word processor terms), but when saving the file there is no difference between them and new lines obtained simply with enter, and when the file is reopened they are shown as normal lines, each with its line number.
I guess I cannot see the difference because I'm on Linux (i.e. with Unix-style new line characters).

In the edit-in-place function the current behaviour is the same as in the view, while the new behaviour inserts the new lines as normal new lines (with line number).

dliessi:
I guess @Jean-Abou-Samra meant to test shift+enter, but did instead ctrl+enter.

fedelibre:
Jean, this is probably the most important PR for the next release. I guess you didn't find time for this yet...

firstnevyn:
Without this patch frescobaldi launches.. then immediately crashes for me debian sid please consider working at all > working optimally

fedelibre:
Can we merge this PR and add a new issue explaining the problems still to be solved? It's not clear to me what this might break...

I'm not sure how many distros have already python3.11. For sure Fedora 37, but probably others will update in the coming months.

frescobaldi_app/view.py
```diff
@@ -94,17 +94,18 @@ def event(self, ev):
...
+              and (ev.matches(QKeySequence.Undo) or ev.matches(QKeySequence.Redo))):
+            # QPlainTextEdit accepts ShortcutOverride for Undo and Redo, which
+            # override the shortcut and causes a KeyPress to be broadcast
+            # instead of a Shortcut, preventing it from reaching our QActions
+            # and making it handled by the QPlainTextEdit instead.
             return False
-        # handle Tab and Backtab
+
         if ev.type() == QEvent.KeyPress:
+            if ev.type() == QEvent.KeyPress and ev.matches(QKeySequence.InsertLineSeparator):
```
dliessi, line 106: The previous line already tests for `ev.type() == QEvent.KeyPress`.

dliessi (commented):
The changes seem too work as expected. There is only a small fix needed to avoid an unnecessary test, then we can merge this. @jeanas I think that you can feel free to merge this yourself.

fedelibre:
Thank you Jean!

PR #1482, opened 2022-11-29
fix #1081: Ignore inactive code in structure

### LINKED ISSUE #1081: structure should ignore commands mentioned in commented lines
As you can see in the below screenshot, the Structure tool picks up also the commands present in a commented line.
I guess it should ignore commented commands as they are not active and sometimes just mentioned for the sake of remembering something, as in this example:

structure comments

In this case \bookpart should **not** appear in the Structure pane.

PeterBjuhr:
This PR can be seen as a start of a more ambitious project to rewrite the structure code with the help of ly as a code parser instead of relying on regex. It's by necessity a little hacky to adapt to the overall implementation.

### DIFF
--- a/frescobaldi_app/documentstructure.py
+++ b/frescobaldi_app/documentstructure.py
@@ -29,6 +29,8 @@ from PyQt5.QtCore import QSettings
 import app
 import plugin
 
+import lydocument
+import ly.document
 
 default_outline_patterns = [
 r"(?P<title>\\(score|book|bookpart))\b",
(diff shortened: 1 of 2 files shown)


fedelibre:
Now the files used as example in #1081 looks good, as the keywords in the last commented section are ignored.

LGTM, thanks

dliessi:
Indeed, as @PeterBjuhr noted in https://github.com/frescobaldi/frescobaldi/issues/1081#issuecomment-1331340329, with this change the keywords BEGIN, FIXME, HACK, XXX are not detected (in comments) and shown in the structure anymore.
I think that stripping comments before the pattern recognition is the wrong solution.

As a long term goal we should ideally port this feature to quickly during the transition, letting the user select (with checkboxes) which (active/uncommented) LilyPond keywords must appear in the structure and specify a list of keywords/regex patterns to recognize in comments.

For the time being, I'm revising the regex patterns and will open a PR soon.

dliessi:
81a1d192e7962550d6b153947498d5fed521be00 on top of this PR seems to work.
Of course, that commit is only a draft as I haven't extended the preferences and I didn't update the documentation strings.
@fedelibre and @PeterBjuhr: can you test it?

PeterBjuhr:
Unfortunately I will not have time to look deeper into this today. But to divide the regex's into two groups one which takes comments into account and one that doesn't seems necessary at this point, given that we can't totally ignore the comments.

dliessi:
I updated my branch based on top of this PR.
Now preferences are adapted to the new pair of lists and comments in the code are updated.
The relevant commit is 91dcae290192c9692b44bc71025b5e5ffef0c679.
@fedelibre @PeterBjuhr: please test it and if everything is fine we can merge this PR and push that commit on top (I see now that I can push to the branch of this PR, so I can first add my commit here and then merge).

PeterBjuhr:
Haven't tested it yet, only looked at the code. But it looks very good. There might be more we can do to avoid repeated code, but I think that would  be my only comment.

PeterBjuhr:
I have tested it now. I think everything looks good.

dliessi:
I refactored my commit to avoid some code duplication and pushed it to the branch of this PR.
@PeterBjuhr Can you please check it again in case I made some mistake?
If it is ok for you, feel free to merge it. Thanks.

fedelibre:
I've tested it and I think it works pretty good now. Thanks!

PR #1487, opened 2022-12-03
WIP: Add autocompletion for string tunings

### LINKED ISSUE #1298: String tunings autocompletion
python-ly doesn't have a list of the string tunings defined in LilyPond `ly/string-tunings-init.ly` file.
Adding the list to python-ly is easy.  But adding the completion in Frescobaldi is a bit more difficult.
(I've played with files analyzer.py and completiondata.py and I think I was close to make it work, but I did not understand how the classes in analyzer.py work)

This is an example of string tuning:

```
\new TabStaff \with {
  stringTunings = #guitar-dadgad-tuning
}
```

Here's the list to be added to python-ly words.py, so it doesn't get lost:

```
string_tunings = (
    'guitar-tuning',
    'guitar-seven-string-tuning',
    'guitar-drop-d-tuning',
    'guitar-drop-c-tuning',
    'guitar-open-g-tuning',
    'guitar-open-d-tuning',
    'guitar-dadgad-tuning',
    'guitar-lute-tuning',
    'guitar-asus4-tuning',
    'bass-tuning',
    'bass-four-string-tuning',
    'bass-drop-d-tuning',
    'bass-five-string-tuning',
    'bass-six-string-tuning',
    'mandolin-tuning',
    'banjo-open-g-tuning',
    'banjo-c-tuning',
    'banjo-modal-tuning',
    'banjo-open-d-tuning',
    'banjo-open-dm-tuning',
    'ukulele-tuning',
    'ukulele-d-tuning',
    'tenor-ukulele-tuning',
    'baritone-ukulele-tuning',
    'violin-tuning',
    'viola-tuning',
    'cello-tuning',
    'double-bass-tuning',
)
```

fedelibre:
Fixes #1298

This depends on [python-ly 149 PR](https://github.com/frescobaldi/python-ly/pull/149).

NOTE: In its current state it doesn't work.
I don't understand how tokens work in analyzer.py, but I guess it could be quite easy to fix for you guys :-)
Feel free to add commits on top of this PR.

Thanks!

### DIFF
--- a/frescobaldi_app/autocomplete/analyzer.py
+++ b/frescobaldi_app/autocomplete/analyzer.py
@@ -238,6 +238,16 @@ class Analyzer(object):
             self.column = self.lastpos
         return completiondata.midi_instruments
 
+    def string_tuning(self):
+        """Complete stringTunings = #... """
+        try:
+            i = self.tokens.index('stringTunings', -7, -1)
+        except ValueError:
+            return
+        if self.last != '#':
+           self.column = self.lastpos
+        return completiondata.string_tunings
+
     def font_name(self):
         """Complete #'font-name = #"..."""
         try:
--- a/frescobaldi_app/autocomplete/completiondata.py
+++ b/frescobaldi_app/autocomplete/completiondata.py
@@ -259,6 +259,8 @@ music_glyphs = listmodel.ListModel(ly.data.music_glyphs())
 
 midi_instruments = listmodel.ListModel(ly.words.midi_instruments)
 
+string_tunings = listmodel.ListModel(ly.words.string_tunings)
+
 language_names = listmodel.ListModel(sorted(ly.pitch.pitchInfo))
 
 def font_names():


PeterBjuhr (changes requested):
You have to add `string_tuning` to the tests section further down in the file for it to actually do the test (run the function).

PeterBjuhr:
I have added a commit for the above.

fedelibre:
> I have added a commit for the above.

Thanks, I see that the words are shown now in the dropdown autocomplete list. So it works.
I only wonder what these two numbers are actually doing:

    i = self.tokens.index('stringTunings', -7, -1)

I wrote this stuff two years ago, but I don't think I knew why I chose those values.

PeterBjuhr:
`list.index`

> The optional arguments start and end are interpreted as in the slice notation and are used to limit the search to a particular subsequence of the list. The returned index is computed relative to the beginning of the full sequence rather than the start argument.
https://docs.python.org/3/tutorial/datastructures.html

If think you can consider fine-tuning these values. :slightly_smiling_face:

PeterBjuhr:
When I was debugging this yesterday I had this printout of `self.tokens`:
`[<_token.Space '  ' at 0>, <lilypond.ContextProperty 'stringTunings' at 2>, <_token.Space ' ' at 15>, <lilypond.EqualSign '=' at 16>, <_token.Space ' ' at 17>, <lilypond.SchemeStart '#' at 18>, <scheme.Word 'gu' at 19>]`

What I was typing at that point was:
> stringTunings = #gu

PeterBjuhr:
What we have is a list with 7 elements. If you slice the list [-7,-1] you will cut out the last element. 
The important thing as I see it is to match with "stringTunings". 

In this case perhaps it's not that sensitive, but in other cases I guess it could be important to do the prefect slice.

PeterBjuhr:
I simply did `print(self.tokens)`.

fedelibre:
> I simply did `print(self.tokens)`.

Where? I did try before asking but I guess I placed it in the wrong place

PeterBjuhr:
> Where? I did try before asking but I guess I placed it in the wrong place

You can put it in the top before `try`. You have to run Frescobaldi from the terminal of course to see the output.

fedelibre:
Thanks, but I'm tired. I give up

PeterBjuhr:
> So the slice should be [-7,1]

Ah, I didn't see at first that you are ending with a positive index. That would probably not work (unless you skip the space in the beginning).

fedelibre:
Honestly, I don't have any understanding of what's going on. I thought that slicing was acting on the "tokens" I'm entering in Frescobaldi, but it has also effect on the list of words appearing in the autocomplete. Why? Obviously, I want to have all the string_tunings words available.

Also when using `[-7,-1]` typing `gu` will show all the `guitar-` tunings available. Typing `uk` once doesn't show anything, but if I delete and retype `uk` I see the ukulele tunings. This is 100% reproducible and I don't understand why it happens.

fedelibre:
Also I don't understand why changing some values as in:

```
-        if self.last != '#':
+        if self.last != '':
```

doesn't produce any difference in Frescobaldi. The autocomplete keeps working in the same way.

PeterBjuhr:
> it has also effect on the list of words appearing in the autocomplete. Why?

I don't see that it has any effect. What setting do you use?

> Also when using [-7,-1] typing gu will show all the guitar- tunings available. Typing uk once doesn't show anything, but if I delete and retype uk I see the ukulele tunings. This is 100% reproducible and I don't understand why it happens.

Actually I can't reproduce that. :thinking: 

> Also I don't understand why changing some values as in:

```
-        if self.last != '#':
+        if self.last != '':
```
> doesn't produce any difference in Frescobaldi. The autocomplete keeps working in the same way.

Yes, I can confirm that. `self.last` will never have the value '#'. And even if you put in something else that would match, `self.column` and `self.lastpos` are already the same (when I debug it).

fedelibre:
> > it has also effect on the list of words appearing in the autocomplete. Why?
> 
> I don't see that it has any effect. What setting do you use?
> 
> > Also when using [-7,-1] typing gu will show all the guitar- tunings available. Typing uk once doesn't show anything, but if I delete and retype uk I see the ukulele tunings. This is 100% reproducible and I don't understand why it happens.
> 
> Actually I can't reproduce that. thinking

I was wrong in describing the issue. Here's a simple way to reproduce it:

1. Type `gu`; the right list will appear
2. Delete `gu` with backspace until # and type `uk`; nothing will appear
3. Delete `uk` with backspace until # and type `uk` again; the second time it appears

You can swap the characters to be typed and you get the same behaviour: the second entry won't work at first, but only at the second try.

So the function seems to be triggered 100% only when typing from # onwards.
Is this that should be fine-tuned?

fedelibre:
However, I cannot always reproduce it.

I put some print() statements and I see why self.last is never #. I copied from the midi_instrument function above, which has a different syntax: it's surrounded by `""`, while string-tuning has only #keyword.

In this output:

```
[<_token.Space '      ' at 0>, <lilypond.ContextProperty 'stringTunings' at 6>, <_token.Space ' ' at 19>, <lilypond.EqualSign '=' at 20>, <_token.Space ' ' at 21>, <lilypond.SchemeStart '#' at 22>, <scheme.Word 'g' at 23>]
g
24
23
[<_token.Space '      ' at 0>, <lilypond.ContextProperty 'stringTunings' at 6>, <_token.Space ' ' at 19>, <lilypond.EqualSign '=' at 20>, <_token.Space ' ' at 21>, <lilypond.SchemeStart '#' at 22>, <scheme.Word 'gu' at 23>]
gu
25
23
[<_token.Space '      ' at 0>, <lilypond.ContextProperty 'stringTunings' at 6>, <_token.Space ' ' at 19>, <lilypond.EqualSign '=' at 20>, <_token.Space ' ' at 21>, <lilypond.SchemeStart '#' at 22>, <scheme.Word 'uku' at 23>]
uku
26
23
[<_token.Space '      ' at 0>, <lilypond.ContextProperty 'stringTunings' at 6>, <_token.Space ' ' at 19>, <lilypond.EqualSign '=' at 20>, <_token.Space ' ' at 21>, <lilypond.SchemeStart '#' at 22>, <scheme.Word 'g' at 23>]
g
24
23
[<_token.Space '      ' at 0>, <lilypond.ContextProperty 'stringTunings' at 6>, <_token.Space ' ' at 19>, <lilypond.EqualSign '=' at 20>, <_token.Space ' ' at 21>, <lilypond.SchemeStart '#' at 22>, <scheme.Word 'gu' at 23>]
gu
25
23
[<_token.Space '      ' at 0>, <lilypond.ContextProperty 'stringTunings' at 6>, <_token.Space ' ' at 19>, <lilypond.EqualSign '=' at 20>, <_token.Space ' ' at 21>, <lilypond.SchemeStart '#' at 22>, <scheme.Word 'uku' at 23>]
uku
26
23
```

there are:

- self.last = the characters I've entered
- self.lastpos is always 23
- self.column changes depending on how many characters I typed.

PeterBjuhr:
Yes, I can reproduce it now. I will check and see if some setting prevents. Otherwise it seems from my own study and from what you have written that the settings have little effect on the outcome.

PeterBjuhr:
I'm trying to get into this again. 

Regarding the strange behavior with `uke` the settings inside your code doesn't seem to affect this, so I suspect it depends on stuff outside in the analysis.

Some observations about the settings themselves:
1. To narrow it  done as much as possible I can write:
`i = self.tokens.index('stringTunings', -6, -5)`
2. I tried:
`if self.last != '"':` but since the quotation marks aren't needed I'm not sure it's a good idea.

fedelibre:
There's also this in analyzer.py:

```
    def string_tuning(self):
        """Complete stringTunings = #... """
        try:
            i = self.tokens.index('stringTunings', -7, -1)
        except ValueError:
            return
        if self.last != '#':
           self.column = self.lastpos
        return completiondata.string_tunings
```

Sometimes the list returned is not limited to the string_tunings list, but it includes Scheme keywords, probably triggered by the `#`. This sounds like a bug, as I think the list should be limited to string tunings in that case. But anyway the point is: maybe this strange behaviour might be caused by a too big list?

PS
I've added a commit to sort the string tunings.

fedelibre:
If I delete the string tuning I'm typing and also the #, then type again, it always work. The problem appears when I don't delete  the #. It's like another function is used instead of string_tuning.

PR #1498, opened 2023-01-02
Autocomplete: add unfoldRepeats

### LINKED ISSUE #1496: Implement autocomplete for \unfoldRepeats
It seems to me that the command \unfoldRepeats is never auto-completed. This should be fixed.

fedelibre:
Fixes #1496

### DIFF
--- a/frescobaldi_app/autocomplete/completiondata.py
+++ b/frescobaldi_app/autocomplete/completiondata.py
@@ -65,6 +65,7 @@ inputmodes = (
     'lyrics',
     'lyricmode {',
     'addlyrics {',
+    'notemode {',
 )
 
 # commands that only occur at the global file level
@@ -81,6 +82,8 @@ start_music = (
     'repeat',
     'alternative {',
     'relative',
+    'absolute',
+    'fixed',
     'transpose',
     'partCombine',
     'keepWithTag #\'',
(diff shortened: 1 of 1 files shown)


PeterBjuhr (approved):

fedelibre:
Yes, probably. But I was lazy and just wanted to pick the low-hanging fruit.

I may add some more in the coming days. It's a manual check with words.py of python-ly...

fedelibre:
IIUC, the top level is when I write a command which is not wrapped by anything.
Not all the music commands in [lilypond_music_commands](https://github.com/frescobaldi/python-ly/blob/master/ly/words.py#L64) can be written in the top level.

This works:

```
\version "2.24.0"

\absolute { c''4 c' }
```

while this doesn't:

```
\version "2.24.0"

\autoBeamOff c4 d8 e f16 g f e
```

I'm missing something?

Maybe lilypond_music_commands should be split in more lists in order to have a better use of these commands in completiondata.py? This would be a difficult task for me though.

fedelibre:
Ok, I've added a few more words from `ly.words.lilypond_music_commands`.

For the shortcuts, I think we should add a new function in completiondata.py which returns the list of `ly.words.lilypond_music_commands` when a shortcut is created at toplevel.
Maybe something similar to [this](https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi_app/autocomplete/completiondata.py#L228)? I'm not sure if I'm able to do it... I would not consider this for this PR.

fedelibre:
I'm happy with my humble PR. Ok to merge it?

PR #1510, opened 2023-02-04
further update on Japanese translation

JunTamura1959:


JunTamura1959:
The ja.po file was not up-to-date w.r.t. frescobaldi.pot.

PR #1515, opened 2023-02-08
French translation updates

jeanas:
From Olivier Miakinen.

### DIFF
--- a/i18n/userguide/fr.po
+++ b/i18n/userguide/fr.po
@@ -17,7 +17,7 @@ msgstr ""
 "Project-Id-Version: frescobaldi 1.9.0\n"
 "Report-Msgid-Bugs-To: info@frescobaldi.org\n"
 "POT-Creation-Date: 2023-02-07 14:05+0100\n"
-"PO-Revision-Date: 2023-02-04 01:14+0100\n"
+"PO-Revision-Date: 2023-02-14 20:25+0100\n"
 "Last-Translator: Raphaël Doursenaud <rdoursenaud@free.fr>\n"
 "Language-Team: French <kde-i18n-doc@kde.org>\n"
 "Language: fr\n"
@@ -908,7 +908,6 @@ msgid "Local Font Repository"
 msgstr "Dépôt de polices local"
 
 #: ../../frescobaldi_app/userguide/documentfonts_music.md:37
-#, fuzzy
 msgid ""
 "If a directory is made known as a local repository for music fonts (see "
 "{paths}) its content can easily be used as a source for font installation in "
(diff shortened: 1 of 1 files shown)


i18n/userguide/fr.po
```diff
@@ -16,8 +16,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: frescobaldi 1.9.0\n"
 "Report-Msgid-Bugs-To: info@frescobaldi.org\n"
-"POT-Creation-Date: 2023-02-07 14:05+0100\n"
-"PO-Revision-Date: 2023-01-09 21:16+0100\n"
+"POT-Creation-Date: 2023-02-02 23:39+0100\n"
```
fedelibre, line 19: It seems it's based on an older POT file?

i18n/userguide/fr.po
```diff
@@ -888,36 +886,45 @@ msgid ""
...
-msgstr "Aller à l’emplacement suivant"
+msgstr "Dépôt de polices local"
 
 #: ../../frescobaldi_app/userguide/documentfonts_music.md:37
 msgid ""
 "If a directory is made known as a local repository for music fonts (see "
 "{paths}) its content can easily be used as a source for font installation in "
 "arbitrary numbers of LilyPond instances. The easiest way to make use of this "
-"is to download or clone the font repository from GitHub ({notationfonts}) "
+"is to download or clone the font repository from Github ({notationfonts}) "
```
fedelibre, line 915: IIRC when I skimmed through the previous PR Github was changed to GitHub. Now it's changed back to Github. GitHub is correct, but we'd better fix also the English manual.

i18n/userguide/fr.po
```diff
@@ -770,40 +770,38 @@ msgstr ""
...
 #. NOTE: markdown formatting
 #: ../../frescobaldi_app/userguide/documentfonts_command.md:51
-#, fuzzy
 msgid ""
 "*Font stylesheets:* openLilyLib provides default stylesheets for each "
 "supported font, adjusting the visual appearance (e.g. line thickness of "
 "various score items) to match the characteristics of the given font. By "
-"default these are loaded automatically by the font command "
-"*\\useNotationFont*. However, in some cases it may be necessary to *not* use "
+"default these are loaded automatically by the font command *"
```
fedelibre: The changes in msgid strings of .po files don't make any sense, as they will get overwritten when the strings are extracted the next time. Translators should only work on msgstr strings: in fact PO editors usually prevent the editing of msgid strings. Anyway, I would just explain it to Olivier for next contributions; I'm not asking you to fix it in this PR.

fedelibre: Or maybe Github is showing me the diff of the entire PR even if I entered the first commit? If the answer is yes, ignore my previous comment.

fedelibre (approved):

PR #1520, opened 2023-02-13
correct jump to code from error log

mr-bronson:
This corrects the cursor position when jumping from an error message in the log to the referenced place in the code.

If column numbers ever did start with 0 in LilyPond, they no longer do, so changed that comment (line 117).

QTextCursor.setPosition() works on QChar (16-bit) offsets and is thus not compatible with supplementary plane characters.
QTextCursor.movePosition() with MoveOperation NextCharacter almost works, but jumps over composed characters as one, whereas LilyPond counts each codepoint.
So the strategy is counting characters via Python str, as with determining correct cursor position in status bar.

The handling of column number overrun (lines 156-158) may seem overcomplicated vs. just taking a minimum of the qchar_offset and the block length(), but QTextBlock.length() "includes all formatting characters, for example, newline", and I don't know if it's guaranteed to only and always be exactly one LF character.  This approach avoids needing that guarantee.

Realized that arbitrarily setting position of cursor can result in it being in the middle of a composed glyph, or, if the block changes during the offset calculation, in the middle of a supplementary plane codepoint.  Either way leads to undesirable behavior when input resumes, e.g. moving a combining character to a different base or splitting a surrogate pair.  That's what line 155 takes care of.  And yes, the order (next, then previous) matters, because we want the cursor just in front of the offending glyph, not after it.

This works on my end.  Tested with this input (test.ly)
```
\version "2.22.2"
\header {
) % <-- first of this line
  title = "Testing Error Jumps" % no error here; last of next line -->
  subtitle = "in Frescobaldi")
  source = "AAAAAAAAA" ) % ASCII range
  composer = "ÅÅÅÅÅÅÅÅÅÅÅ") % composed character
  style = "ÅÅÅÅÅÅÅÅÅ" ) % precomposed character
  meter = "𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴𐌴") % supplementary plane letter
  poet = "😊😊😊😊😊😊😊😊😊" ) % supplementary plane emoji
  copyright = "❤❤❤❤❤❤❤❤❤❤") % basic plane emoji
}
\score {
  << a >>
}% this is the end of the file.
```
If you try this, you'll notice there are still various display/formatting problems with the supplementary plane characters, but the cursor goes to the right place.

Also tested out-of-bounds behavior by temporary hard-coding of line and column numbers.

### DIFF
--- a/frescobaldi_app/logtool/errors.py
+++ b/frescobaldi_app/logtool/errors.py
@@ -38,7 +38,7 @@ import util
 
 
 # finds file references (filename:line:col:) in messages
-message_re = re.compile(br"^((.*?):(\d+)(?::(\d+))?)(?=:)", re.M)
+message_re = re.compile(br"^((.*?):([1-9]\d*)(?::([1-9]\d*))?)(?=:)", re.M)
 
 
 def errors(document):
@@ -95,7 +95,7 @@ class Errors(plugin.DocumentPlugin):
                 url = m.group(1).decode(enc)
                 filename = m.group(2).decode(enc)
                 filename = util.normpath(filename)
-                line, column = int(m.group(3)), int(m.group(4) or 1) - 1
+                line, column = int(m.group(3)), int(m.group(4) or 1)
                 self._refs[url] = Reference(filename, line, column)
 
     def cursor(self, url, load=False):
@@ -114,7 +114,7 @@ class Reference(object):
     def __init__(self, filename, line, column):
         """Creates the reference to filename, line and column.
 
-        lines start numbering with 1, columns with 0 (LilyPond convention).
+        By LilyPond convention, lines and columns are numbered starting with 1.
 
         If a document with the given filename is already loaded (or the filename
         refers to the scratchdir for a document) a QTextCursor is created immediately.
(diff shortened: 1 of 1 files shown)


mr-bronson:
Sorry, I don't know what I'm doing.  I didn't realize that would happen.  I have my revised suggestion in a new branch.

mr-bronson:
It doesn't let me Reopen.  Do you want me to start another PR, or is there a way you can reopen this and connect it to the correct branch?

mr-bronson:
I was trying to follow those instructions.  There is no "base branch drop-down menu" after clicking Edit.  It only lets me change the title.

> BTW, what is the purpose of checking the first value?

I guess I just was thinking that if the first operation failed, the second would be unnecessary and unwelcome.  I'm still very new to Qt and had it in my mind that changes could be happening to the widgets while this slot is being executed, but I'm getting the idea that is incorrect, and it's all synchronous.  Is that correct?  If it's all synchronous, then it should be impossible for either operation to fail.  And it would be impossible to land in the middle of a surrogate pair, as I was talking about above.  (It would still be very possible to land in the middle of a composed glyph, e.g. base character + diacritic, or emoji + modifier, or emojis joined by zero-width joiner, etc., which is why this pair of operations is needed.)  My revised suggestion simplifies some things.

> In Python, using `and` for its side effects like this is uncommon.

For future reference, what exactly do you mean?  Surely you don't mean the short-circuit evaluation, because I see that used all the time.

mr-bronson:
Okay, I renamed the new branch to master, and that let's me reopen.

> I mean writing an `and` expression and discarding its result entirely

Thanks.

PR #1524, opened 2023-02-15
Fix command line options --line and --column

mr-bronson:
### Description of work done

1) Fixed operation of command line options to jump to line and column position in the document (`frescobaldi_app/main.py`).
2) Updated command line documentation to reflect the fact that columns start with 1 (`frescobaldi.1` and `frescobaldi_app/main.py`).

3) _Manually updated relevant .pot file and .po files to reduce diff headaches._
4) _Because of the similarity of the problem addressed in [PR 1520](https://github.com/frescobaldi/frescobaldi/pull/1520), a new method has been created in `document.py` which is used by both `main.py` and `logtool/errors.py`._

~~3) Updated .pot files in `i18n` directory by running `make pot`.~~
~~4) Updated translations of the modified message (`i18n/frescobaldi/*.po`).~~
~~5) Noticed and confirmed that `i18n/userguide/ja.po` contained no translations and no file header; it is automatically generated.~~
~~6) Removed "ja" from `i18n/userguide/LINGUAS` in keeping with this message in `i18n/Makefile`:~~
~~> `# We want translation of the userguide to be optional.`~~
~~> `# That's why we have two sets of PO files and two lists as LINGUAS files.`~~
~~7) Replaced `i18n/userguide/ja.po` with a placeholder file describing why it is empty but why Makefile still needs it to exist.~~
~~8) Checked work in step 4 and successfully built .mo files (not git tracked) with `make`.~~

### Other comments

~~The code changed in step 1 above has obvious similarities to my suggested fix in [PR 1520](https://github.com/frescobaldi/frescobaldi/pull/1520) of the error log's jump to code feature.  But I wasn't sure it warranted abstracting it into a unified function.  Either way is fine with me.~~

I trimmed trailing spaces from 4 lines in frescobaldi.1 while I was in there.

The two languages that had blank translations for the message being changed, I left blank (ja and zh_HK).

~~I don't understand the purpose, if any, of the apparent junk comments beginning with `#~ ` at the end of most of the .po files.  Thirteen of these files still contain the old message being replaced:~~
~~`#~ msgid "Column to go to, starting at 0"`~~
~~Assuming these truly are junk, I'll leave them to be cleaned up by someone else.~~


mr-bronson:
> > But I wasn't sure it warranted abstracting it into a unified function.
> 
> Yes, please.

Where should such a function go?  My vote would be in `frescobaldi_app/document.py` as a new method of class `EditorDocument`.

> this is a case where two separate PRs ... would have been warranted

Sorry.  You're right.  It seemed like a minor change at the time, but I see your point.  I'll try to address that issue separately.

mr-bronson:
I've updated the initial comment.  But I still can't find a way to update the pull request to a different branch.  When it's open, it lets me change the branch I want to merge it into, but not the branch it's from.  When it's closed, it doesn't let me do a thing.

Last time, changing the new branch to the name of the old branch worked, but that doesn't seem to be working now, so I will have to try something else.

PR #1534, opened 2023-02-19
Tiny update in README.md

JunTamura1959:
Japanese has been added at the last of "Translated into:...". I'm not sure about the appropriate order of the languages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Features:
 - Experimental export to MusicXML
 - Modern user interface with configurable colors, fonts and keyboard shortcuts
 - Translated into: Dutch, English, French, German, Italian, Swedish, Czech,
-  Russian, Spanish, Galician, Turkish, Polish, Brazilian and Ukrainian.
+  Russian, Spanish, Galician, Turkish, Polish, Brazilian, Ukrainian, and Japanese.
 
 Music functions:
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ Features:
 - Experimental export to MusicXML
 - Modern user interface with configurable colors, fonts and keyboard shortcuts
 - Translated into: Dutch, English, French, German, Italian, Swedish, Czech,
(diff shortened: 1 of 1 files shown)


fedelibre:
Would you mind adding also chinese to the list?
Nitpick: the comma before `and` can be removed, so I would write:

> Brazilian, Ukrainian, Chinese and Japanese.

Here's the current list of languages:

```
$ ls i18n/frescobaldi
cs.po  es.po  gl.po  it.po~  LINGUAS  pl.po     ru.po  tr.po  zh_CN.po  zh_TW.po
de.po  fr.po  it.po  ja.po   nl.po    pt_BR.po  sv.po  uk.po  zh_HK.po
```

JunTamura1959:
Thanks for your comments.
> Would you mind adding also chinese to the list?
Yes, now it's -
- Translated into: Dutch, English, French, German, Italian, Swedish, Czech,
  Russian, Spanish, Galician, Turkish, Polish, Brazilian, Ukrainian,
  Traditional Chinese, Simplified Chinese and Japanese.
> Nitpick: the comma before "and" can be removed, so I would write:
Thanks and it's interesting. I agree that I do not find comma before "and" very often. On the other hand, an American classic book about English writing "The Elements of Style" in its first edition of 1918 (and as fare as I recall, also in the 1959 edition, which I used to have and misplaced) says:
----------
2. In a series of three or more terms with a single conjunction, use a comma after each term except the last.
Thus write,

red, white, and blue
gold, silver, or copper
He opened the letter, read it, and made a note of its contents.

This is also the usage of the Government Printing Office and of the Oxford University Press.

In the names of business firms the last comma is omitted, as,
Brown, Shipley & Co.
----------
https://www.gutenberg.org/ebooks/37134

As I'm not a native English speaker, I'm not sure which is considered standard nowadays. Also The Elements of Style has been revised many times to adopt the changes in English language, the latest edition might say something different.

JunTamura1959:
Thanks, Jean. I never realized that there is a dedicated Wikipedia article about it.

PR #1535, opened 2023-02-21
Improve selection behavior in list of snippets

mr-bronson:
This fixes a few problems with the snippets list.

1) Currently, if you have a snippet selected, then add a new one, both what was selected and the new one will be selected.  This isn't always obvious to a user, because they could be in a totally different place in the list.  If the user then deletes the snippet he just made, he'll also be deleting what else just happened to be selected before he made a new snippet.  This is fixed by clearing the selection and only selecting the new snippet.

2) Currently, the `updateFilter` method selects the last item in the list that has no `nameid` whenever the search box becomes empty.  This is because `"" == ""`, which is an insufficient reason to select a snippet from the search.  This is fixed by testing for empty search `text`.

3) Currently, the `QTreeView` list of snippets maintains previous selections while searching.  This leads to the odd behavior of snippets leaving the list and coming back onto the list already selected.  Or, it can mean that a snippet stays selected even when you've clearly typed something that doesn't match its `nameid`, simply because some previous state of the search box text did match a one or two letter `nameid`.  These things are fixed by clearing the selection state before the for loop in `updateFilter`.

4) Currently, the list of snippets ends up scrolled to relatively random places.  This is fixed by scrolling to the first item on the list when the search box makes no selection, or to the last (usually only one) snippet selected by the search.  This provides a much more intuitive and dependable behavior than the user wondering why sometimes it scrolls him down and sometimes it doesn't.  If he wants to select something manually, it's always going to be with the down arrow or page down instead of having to check whether there are also results hidden above the current top result.

Notes:

I'm not really sure why `QItemSelectionModel.Clear` is needed in `edit.py` to avoid multiple selection but isn't needed in `widget.py` to avoid multiple selection.  (It probably wouldn't hurt to add it, but I didn't because it works without it in `widget.py`.)  Whatever the reason, I've tested it and it works as expected for me.

### DIFF
--- a/frescobaldi_app/snippet/edit.py
+++ b/frescobaldi_app/snippet/edit.py
@@ -186,7 +186,7 @@ class Edit(QDialog):
             self.text.toPlainText(), self.titleEntry.text())
         # set snippet current in the editor that called us
         self.parent().treeView.selectionModel().setCurrentIndex(
-            index, QItemSelectionModel.SelectCurrent | QItemSelectionModel.Rows)
+            index, QItemSelectionModel.Clear | QItemSelectionModel.SelectCurrent | QItemSelectionModel.Rows)
         #remove the shortcuts conflicts
         self.actionManager().removeShortcuts(self.shortcuts())
         self.parent().treeView.update()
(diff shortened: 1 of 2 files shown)


mr-bronson:
https://user-images.githubusercontent.com/122595269/221036138-abbf4b77-9707-4a6f-ae85-1bdb729f77ea.mp4

I mean that the list doesn't scroll to any dependable position; it's often just kind of in the middle without any rhyme or reason.  Even the one that is erroneously selected, because of point 2, isn't always in view with a blank search box.  Hope that explains better.

PR #1541, opened 2023-02-27
Fix parent for 'Viewers', 'Coding', 'Structure' and 'MIDI' submenus

### LINKED ISSUE #1462: Menu sometimes detached from its parent [Wayland]
On Fedora, sometimes, when browsing menus, a submenu is detached from its parent menu and appears on a random location on the screen.

image

When running Frescobaldi through the command line, it prints this warning on the console:

```
qt.qpa.wayland: setGrabPopup called with a parent, QtWaylandClient::QWaylandXdgSurface(0x55c199d56540) which does not match the current topmost grabbing popup, QtWaylandClient::QWaylandXdgSurface(0x55c19973e190) According to the xdg-shell protocol, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly because xdg-shell mandate that child popups close before parents
```

jeanas:
These need to be owned by their parent menu to be correctly positioned, at least on Wayland. From the documentation of QMenu::addMenu(QMenu *menu):

  This menu does not take ownership of menu.

Fixes #1462

### DIFF
--- a/frescobaldi_app/panelmanager.py
+++ b/frescobaldi_app/panelmanager.py
@@ -49,19 +49,19 @@ class PanelManager(plugin.MainWindowPlugin):
         self._panels = []
         self._submenus = collections.OrderedDict([
             ('viewers', {
-                'menu': QMenu(_('&Viewers')),
+                'menu': _('&Viewers'),
                 'panels': []
             }),
             ('coding', {
-                'menu': QMenu(_('&Coding')),
+                'menu': _('&Coding'),
                 'panels': []
             }),
             ('structure', {
-                'menu': QMenu(_('&Structure')),
+                'menu': _('&Structure'),
                 'panels': []
             }),
             ('midi', {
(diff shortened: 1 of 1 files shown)


fedelibre:
I confirm it fixes the issue.

When I launch it from a terminal and go through the menus, I see this message printed sometimes:

```
qt.qpa.wayland: setGrabPopup called with a parent, QtWaylandClient::QWaylandXdgSurface(0x5582dfdf0490) which does not match the current topmost grabbing popup, QtWaylandClient::QWaylandXdgSurface(0x5582e1327d80) According to the xdg-shell protocol, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly because xdg-shell mandate that child popups close before parents
```

Is this a different issue, I guess?

PR #1559, opened 2023-03-18
Use GitHub issues for letting users report bugs

### LINKED ISSUE #1558: Bug reporting system
Currently, when it crashes, Frescobaldi allows the user to write an email to @wbsoft with the traceback. Since @wbsoft currently has less time for Frescobaldi than he used to, this is maybe not ideal. Perhaps we should change it to let the user create a new issue on GitHub instead.

jeanas:
Fixes #1558

### DIFF
--- a/frescobaldi_app/exception.py
+++ b/frescobaldi_app/exception.py
@@ -46,11 +46,11 @@ class ExceptionDialog(QDialog):
         # _tbfull is the full traceback
         tbfull = traceback.format_exception(exctype, excvalue, exctb)
         self._tbfull = ''.join(tbfull)
-        # _tbshorter is the traceback, but truncated to 5 frames max.
+        # _tblimited is the traceback, truncated to 5 frames max.
         # We do this because of size limits on what you can pass to GitHub
         # in a URL.
-        tbshorter = traceback.format_exception(exctype, excvalue, exctb, limit=5)
-        self._tbshorter = ''.join(tbshorter)
+        tblimited = traceback.format_exception(exctype, excvalue, exctb, limit=5)
+        self._tblimited = ''.join(tblimited)
 
         self._ext_maintainer = app.extensions().is_extension_exception(tbfull)
 
@@ -122,6 +122,6 @@ class ExceptionDialog(QDialog):
                 + _("Optionally describe below what you were doing:"),
                 recipient=rcpt)
         else:
-            body = (self._tbshorter + '\n'
+            body = (self._tblimited + '\n'
                     + _("Optionally describe below what you were doing:"))
             bugreport.new_github_issue(title=self._tbshort, body=body)


PeterBjuhr:
Yes, I think it could be good to have an alternative. Would that option require you to join the mailing list?

frescobaldi_app/bugreport.py
```diff
@@ -18,16 +18,29 @@
...
+    Open a web brower on a page to create a new issue on GitHub.
+
+    Information about the versions of Frescobaldi and its dependencies will
+    be appended to the body.
+    """
+    body += "\n\n" + debuginfo.version_info_string()
+    url = QUrl(appinfo.issues_url)
+    query = QUrlQuery()
+    query.addQueryItem("title", title)
+    query.addQueryItem("body", body)
```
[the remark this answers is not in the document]
PeterBjuhr: I think we could try to truncate the traceback, with the risk of course that we'll miss something important. Often we'll need to have a dialogue with the user anyway.

fedelibre:
> Yes, I think it could be good to have an alternative. Would that option require you to join the mailing list?

If you use the website interface, you need a Google account. I'm afraid a user account is required everywhere to prevent spam.

IMO pointing to GitHub is better, but why not also mentioning the Google Group? Let's offer two options and let the user decide.
Maybe an issue template could be a good idea, but it can be added later.

fedelibre:
> From the user point of view, if two options are given, I think it's helpful if we state what the advantage of each of them is. In your opinion, what could be the deciding factors between the mailing list and a GitHub issue directly?

I think we should point the user to GitHub. It's better for everybody.
The mailing list should be the last resort.

Some considerations:

- The user may already have a Google account and be tempted to use the Google Group to avoid the creation of a new account. Let's highlight that using a GitHub account allows the user to provide feedback to developers, follow an issue and receive an email if a fix is coming.

- Let's try to engage the users and let them join the place where the development happens (currently GitHub). As we have small resources, we want the user to open new issues, not the developers. You may find a nice way to say it :-)
You know from LilyPond experience that you'll have to open a new issue on behalf of the user if the user first send the report to a mailing list. I think it's not the best way to do it. A good issue template may prevent many invalid issues, hopefully.

- We might say that the mailing list is mainly for users to help each other, while GitHub is the place where developers listen for feature requests, etc.

PeterBjuhr:
> I think we should point the user to GitHub. It's better for everybody.
The mailing list should be the last resort.

I agree, the mailing list should be considered plan B.

frescobaldi_app/appinfo.py
```diff
@@ -35,6 +35,7 @@
 maintainer_email = "info@frescobaldi.org"
```
fedelibre, line 35: Should we change this email address to `frescobaldi@googlegroups.com`? Don't know if you have to be subscribed.

frescobaldi_app/exception.py
```diff
@@ -40,9 +41,17 @@ class ExceptionDialog(QDialog):
...
 
+        # _tbshort is the exception line only (last line of the traceback)
         self._tbshort = ''.join(traceback.format_exception_only(exctype, excvalue))
+        # _tbfull is the full traceback
         tbfull = traceback.format_exception(exctype, excvalue, exctb)
         self._tbfull = ''.join(tbfull)
+        # _tbshorter is the traceback, but truncated to 5 frames max.
+        # We do this because of size limits on what you can pass to GitHub
+        # in a URL.
+        tbshorter = traceback.format_exception(exctype, excvalue, exctb, limit=5)
```
PeterBjuhr, line 52: nitpick: I fear the wording here is potentially misleading thinking that shorter is even shorter than "short". Could it be called "tbmedium" or perhaps "tblimited"?

PeterBjuhr (approved):

fedelibre (approved):
Ok, merge it! Thanks

PR #1565, opened 2023-03-24
Always respect the shebangs of LilyPond's Python tools on macOS

### LINKED ISSUE #1492: "Force use of the tools’ #! lines" needs to be on in order to let Frescobaldi recognize convert-ly in LilyPond 2.24
"Force use of the tools’ #! lines" in the "Tool Commands" tab in the Edit dialog for the "LilyPond version to use" in the "LilyPond Preferences" needs to be checked on in order to get LilyPond 2.24 appeared in the pull down menu in the Convert-ly dialog. I'm not sure whether this is expected behavior or not. I found this on macOS Monterey 12.6.2 with Frescobaldi 3.2 installed with MacPorts and "official" lilypond-2.24.0-darwin-x86_64.tar.gz binaries.

jeanas:
According to the docstrings, a system Python was searched because the LilyPond-provided one did not function correctly.  With current versions of LilyPond, which use a completely rewritten system for generating the binaries instead of GUB, it does.

Fixes #1492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ All notable changes to the Frescobaldi project are documented in this file.
 
 - Require Python 3.8 (#1519)
 - Make bug reports sent via GitHub instead of by email (#1559)
+- Always respect the shebangs of LilyPond's Python scripts on macOS
+  and remove the corresponding configuration option (#1565)
 
 ### Removed
 


frescobaldi_app/macosx/__init__.py
```diff
@@ -55,97 +55,3 @@ def use_osx_menu_roles():
     except NameError:
         _use_roles = inside_app_bundle()
     return _use_roles
-
-def midi_so_arch(lilypondinfo):
```
PeterBjuhr, line 59: question: Are these deletes related? Perhaps a separate commit?

PeterBjuhr (approved):

dliessi (approved):
All these changes are fine, but there is one missing: we should delete the preference `useshebang` by means of `install/update.py`.
Anyway @jeanas you can already merge this PR: I'm going to add a follow-up commit (here if before you merge it, otherwise directly to master).

I am in favour of this change, but for the record I want to point out that this change may break the use of LilyPond's Python tools for older versions of LilyPond.
In my opinion this is not a problem, as convert-ly and imports usually make sense only with a recent version of LilyPond, if not the latest, and this PR has no effect on running LilyPond proper.

dliessi:
You are right that unused preferences don't hurt, unless the same name is reused at a later time (the old preference value would be picked up). Even though that's only a theoretical risk, I would prefer to delete the unused preference entry.

fedelibre (approved):

fedelibre:
Jean, I think you can merge it.

I don't see any other open PR ready to be included in tomorrow release.
Let's release 3.3.0 tomorrow, then we can make a bug fix release (3.3.1) in a few weeks to include other stuff.

I'll make the release tomorrow in the afternoon, Ok?
Then I would wait until Mac and Windows packages are ready (if it doesn't take too long) before announcing the release in the mailing lists.

PR #1568, opened 2023-03-25
Remove warnings that -dembed-source-code requires LilyPond 2.19.39+

jeanas:
This version is old enough by now that it should be safe to assume the
user has a later version.

### DIFF
--- a/frescobaldi_app/preferences/lilypond.py
+++ b/frescobaldi_app/preferences/lilypond.py
@@ -337,8 +337,7 @@ class Running(preferences.Group):
         self.embedSourceCode.setText(_("Embed Source Code files in publish mode"))
         self.embedSourceCode.setToolTip(_(
             "If checked, the LilyPond source files will be embedded in the PDF\n"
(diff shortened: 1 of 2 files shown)


frescobaldi_app/preferences/lilypond.py
```diff
@@ -337,8 +337,7 @@ def translateUI(self):
         self.embedSourceCode.setText(_("Embed Source Code files in publish mode"))
         self.embedSourceCode.setToolTip(_(
             "If checked, the LilyPond source files will be embedded in the PDF\n"
-            "when LilyPond is started in publish mode.\n"
-            "This feature is available since LilyPond 2.19.39."))
```
PeterBjuhr, line 341: question: Couldn't we keep the reference to the Lilypond version?

PeterBjuhr: Yes, I agree it should be more useful than a historical reference.

PeterBjuhr (approved):

PR #1597, opened 2023-05-02
Use standard (modern) Python packaging

### LINKED ISSUE #1518: Migrate from setup.py to new packaging standards and away from distutils
Currently, Frescobaldi uses the traditional `setup.py` file to define its metadata with setuptools. However, the modern recommended way is the `pyproject.toml` file. Furthermore, if setuptools is not available, it falls back on distutils, which is deprecated and will be removed in Python 3.12. Moving to `pyproject.toml` would allow getting rid of that, since one of the initial goals of `pyproject.toml` was to enable projects to specify their build/packaging tools, allowing to specify a dependency on setuptools.

jeanas:
This is an overhaul of all packaging-related files to comply with current Python packaging standards, which addresses the deprecation of the distutils module for Python 3.12, and should solve a certain number of issues caused by the use of ancient methods.  (For example, `pip uninstall frescobaldi` should now work better.)

* Move metadata from setup.py and setup.cfg to pyproject.toml.

* Use hatch as build backend.  Its support for data files seemed easier to use than setuptools' (the latter is also commonly considered overcomplex due to all the legacy it supports).

* Do not invoke `setup.py` directly in the Windows scripts; this is deprecated (and `setup.py` is removed anyway).  Use the compileall module instead.

* Rework the `INSTALL.md` file to use `pip` for installation, rather than `setup.py` invocation.

* Remove `MANIFEST.in`, which is obviated by hatch as it reads .gitignore automatically (we just need to include autogenerated files in pyproject.toml).

* Remove the top-level `frescobaldi` script, as modern Python packaging does not support direct installation of script files.  Move its functionality to `frescobaldi_app/main.py`; pip automatically creates a wrapper script that runs frescobaldi_app.main.main().

Fixes #1518


fedelibre:
In fact I've always missed this installation feature. Good to know that it is supported in the wheel format!

sincere-music:
The documentation in INSTALL.md seems more daunting than it needs to be right now, and I think a different structure of headers could help. Should I suggest such headers in the review here or prepare another pull request once this is merged?

fedelibre:
> @fedelibre Do you happen to know where the icon file should be put? Right now, `python setup.py install --user` puts it as `~/.local/share/icons/hicolor/scalable/apps/org.frescobaldi.Frescobaldi.svg`, but over here (Fedora 38) the icon isn't found; I get the generic icon
> 

It seems correct. Don't know why it's not working for you...

Nodd:
The top level script was useful to simply run frescobaldi from the git directory. There are other means to do it, but it was so easy! Would you consider keeping an equivalent of it ?

sincere-music:
Sounds good as well, no duplication.

INSTALL.md
```diff
@@ -1,119 +1,181 @@
...
-    sudo python3 setup.py install --prefix=/usr
+**In this case, be sure to grant the Frescobaldi application access on your home
+directory in the Flatseal application, or you will not be able to use LilyPond
+files other than in specific directories (like `~/Documents`) due to the
+sandboxing that Flatpak imposes.**
 
-If you have a Debian-based system such as Ubuntu, and you get the error
-message "ImportError: No module named frescobaldi_app.main", try:
+Also, if you previously installed Frescobaldi via your distribution, remember to
+remove it before installing it through Flatpak.
```
fedelibre, line 30: I know you wrote this to make things simple for the user. It's good that distro package installation and flatpak installation can coexist, but it's unfortunate that currently you see only one application in the desktop environment. Maybe we can tell the full truth to let the user make his/her choice. For example:

> "If you previously installed Frescobaldi via your distribution, be aware that it can coexist with the Flatpak installation but your desktop environment might launch only the distro package. You have two options: 1- open the flatpak app from a terminal with the command `flatpak run org.frescobaldi.Frescobaldi`; 2- remove the distro package. "

Writing complete and concise doc is difficult. I don't like my wording, but you see what I'm trying to communicate.

INSTALL.md
```diff
@@ -1,119 +1,181 @@
...
+### Uninstalling other versions
 
-You should also generate the desktop and AppStream metainfo files, as the
-repository contains only the template without translations:
+While not required, it is a good idea to save yourself confusion in the future
+by uninstalling other versions of Frescobaldi. For example, `sudo apt remove
+frescobaldi` if you installed via the system package on Debian/Ubuntu, `sudo dnf
+remove frescobaldi` on Fedora, or `flatpak remove org.frescobaldi.Frescobaldi`
+if you installed it via Flatpak, or `pip uninstall frescobaldi python-ly
+qpageview` if you used Pip.
```
fedelibre, line 48: Later you suggest using pipx. Should this be pipx as well?

fedelibre (approved):

PR #1600, opened 2023-05-03
Move functionality from "./frescobaldi" to "./frescobaldi_app/__main__.py"

jeanas:
This is split out from #1597.

We want to move away from using setuptools' `data_files` argument, which is deprecated. But we still want this functionality, for installing the desktop files, the icon and the man page.

In setuptools, there is no non-deprecated replacement for it. On the other hand, Hatch supports an equivalent. However, due to the way Hatch autodiscovers the files that should be put into wheels, we need the "frescobaldi_app" directory renamed to "frescobaldi" (to match the overall project's name), which clashes with the existing "./frescobaldi" script.

This moves the functionality from that script to `frescobaldi_app/__main__.py` and reduces its use; the script itself will be removed in a separate PR.

The new way to run Frescobaldi from source without installing is `python -m frescobaldi_app` (soon to become `python -m frescobaldi`).

### DIFF
--- a/frescobaldi_app/__main__.py
+++ b/frescobaldi_app/__main__.py
@@ -161,7 +161,7 @@ def check_ly():
     sys.exit(1)
 
 
-def main():
+def main(debug=False):
     """Main function."""
     app.instantiate()               # Construct QApplication object
     args = parse_commandline()
@@ -266,7 +266,11 @@ def main():
         view.centerCursor()
 
     sys.excepthook = app.excepthook # Show Python errors in a bugreport window
-    sys.exit(app.run())             # Go!
+
+    app.appStarted() # Emit the appStarted signal
+
+    if not debug:
+        sys.exit(app.run())
 
 if __name__ == "__main__":
     main()
--- a/frescobaldi_app/app.py
+++ b/frescobaldi_app/app.py
@@ -159,8 +159,7 @@ def oninit(func):
     return func
 
 def run():
-    """Emit the appStarted signal and enter the Qt event loop."""
-    appStarted()
+    """Enter the Qt event loop."""
     result = qApp.exec_()
     aboutToQuit()
     return result
--- a/frescobaldi_app/debug.py
+++ b/frescobaldi_app/debug.py
@@ -10,10 +10,11 @@
 
 import sys
 
+from frescobaldi_app.__main__ import main
+
 from . import toplevel
 toplevel.install()
 
-import main
 import app
 import document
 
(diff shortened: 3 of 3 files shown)


dliessi:
I'm going to test on Mac later tonight.
I already see a problem: the new version of both the Mac and the Windows scripts is looking for `frescobaldi/__main__.py` instead of `frescobaldi_app/__main__.py`.
I'll fix it locally and test both runnign from source and building the application bundle.

fedelibre:
I managed to run your main branch from source on Windows! (well, I don't have python-poppler-qt5 installed of course! but it's not a hard dependency)

I thought it was difficult, but it was pretty easy actually. Thanks to the instructions in #1597

macosx/mac-app.py
```diff
@@ -49,7 +49,7 @@
 parser.add_argument('-s', '--script', \
   help = 'path of {0}\'s main script; you should use an absolute path, \
   so that the application bundle can be moved to another \
-  directory'.format(appinfo.appname), default = '{0}/{1}'.format(root, appinfo.name))
+  directory'.format(appinfo.appname), default = '{0}/{1}/__main__.py'.format(root, appinfo.name))
```
dliessi, line 52: In this PR you do not rename the `frescobaldi_app` directory, so this should be `default = '{0}/{1}_app/__main__.py'.format(root, appinfo.name)`.
I made this change locally and tested building the application bundle and it worked.

windows/freeze.py
```diff
@@ -84,7 +84,7 @@
 shutil.rmtree(target_dir, ignore_errors = True)
 
 frescobaldi = Executable(
-    '../frescobaldi',
+    '../frescobaldi/__main__.py',
```
dliessi, line 87: This should be `'../frescobaldi_app/__main__.py'`.

windows/frescobaldi-wininst.py
```diff
@@ -41,7 +41,7 @@
 
 
 # Main application script and Python executable:
-script = os.path.join(sys.prefix, 'Scripts', 'frescobaldi')
+script = os.path.join(sys.prefix, 'Scripts', 'frescobaldi', '__main__.py')
```
dliessi, line 44: I believe this should be `os.path.join(sys.prefix, 'Scripts', 'frescobaldi_app', '__main__.py')`, but I'm not sure (where does `Scripts` come from?) and cannot test on Windows now.

dliessi (changes requested):
Running from source on Mac seems to work correctly, as does building the application bundle after applying the change I request on `macosx/mac-app.py`.

dliessi (approved):

fedelibre:
I'm fine with going ahead anyway. We can fix any bug later.

I've always wanted to test the Windows installer and I have a Windows laptop.. but no time at the moment. I will have some time for this in a couple of weeks. But I wouldn't hold this PR until then.

PR #1602, opened 2023-05-07
Use "org.frescobaldi.Frescobaldi" as icon name

### LINKED ISSUE #1582: no application icon in Windows
This is a minor issue. I haven't checked yet but I guess it's due to the renaming of icon file to org.frescobaldi.Frescobaldi...

Schermata del 2023-04-05 10-17-57

jeanas:
Should fix #1582

### DIFF
--- a/frescobaldi_app/main.py
+++ b/frescobaldi_app/main.py
@@ -190,7 +190,7 @@ def main():
 
     # application icon
     import icons
-    QApplication.setWindowIcon(icons.get("frescobaldi"))
+    QApplication.setWindowIcon(icons.get("org.frescobaldi.Frescobaldi"))
 
     QTimer.singleShot(0, remote.setup)  # Start listening for IPC
 
@@ -252,4 +252,3 @@ def main():
         view = win.currentView()
         view.setTextCursor(cursor)
         view.centerCursor()
-


frescobaldi_app/icons/__init__.py
```diff
@@ -42,9 +42,10 @@ def get(name):
         return _cache[name]
     except KeyError:
         icon = _cache[name] = QIcon()
-        # first try SVG
-        fname = 'icons:{0}.svg'.format(name)
-        if QFile(fname).exists():
+        # first try SVG and ico
+        if QFile(fname := 'icons:{0}.svg'.format(name)).exists():
```
PeterBjuhr, line 46: suggestion: (Saving one line and avoiding some repetitious code)
```
for ext in ['svg', 'ico']:
    if QFile(fname := f'icons:{name}.{ext}').exists():
        icon.addFile(fname)
```

PeterBjuhr: Sorry, maybe we want a break if the svg extension works - and then the line count will be the same!? :slightly_smiling_face:

PeterBjuhr (approved):

19joho66:
No, the icon was ok before and is also ok now

fedelibre:
> I borrowed a Windows machine to test this. However, #1582 doesn't reproduce in the first place. This was the result, both without and with this patch

The screenshot seems to be run from the installer (no git menu), so it's 3.3.0 I guess.
How did you test this patch? Did you run it from source?

I cannot run the current master from source on Windows. Wait, I can if I merge #1600 into master... and now the Frescobaldi icon appears in the applications bottom bar and in the top bar. So this patch seems to solve the problem for me.
In the same Windows computer the Frescobaldi installed from the 3.3.0 installer doesn't display any icon.

PR #1614, opened 2023-05-21
[fix #1546] Quick Insert: \melisma and \melismaEnd after all other spanners

### LINKED ISSUE #1546: Quink Insert: Beam and Melisma applied in sequence
I'm working with Frescobaldi 3.1.3. I have encountered anomalous behavior during the insertion of Beams and Melismas through the _Quick Insert_ tools.
When I try to put (by _Quick Insert_) a melisma around a group of notes and then to insert a beam around the same group, I obtain this code

> `c8[\melisma c c c]\melismaEnd`

that Lilypond compile correctly, without warnings.

If instead I change the order of insertion (beam and than melisma), the resulting code is 

> `c8\melisma[ c c c\melismaEnd]`

and Lilypond warn an erroneous _BeamEvent_, reporting the illegal positions of the brackets after the `\melisma` command.

When I have to insert both beams and melismas I bypass the problem by reminding to insert first the melisma. But it gets tedious when I have to put melismas in a score where beaming is already present and I have to correct manually every bracket position.

It would be possible to fix the behavior of the two commands so that you get the same code in both combinations (melisma->beam, beam->melisma)?

sesquialtera87:
I've tried to fix the positioning of the `\melisma` and `\melismaEnd` commands from the QuickInsert Tool (issue that I had detected in #1546).

To change as little as possible, I added only a workaround during the calculation of the spanner locations. In the preceding code, all spanners were added right after the note, causing a bug when a melisma was inserted with other spanners already present. So, during insertion of melismas, all the spanner symbols attached to the notes (first and last in the selection) are scanned and the two commands `\melisma` and `\melismaEnd` are placed at the end index of the last detected spanner token.

I did several test with various combinations of the spanners and everything now seems to be going right.

I've only recently looked at Frescobaldi code, so I assume that some function that I have written by hand as utility to search for Lilypond code elements is already present in some package: I apologize for that.


frescobaldi_app/quickinsert/spanners.py
```diff
@@ -283,10 +278,16 @@ def spanner_positions(cursor):
         partial = ly.document.OUTSIDE
 
     items = list(ly.rhythm.music_items(c, partial=partial))
-    if cursor.hasSelection():
-        del items[1:-1]
+
+    if command.endswith('melisma'):
```
PeterBjuhr, line 282: thought: At first glance it seems better to add an own function for melismas than to add a special case in this function.

frescobaldi_app/app.py
```diff
@@ -137,7 +137,7 @@ def instantiate():
     QApplication.setDesktopFileName(appinfo.desktop_file_name)
     QApplication.setOrganizationName(appinfo.name)
     QApplication.setOrganizationDomain(appinfo.domain)
-    qApp.focusChanged.connect(clear_selection)
+    # qApp.focusChanged.connect(clear_selection)
```
PeterBjuhr, line 140: question: Is this some forgotten debug change?

frescobaldi_app/quickinsert/spanners.py
```diff
@@ -267,13 +268,7 @@ def actionTriggered(self, name):
                     cursor.insertText(outer[0])
 
 
-def spanner_positions(cursor):
-    """Return a list with 0 to 2 QTextCursor instances.
```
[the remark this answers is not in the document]
sesquialtera87: Sorry, I didn't notice that. It was unintentionally deleted

frescobaldi_app/app.py
sesquialtera87: Yes, I forgot to restore it. That line was commented out because it clears the selection and so Quick Insert doesn't work anymore over selection when firing action with the mouse, clicking on the buttons in the panel.

frescobaldi_app/quickinsert/spanners.py
sesquialtera87: I thought so too, but being the first time I intervened on the code I didn't dare to extrapolate the command into a separate function. An independent function would certainly make everything clearer.
I could improve it by putting melisma insertion in its own function.

frescobaldi_app/quickinsert/spanners.py
```diff
@@ -34,19 +32,21 @@
...
 class Spanners(tool.Tool):
     """Dynamics tool in the quick insert panel toolbox."""
+
     def __init__(self, panel):
         super(Spanners, self).__init__(panel)
         self.removemenu = QToolButton(self,
-            autoRaise=True,
-            popupMode=QToolButton.InstantPopup,
-            icon=icons.get('edit-clear'))
+                                      autoRaise=True,
```
PeterBjuhr, line 47: question: This indent looks strange. Are you using tabs?

frescobaldi_app/quickinsert/spanners.py
```diff
@@ -296,6 +300,107 @@ def spanner_positions(cursor):
...
+    """ Return the correct positions of the melisma commands, after all spanners already attached to the note """
+
+    def search_spanners_symbols(item1, item2, spanners) -> int:
+        """Find the spanners in the list lying between two music items.
+        :param item1: First item
+        :param item2: Second item
+        :param spanners: A list of spanners to be examined
+        :return: The position in the document of the last spanner attached to the first item, or None if nothing is found
+        """
+
```
PeterBjuhr, line 313: nitpick: (I know that we are not exactly following PEP style guides on every line of the code. But for new additions I will try to remark on stuff that I observe) - "There's no blank line either before or after the docstring."

frescobaldi_app/quickinsert/spanners.py
```diff
@@ -296,6 +300,107 @@ def spanner_positions(cursor):
...
+    :param doc: a lilypond document
+    :param pos: int indicating a position in the document
+    :return: (token.pos, token.end)
+    """
+    cursor = ly.document.Cursor(doc, pos)
+    source = ly.document.Source(cursor, True, ly.document.INSIDE, True)
+
+    for token in source:
+        if token.pos >= pos:
+            if isinstance(token, Note) or isinstance(token, Rest) \
```
PeterBjuhr, line 377: suggestion: `isinstance` takes a tuple of types as the second argument as in this borrowed example `isinstance("Hello", (float, int, str, list, dict, tuple))` so you don't need to do a lot of "ors".

PR #1625, opened 2023-06-10
Fix reloading in manuscript viewer

jeanas:
This seems to work, even though I'm not 100% sure it's the right fix yet.


PeterBjuhr:
Did we have an issue for this fix?

PR #1638, opened 2023-06-30
From Make to tox

### LINKED ISSUE #1113: build tool for Frescobaldi
As I wrote on issue #673 , we have several reasons to include a build tool (besides setup.py):

- handle installation and uninstallation cleanly (as a distro package does), while setuptools doesn't offer a clean uninstall and you basically have to do it manually (which is error prone).
- copy share_files in the install directory. See issue #673 
- .desktop file is currently not translated. We should use a .desktop.in file to be translated using gettext. This should be automated by a build tool. Same for the AppData file, still to be added - issue #879 
- PO files are currently handled by a Makefile in the po/ directory, but we are not happy about the current status (e.g. mo files tracked in git), see issue #1099 

In general, I'd like to have a *single tool*, available at top directory, to manage any aspect: update a po file, update a desktop file, build the source to test a branch (this will be needed if we remove the .mo files), install in the system, etc.

As build tool I recommend [Meson](https://mesonbuild.com/), which has a friendly and pythonic syntax (it's written in python, BTW).
Meson (or rather ninja backend) supports only out-of-tree builds. This is intended as a feature and not a limitation.
If nobody objects, I may give it a try.

@wbsoft @uliska Let me know what you think

jeanas:
Fixes #1113

### DIFF
--- a/i18n/frescobaldi/cs.po
+++ b/i18n/frescobaldi/cs.po
@@ -10,7 +10,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: frescobaldi 0.7.5\n"
 "Report-Msgid-Bugs-To: frescobaldi@googlegroups.com\n"
-"POT-Creation-Date: 2023-06-17 20:04+0200\n"
+"POT-Creation-Date: 2023-06-17 20:09+0200\n"
 "PO-Revision-Date: 2020-11-29 20:51+0100\n"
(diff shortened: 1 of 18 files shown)


fedelibre:
I'll test it as soon as I have some time. Please wait a few more days

fedelibre (approved):

PR #1643, opened 2023-07-01
Clean up unused icons

jeanas:


fedelibre:
I think the PR is Ok, but I would keep the splash.xcf file since it's the "source" used to generate the .png file and it might be useful in case we have to make any modification in the future.

fedelibre (approved):

PR #1646, opened 2023-07-07
Display error for unfound language

jeanas:
In case someone installed Frescobaldi but missed the MO generation step, it's friendlier to print a message than to fail with an assertion error.


PeterBjuhr:
Looks good, I think! Approving!

PeterBjuhr (approved):

PR #1652, opened 2023-08-01
i18n: Remove the deduplication between the frescobaldi and userguide domains

jeanas:
The previous po-update script took care of not inserting messages into
userguide PO files if they were already in the frescobaldi PO files, to
avoid duplicate work. However, this complicates the Sphinx support
because we cannot use the existing Sphinx translation workflow tools due
to this peculiarity. Moreover, this might lead to messages that need
context to be translated differently between frescobaldi and userguide,
but it's suboptimal because there is no way to add a translation context
in the userguide.

This adds a copy of all those messages to the userguide PO file and
adapts the PO/MO generation scripts to make the frescobaldi and
userguide domains independent.

### DIFF
--- a/frescobaldi_app/i18n/__init__.py
+++ b/frescobaldi_app/i18n/__init__.py
@@ -39,7 +39,7 @@ def available():
     """
     return [str(path.parent.stem) for path in Path(modir).rglob("LC_MESSAGES")]
 
-def translator(language):
+def translator(language, domain="frescobaldi"):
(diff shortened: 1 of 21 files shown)


PeterBjuhr:
I didn't actually manage to take a look. But I can take a quick look now! :+1:

PeterBjuhr:
I didn't find any issues! 

I hope the remorse was not due to some post-merge problem!

PeterBjuhr:
Great!

PR #1659, opened 2023-08-05
Fix Extensions.is_extension_exception

### LINKED ISSUE #1658: Exceptions are silenced on Windows
As evidenced in https://github.com/frescobaldi/frescobaldi/pull/1655#issuecomment-1666263609, there is a bug on Windows where an exception occurs while we're trying to display a bug report window for another exception.

Consequently, internal errors are currently being ignored on Windows, and not shown to the user with a button to report like they should be.

jeanas:
This method had two problems: first, there was a missing regex escaping leading to errors on Windows, and second, it did probably not work (not actually tested though) if the error was raised in a file deep inside the extension (like /.../extension/subpackage/subsubpackage/foo.py rather than /.../extension/foo.py) due to a ".*" in a regex that should have been ".*?".

Fix these by using traceback and path manipulations instead of regexes, making the method more robust.

Closes #1658

### DIFF
--- a/frescobaldi_app/extensions/__init__.py
+++ b/frescobaldi_app/extensions/__init__.py
@@ -26,7 +26,7 @@ import os
 import pathlib
 import sys
 from time import perf_counter
-import traceback
+from traceback import extract_tb
 
 from PyQt5.QtCore import (
     QDir,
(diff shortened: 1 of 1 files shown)


frescobaldi_app/extensions/__init__.py
```diff
@@ -791,28 +793,24 @@ def installed_extensions(self):
         result = list(self._infos.keys())
         return sorted(result)
 
-    def is_extension_exception(self, traceback):
+    def is_extension_exception(self, tb):
```
PeterBjuhr, line 796: question: Can't we keep "traceback"? That is a bit more verbose.

PeterBjuhr: Yes, of course. We can't use that now. I guess it's alright then. Or could you come up with something less vague?

PeterBjuhr: Another idea could perhaps be to `import extract_tb from traceback` to avoid the name collision!??

PeterBjuhr (approved):

PR #1663, opened 2023-08-27
Fix column inconsistency

igneus:
This

- fixes a new bug introduced in my recent PR #1662 (`--column` values are 1-based, not 0-based, the default value wasn't taking this in account)
- addresses another bug pre-existent in the code base: different internal APIs are used for cursor position setting when starting a new Frescobaldi instance vs. when processing a remote `set_cursor` command. The latter internal API expects 0-based values, which was not accounted for in case of `--column`, resulting in inconsistent behaviour of `--column`

### DIFF
--- a/frescobaldi_app/remote/api.py
+++ b/frescobaldi_app/remote/api.py
@@ -125,7 +125,7 @@ class Incoming(object):
         elif cmd == b'set_cursor':
             line, column = map(int, args)
             cursor = win.textCursor()
-            pos = cursor.document().findBlockByNumber(line - 1).position() + column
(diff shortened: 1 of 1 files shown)


PeterBjuhr:
I see no issues with the code, but perhaps someone wants to try it out before approve (and merge)?

igneus:
TBH I don't remember if the first commit makes an observable difference in the application's behavior and right now I didn't manage to quickly replicate such a difference, but it definitely improves internal data coherence and thus reduces complexity. The line and column values passed around internally when calling the remote API and handling the call should (if there is no grave reason for the contrary) adhere to the contract of the respective command line option values.

PR #1667, opened 2023-09-10
100 percent complete German localization in de-complete.po file

papioara:
I have added a new file de-complete.po file to the /i18n/frescobaldi directory, but left the de.po (the original one which was already there untouched. Since this is the first time ever that I make a PR on Github in my life, I may have chosen the wrong way to handle the matter. But since I do not know how the process of reviewing and approval works here, we have two kinds of de.po at the moment. Actually, in my own local frescobaldi installation, I have renamed the de-complete.po file to de.po, and the original de.po became de.po.bak
I'd be grateful to learn how to do better in order to save everyone some unnecessary work  ;-)


papioara:
> git mv de-complete.po i18n/frescobaldi/de.po
If I do this in /frescobaldi/i18n/frescobaldi, does this command rename
de-complete.po to de.po? I tried it anyway, but I got the error:

fatal: nicht unter Versionskontrolle, Quelle=i18n/frescobaldi/de-
complete.po, Ziel=i18n/frescobaldi/i18n/frescobaldi/de.po

meaning: fatal, not under version control, Source=i18n/frescobaldi/de-
complete.pl, Target=i18n/frescobaldi/de.po

What's wrong?

Seems I miss something...

Am Sonntag, dem 10.09.2023 um 05:37 -0700 schrieb Jean Abou-Samra:
> You should not add a new file but modify the existing
> `i18n/frescobaldi/de.po` file. The whole point of using Git is that
> we don't need backup files or such because everything is recorded in
> the Git history.
> 
> You can fix your PR using
> 
> ```
> git mv de-complete.po i18n/frescobaldi/de.po
> git commit -m "Modify de.po instead of adding new file"
> git push
> ```
>

papioara:
Hello again, I think I understood what was missing: I did not do 'git
add de-complete.po' before doing the three commands below.
After that everything completed without any errors, so I assume we are
fine now.
Thanks for your support once more.

Am Sonntag, dem 10.09.2023 um 05:37 -0700 schrieb Jean Abou-Samra:
> You should not add a new file but modify the existing
> `i18n/frescobaldi/de.po` file. The whole point of using Git is that
> we don't need backup files or such because everything is recorded in
> the Git history.
> 
> You can fix your PR using
> 
> ```
> git mv de-complete.po i18n/frescobaldi/de.po
> git commit -m "Modify de.po instead of adding new file"
> git push
> ```
>

papioara:
It seems you guessed right. So I checked out switching to 'german
localization' by doing: 
git switch german-localization
Feedback in Terminal:
Zu Zweig »german-localization« gewechselt
Ihr Branch ist auf demselben Stand wie 'origin/german-localization'.
Switched to branch >>german-localization<<
Your branch is up to date with 'origin/german-localization'.

Then:
git mv de-complete.po i18n/frescobaldi/de.po
Feedback:
fatal: Ziel existiert bereits, Quelle=de-complete.po,
Ziel=i18n/frescobaldi/de.po
Fatal: Target exists already, Source=de-complete.po,
Target=i18n/frescobaldi/de.po

Then:
git commit -m "Modify de.po instead of adding new file"
Feedback:
Auf Branch german-localization
Ihr Branch ist auf demselben Stand wie 'origin/german-localization'.
nichts zu committen, Arbeitsverzeichnis unverändert
On branch german-localization
Your branch is up to date with 'origin/german-localization'.
Nothing to commit, working directory unchanged

Then:
git push
Feedback after logging in:
Everything up-to-date

Then I went on Github in order to see if something changed in my
'german-localization' branch, but no luck, it seems. No de.po with a
different message. Still 'Update all PO files'.
When I go to the master branch, however, there I see the de.po with the
message 'Modify de.po instead of adding new file'. Is this how it
should be? I would have expected the other way round.

And if things are hopelessly messed up by now: should I just ditch that
PR and start afresh rather? It is true unfortunately what it says in
the git manpage: "Git is a fast, scalable, distributed revision control
system with an unusually rich command set". So for a newbie the
learning curve is rather steep - at least it feels like this at the
moment. But I won't give up too soon...
I wonder if there is something like 'Git pour les nuls'. I have one
like that in French (which I speek, too) about Linux - 'Linux pour les
nuls' ;-)

Am Sonntag, dem 10.09.2023 um 14:41 -0700 schrieb Jean Abou-Samra:
> OK, I get the problem now. Judging from
> https://github.com/papioara/frescobaldi/commits/master, you have run
> these commands while on the `master` branch. But when you work on a
> PR, you should first check out that PR's branch (in this case, `git
> switch german-localization`).
>

papioara:
Will do this tomorrow, and also will take a look at the new strings - probably not a lot. Thank you so much again for your assistance and patience. My next 'victim' is already waiting: the user guide.
I am happy to be able to contribute to such a great and useful project thus giving something back to the community. It was a very thrilling job. Good translations, especially software ones, are tricky. It's quite easy to produce nonsense when you do not know the context of a message/a command. I may have been a little annoying at times, but I better keep asking until I know what I am doing. The results will definitely be better. And I have learned about features in Frescobaldi which are very interesting and ingenious, even if I personally do not use them. But offering all this Frescobaldi addresses a wide audience with quite different requirements - an almost all-purpose tool, or, as they say in Bavaria: an egg-laying, wool and milk-providing sow. :-)

papioara:
Fork deleted from the settings. Now here comes a list of new questions, before I start translating the newly added strings:

1. In my browser, when I did 'Save Link as', the resulting de.po could not be opened by poedit. It complained that it did not recognize the file format. Only creating a new text file, pasting the contents of de.po from the frescobaldi/frescobaldi master into it and saving it as de.po resulted in a file which could be opened by poedit. Is there an explanation for this?
2. In that de.po file now, I see a couple of untranslated strings, but also some with the label 'Needs Review'. It's good they are labeled, because some of those translations are wrong. I cannot remember having left them there like this, because I think, I really looked at every single one. How come they are in there now? Not many, thank God...
3. There are four about which poedit complains, because I did not add the colon, comma or full stop at the end. I left them out, because they are in part unnecessary or wrong in the English original. One of the tooltips erroneously has a comma at the end, although it should have a full stop, because it's the end of a sentence. But I guess it's all right if I ignore this for the German localization. Unfortunately I cannot change those little flaws myself.
4. After having finished my work and after having studied some more git: should I make a new fork and send in another PR?

In order to be on the safe side, however, I will proof-read everything once more, so no unnecessary mistakes will be left. It's a lot, but it's worth it.

papioara:
Too late for the fork...
So I'll just create a new one and keep it afterwards... PR will arrive in a while.

PR #1692, opened 2023-10-15
German localization revised

papioara:
User guide for version 3.0.0 plus small modifications in credits.md and i18n/frescobaldi/de.po


papioara:
I have seen what git complains about, but I do not quite understand what is wrong here. Please teach me for future situations like this...
Thank you

papioara:
Thanks a lot! In order to avoid situations like that in the future: would it be better to delete the branch after the PR (and successful merge with master) and open up a new one instead?

papioara:
Will do...

papioara:
Wonderful - so we can close this chapter for now. I have deleted the branch, and I will keep an eye on Frescobaldi development and possible changes in the menus and the user guide in a future version. Maybe on Christmas already... ;-)

PR #1732, opened 2024-02-06
Use standard clefs for guitar, bass guitar, and double bass (fixes #1571)

### LINKED ISSUE #1571: Proposal: Improve handling of transposing instruments in the Score Wizard
I'm going to split this into two parts because Frescobaldi handles instruments that transpose at the octave differently than instruments that transpose at other intervals.

## Transposition at the octave
Guitar, bass guitar, and double bass are all written an octave higher than their sounding pitch. Most publications I'm familiar with format these using the standard treble and bass clefs, respectively, leaving the transposition implied. However, Frescobaldi uses octave clefs for these instruments, presumably because LilyPond also uses that formatting in its own documentation.

I propose either:
* Always formatting these parts using standard clefs and using `\transposition` to alter the sounding pitch, **or**
* Adding an option to the "Score settings" page to toggle between standard and octave clefs for these instruments. (This would make more sense there than attached to an individual part, since presumably the user would want to format all such instruments in a score the same way.)

My own instinct is to provide the option for greater flexibility. I'll add that the Score Wizard already formats other instruments that transpose at the octave, such as the soprano recorder (written an octave below sounding pitch), using standard clefs.

## Transposition at other pitches
Many wind instruments in particular transpose at pitches other than the octave. For example, the B-flat trumpet is written a whole step above its sounding pitch. Right now the Score Wizard always creates parts for these instruments with a `\transposition` command after `\global`, so they sound correct when input at their written pitch. This is convenient when typesetting existing scores, but the current implementation has some drawbacks:
* The Score Wizard does not override the key signature for individual parts, so the user must set it manually. For example, if a piece is in C major, the user would need to manually add `\key d \major` to each B-flat trumpet part.
* When composing new music, some users (like me!) might prefer to input notes at concert pitch, then transpose them for display.

I propose adding an option to the "Score settings" page to allow:
* Inputting parts at their written pitch (the current behavior).
* Inputting parts at concert pitch and automatically transposing them for display.
* Both inputting and displaying parts at concert pitch.

When the first option is selected, I also propose having the Score Wizard automatically add the appropriate `\key` command to each part that needs it. This would save time, and also clarify that the part should be input at written rather than concert pitch.

bmjcode:
This PR corrects one of my previous submissions. I have learned that 8vb clefs not considered standard anywhere for either bass guitar or double bass, so rather than add them to the latter I should have removed them from the former. (I had always wondered why Frescobaldi used them myself, but assumed there was a reason it was programmed that way.)

From what I understand, the 8vb clef _is_ generally considered acceptable for guitar, but standard treble clef remains the more established convention. I thus changed it for consistency with the other octave-transposing instruments -- not just the basses but also the recorder family, pitched percussion, and so on. That said, I'm not as familiar with current practices for guitar notation, so feel free to skip or revert that one if I'm off the mark.

### DIFF
--- a/frescobaldi_app/scorewiz/parts/plucked_strings.py
+++ b/frescobaldi_app/scorewiz/parts/plucked_strings.py
@@ -201,6 +201,7 @@ class TablaturePart(_base.Part):
             if self.tabFormat:
                 tabstaff.getWith()['tablatureFormat'] = ly.dom.Scheme(self.tabFormat)
             self.setTunings(tabstaff)
+            builder.setMidiInstrument(tabstaff, self.midiInstrument)
             sim = ly.dom.Simr(tabstaff)
             if numVoices == 1:
                 ly.dom.Identifier(assignments[0].name, sim)
@@ -215,7 +216,6 @@ class TablaturePart(_base.Part):
             p = staff
         elif staffType == 1:
             # only a TabStaff
-            builder.setMidiInstrument(tabstaff, self.midiInstrument)
             p = tabstaff
         else:
             # both TabStaff and normal staff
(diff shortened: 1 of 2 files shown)


bmjcode:
Canceling this PR because I'm not satisfied with the implementation. I think I can do a cleaner job of this.

PR #1734, opened 2024-02-08
Improve handling of transposing instruments in the Score Wizard (fixes #1733)

### LINKED ISSUE #1733: Proposal: Enter all instruments at concert pitch in Score Wizard scores
Right now the Score Wizard has three different ways of handling transposing instruments:

* Instruments that transpose at the octave (except guitar) are entered at the written pitch.
* Instruments that transpose at other intervals are entered at concert pitch.
* Guitar, which normally transposes at the octave, is entered at concert pitch and notated using the `treble_8` staff.

This is confusing since users have no reason to expect this behavior. Simply knowing about the different conventions doesn't necessarily help; I may be able to explain them, but still get lost myself on octave-transposing instruments like soprano recorder.

I propose entering all instruments in Score Wizard scores at concert pitch and transposing them as needed for display. I believe the benefits of having a consistent scheme for all instruments are self-explanatory, and I believe concert rather than written pitch is the appropriate default because:

1. Transposition in this context is not intrinsic to the music, but a convention specific to staff notation.
2. Working in concert pitch makes it possible to implement things like the "display in concert pitch" mode requested in #318.
3. Working in concert pitch is far simpler to implement than working at the written pitch.

There are, of course, cases where entering music directly at written pitch is preferred, such as when typesetting existing music. This proposal does not directly address that case, but it does indirectly help, since it makes the transposition logic more transparent and therefore easier to manually override.

bmjcode:
This PR makes various improvements to the Score Wizard.

Most notably, it standardizes the handling of transposing instruments so that **music for all instruments is now entered at concert pitch**. Previously there was separate transposition logic for instruments that transpose at the octave vs. other intervals, with subtly different behavior for each. Now all instruments use the same transposition logic.

As part of this work, I have implemented the change I recommended in issue #1733 moving `\transposition` from the music assignment to the staff definition block. This should have several other benefits as detailed in the issue thread.

In a previous patch I changed double bass to use the `bass_8` clef to match the existing behavior for bass guitar. I have since learned this was incorrect, as octave clefs are not considered standard for either instrument. Part of my motivation for cleaning up the transposition logic was to correct this; both instruments now use the standard bass clef.

While the `treble_8` clef *is* sometimes used for guitar (including for most examples in LilyPond's documentation), my understanding is the standard clef remains more common. I have therefore changed guitar to use it for consistency with other transposing instruments.

Finally, this also fixes a bug I discovered during testing where a MIDI instrument is not assigned to the tab staff when both standard and tab staffs are selected for a part.

### DIFF
--- a/frescobaldi_app/scorewiz/build.py
+++ b/frescobaldi_app/scorewiz/build.py
@@ -23,7 +23,6 @@ Builds the LilyPond score from the settings in the Score Wizard.
 
 
 import collections
-import fractions
 import re
 
 import ly.dom
@@ -127,11 +126,6 @@ class PartData:
         ly.dom.Pitch(octave, 0, 0, stub)
         s = ly.dom.Seq(stub)
         ly.dom.Identifier(self.globalName, s).after = 1
-        if transposition is not None:
-            toct, tnote, talter = transposition
-            # this corrects the sounding pitch for MIDI after
-            # the written pitch is transposed by SingleVoicePart.build()
-            ly.dom.Pitch(toct, tnote, fractions.Fraction(talter, 2), ly.dom.Transposition(s))
         ly.dom.LineComment(_("Music follows here."), s)
         ly.dom.BlankLine(s)
         return a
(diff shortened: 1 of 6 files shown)


bmjcode:
I'm closing this PR for now to focus on the Qt 6 port, but I am continuing to work on these fixes in a separate [qt6-scorewiz branch](https://github.com/bmjcode/frescobaldi/tree/qt6-scorewiz).

PR #1743, opened 2024-03-06
Add a tag required by Flathub validation

fedelibre:
This tag is required to pass Flathub validation and publish new versions on Flathub.

As developer name I chose "Frescobaldi project".
Wilbert is the original author of course, but in this case I think we want to show who is actively developing and maintaining Frescobaldi.


fedelibre:
I'll merge it as it's a trivial change

PR #1780, opened 2024-08-17
Port Frescobaldi to PyQt6

bmjcode:
This is a port of Frescobaldi to PyQt6 as discussed under #1407. To run it you'll also need the [Qt 6 version of qpageview](https://github.com/frescobaldi/qpageview/pull/24).

All the basic features like engraving, previews, and MIDI playback work as expected, and most other things at least shouldn't crash. That said, I haven't yet tried every feature, and there are a couple known bugs I still need to fix. As with the qpageview port, I expect you'll want to keep this on a separate branch until more folks have tested it. Suggestions, patches, questions, and other feedback are welcome.

PR #1787, opened 2024-08-31
Don't crash when running from a git repo, but git is not on PATH

igneus:
How to reproduce the issue:

Clone the Frescobaldi repository. Run from the git-controlled sources while making sure that the git executable is not on PATH, e.g.

```sh
PATH= /usr/bin/python3 ./frescobaldi
```

It crashes miserably on startup.

(Originally reported [here](https://github.com/frescobaldi/frescobaldi/pull/1780#discussion_r1730039265).)


igneus:
Closing, the code is now part of #1780

PR #1812, opened 2024-11-19
Fix string that is passed to setDesktopFilename

filmor:
Currently, a warning is generated. According to the docs (and this warning), the
`QGuiApplication::desktopFileName` should be called without the `.desktop`
extension.

https://doc.qt.io/qt-6/qguiapplication.html#desktopFileName-prop


jeanas (approved):
Thanks!

Issue #1811, opened 2024-11-18
Change Copy to Image to use PNG only?


bmjcode:
The Copy to Image dialog provides a choice of file types, but only PNG and JPEG are currently supported under Qt 6 because of QtPdf limitations. Copying to any of the vector formats -- SVG, PDF, and EPS -- requires Poppler to work.

Even under Frescobaldi 3.3.0, only PNG and JPEG work for what seems like an obvious use case: copy-and-pasting a selection from Frescobaldi to LibreOffice Writer. (I may be wrong on this, but I think the [OOoLilyPond extension](https://github.com/OOoLilyPond) also uses PNG for that reason.)

And since JPEG's lossy compression is poorly suited to musical scores, PNG is really the only sensible choice left.

Does anyone object to this change?

jeanas:
I think it's fine. We're very unlikely to produce any new versions of Frescobaldi that relies on Poppler anyway.

bmjcode:
Implemented in #1815.

PR #1815, opened 2024-11-20
Always use PNG for copy to image (#1811)

bmjcode:
This PR changes the copy to image dialog to always use PNG as discussed under #1811.

I also increased the default copy resolution to 300 dpi because it better fits the use case of copying a selection to a document intended for print, like in Word or LibreOffice Writer for example.

(And sorry, I just realized the first two commits overlap with #1814. Still getting the hang of juggling multiple Git branches.)

### DIFF
--- a/frescobaldi_app/copy2image.py
+++ b/frescobaldi_app/copy2image.py
@@ -179,7 +179,8 @@ class Dialog(QDialog):
     def readSettings(self):
         s = QSettings()
         s.beginGroup('copy_image')
-        self.dpiCombo.setEditText(s.value("dpi", "100", str))
+        # 300 dpi is standard print resolution
+        self.dpiCombo.setEditText(s.value("dpi", "300", str))
         color = s.value("papercolor", QColor(), QColor)
         self.colorButton.setColor(color if color.isValid() else Qt.GlobalColor.white)
         self.colorCheck.setChecked(color.isValid())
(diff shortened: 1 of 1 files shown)


fedelibre:
> (And sorry, I just realized the first two commits overlap with #1814. Still getting the hang of juggling multiple Git branches.)

You can easily remove these two commits in your qt6-copy2image branch and force push:

```
git rebase -i 011621a
```

and replace `pick` with `drop` for the first two commits you want to remove.

bmjcode:
Thanks @fedelibre. For something so easy (now that I know how) it's not clear from Git's documentation how to do it. Just rebased without those commits.

frescobaldi_app/copy2image.py
```diff
@@ -179,7 +170,7 @@ def updateFileTypeUITexts(self):
         Called from translateUI() and from updateExport().
 
         """
-        filetype = self.exportTypes()[self.typeCombo.currentIndex()][1]
+        filetype = _("PNG")
```
jeanas, line 173: Nit: Just inline the word “PNG” into the translatable strings?

jeanas (approved):

frescobaldi_app/copy2image.py
bmjcode: Sure. How do I do that?

bmjcode: Ah, that does look a bit cleaner doesn't it.

After pondering the distinction between plain "Copy/Drag" and "Copy/Drag File" I revised the wording further to make it clearer. I'm not sure having them as separate options is actually helpful, to be honest. The plain versions are fine for embedding in a document, while "Save As" offers more control if a file is needed.

jeanas (approved):

jeanas:
I tested it locally, it looks fine.

jeanas:
(And re-reviewed, code also looks fine.)

PR #1818, opened 2024-11-21
Prefer remote docs for the user's default LilyPond version

bmjcode:
This helps when users are running an older version of LilyPond that doesn't include local documentation.

### DIFF
--- a/frescobaldi_app/appinfo.py
+++ b/frescobaldi_app/appinfo.py
@@ -47,5 +47,6 @@ required_python_version = (3, 8)
 required_python_ly_version = (0, 9, 4)
 
 # LilyPond doc URLs to be used in lilydoc/manager.py
+# see also lilypondinfo.LilyPondInfo.lilydoc_url()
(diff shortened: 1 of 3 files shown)


fedelibre:
By default, if no path is added in the LilyPond documentation preference, we see the remote docs of the latest stable and unstable. This patch will change it to the user's default version, so it's helpful in case the user has an older version.

LGTM

fedelibre:
I made a test and it's not working here:

immagine

bmjcode:
What isn't working? LilyPond's online docs are organized by the first two version numbers only (e.g. 2.25). So if 2.25.22 is the latest release in that series, that's what I'd expect to see for either 2.25.11 or 2.25.19.

fedelibre:
You are right! And I confirm that the code works fine:

image

fedelibre:
@bmjcode Can I merge this PR?

bmjcode:
@fedelibre Any time you're ready. I was waiting on you.

PR #1826, opened 2024-11-25
Use urllib to unquote textedit URLs

bmjcode:
This fixes a bug reported in frescobaldi/qpageview#30 where point-and-click fails when the filename contains both punctuation and non-latin1 characters.

### DIFF
--- a/frescobaldi_app/textedit.py
+++ b/frescobaldi_app/textedit.py
@@ -25,8 +25,9 @@ Parse textedit:// urls.
 import re
 import sys
 import collections
+import urllib.parse
 
-__all__ = ['link', 'percent_decode']
+__all__ = ['link']
 
 
 textedit_match = re.compile(r"^textedit://(.*?):(\d+):(\d+)(?::\d+)$").match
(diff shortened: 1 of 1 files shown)


jeanas:
LilyPond's conversion routine is

```c++
static bool
is_not_escape_character (char c)
{
  if (('a' <= c) && (c <= 'z'))
    return true;

  if (('A' <= c) && (c <= 'Z'))
    return true;

  if (('0' <= c) && (c <= '9'))
    return true;

  switch (c)
    {
    case '-':
    case '.':
    case '/':
    case ':':
    case '_':
      return true;
    }

  return false;
}

std::string
String_convert::percent_encode (const std::string &orig_str)
{
  std::string new_str;
  for (const auto &cur : orig_str)
    {
      if (is_not_escape_character (cur))
        new_str += cur;
      else
        {
          new_str += '%';
          new_str += String_convert::bin2hex (cur);
        }
    }
  return new_str;
}
```

which looks incomplete (https://en.wikipedia.org/wiki/Percent-encoding). So while this is more correct, it might be worth reporting a bug there?

jeanas (approved):

Issue #473, opened 2014-09-10
Frescobaldi becomes very slow on large files: use background thread for music related parsing/computation while editing


wbsoft:
Heavy computations that are done often should move to a background thread.

The ly.music module is used after every document change to parse the whole (tokenized) file and provide information such as the length of a selected music fragment or the time position of the text cursor in the status bar. This makes the editor feel sluggish on larger LilyPond files.

While editing a document, QSyntaxHighlighter triggers rehighlighting of blocks that drive the tokenizer, which does not work in a background thread but is fast enough. So we always have the tokens up-to-date. But building a `ly.music` tree of the tokenized text should be done in a background thread, also preferable only for the top-level expression that was changed.

Then the music events are traversed to measure time etc. This may traverse a lot of nodes and even load included files (although they are cached). So also those computations should go to the background, so that the editor does not lag behind when a user is typing quickly.

uliska:
I just started to read about threading in Python and learned that this is not always efficient because Python itself runs in a single thread. So for at least some of our time-consuming repeated operations it may not deliver as significant improvements as desired. (From a recent experience I can confirm that the slow response can be a showstopper for potential users and use cases.)

Would it be an option to start up a server in a separate process to handle that kind of analysis tasks?

uliska:
With the new JobQueue code in place it might be an incentive to do parsing for both highlighting and autocompletion in a background *process* rather than a *thread*. It's `done` signal would then update a data structure which would simply be somewhat out-of-date before that.

Also, when the parsing is triggered by a change in the open document it should by default *not* follow the includes (except if the change adds a new include file).

gilbertohasnofb:
I know this is an old issue but the problem seems to still be valid. I am currently editing a very large file (around 10k lines) and Frescobaldi lags for a couple of seconds whenever I type a `-` or a `\`, probably due to the autocompletion issues raised before. If I directly paste something such as `--` for a tenuto or `\shape` for tweaking slurs, Frescobaldi's response is snappy.

ksnortum:
The lag problem seems to have been exacerbated by the PyQt6 port.

bmjcode:
@ksnortum Does 5cf20d56f95ed03703e54c1ad666053490162f7b help any? I know the time position's not the main culprit but it was a relatively easy first step.

ksnortum:
Yes, it seems to help a lot, just based on my perfunctory testing.

ksnortum:
[...]

PR #1829, opened 2024-11-29
Update the music position in a background thread

bmjcode:
This was suggested in #473 because calculating this position is a slow yet frequent operation, so running it in the main thread causes lag when editing larger files.

### DIFF
--- a/frescobaldi_app/app.py
+++ b/frescobaldi_app/app.py
@@ -296,6 +296,25 @@ def job_queue():
         _job_queue = job.queue.GlobalJobQueue()
     return _job_queue
 
+
+_worker_thread = None
+
+def _stop_worker_thread():
+    if _worker_thread:
+        _worker_thread.quit()
+        _worker_thread.wait()
+
+def worker_thread():
+    """Return the global QThread object for background operations."""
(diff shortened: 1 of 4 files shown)


bmjcode:
@ksnortum mentioned in #473 that this helped with the Qt 6 port's lag. It's been a while since I've worked with QThread, so my initial attempt to implement this was a bit sloppy. I'm much happier with the revised version in 3159404988b0096d0270ade32121ef9964694606.

I'd like to have a couple more folks test this before we look at merging it. I think I've accounted correctly for thread safety, but if I'm mistaken then this will very likely cause Frescobaldi to crash horribly without warning. And that would be a very bad thing.

bmjcode:
I'm closing this because I almost have autocomplete threaded too. I'll open a new PR when that is ready.

bmjcode:
Now that I have this working, I'm redoing the branch to get a clean commit history.

bmjcode:
Sorry for the noise. Thread safety is hard. :)

fedelibre:
This has been replaced by #1836, right?

bmjcode:
Yes.

PR #1810, opened 2024-11-16
Score Wizard improvements, Qt 6 edition

bmjcode:
This PR includes several Score Wizard improvements that I'd postponed to focus on finishing the Qt 6 port. Most notably, it reworks the handling of transposing instruments as discussed under #1733 and #1800.

It also adds a couple new instrument types, fixes a few bugs, and standardizes spelling and punctuation of instrument names.

Issue #1800, opened 2024-10-08
Using the wizard with transposing instruments not coded correctly


dontbejeff:
When I was trying out the wizard to create scores, I noticed a few things. 

First, when you put in a key, say E flat major, all parts use the \global block, which is to be expected. However, when printing out the music, all parts show E flat major, which would be incorrect. Instruments in B flat should show F major as the key signature, F instruments should show with B flat major, and E flat instruments should show as C major. I do know it transposes as when you play the midi output, it transposed all the notes, which at that point claashes as it is not the note that you are expecting to hear.

To fix this, I had to go to the section where the parts are declared to put in  the score block, and before the part to be scored was called(ex: \trumpetBb) I would do the appropriate \transpose (in this case, \transpose bf c \trumpetBb as I turn on \english as the language.)

In this, the wizard should transpose the music in the appropriate place so as to correctly transpose the music as expected.

Here is a sample of the text that I have done for this

\version "2.24.3"
\language "english"

\header {
  title = "Test Quintet"
  subtitle = "Number 2"
}

scoreAGlobal = {
  \key ef \major
  \time 4/4
  \tempo "Fast" 4=120
}

trumpetBbI = \relative c'' {
  \scoreAGlobal
  \transposition bf
  % Music follows here.
  ef1
  
}

trumpetBbII = \relative c'' {
  \scoreAGlobal
  \transposition bf
  % Music follows here.
  ef1
  
}

hornF = \relative c'' {
  \scoreAGlobal
  \transposition f
  % Music follows here.
  ef1
  
}

trombone = \relative c {
  \scoreAGlobal
  % Music follows here.
  ef1
  
}

bassTuba = \relative c, {
  \scoreAGlobal
  \transposition c,
  % Music follows here.
  ef1
  
}

trumpetBbIPart = \new Staff \with {
  instrumentName = "Trumpet in Bb I"
  shortInstrumentName = "Tr.Bb. I"
  midiInstrument = "trumpet"
} \transpose bf c \trumpetBbI

trumpetBbIIPart = \new Staff \with {
  instrumentName = "Trumpet in Bb II"
  shortInstrumentName = "Tr.Bb. II"
  midiInstrument = "trumpet"
} \transpose bf c \trumpetBbII

hornFPart = \new Staff \with {
  instrumentName = "Horn in F"
  shortInstrumentName = "Hn.F."
  midiInstrument = "french horn"
} \transpose f c \hornF

trombonePart = \new Staff \with {
[...]

PR #1830, opened 2024-12-01
Add an option to enter music at either written or sounding pitch

bmjcode:
This builds on my previous work in PR #1810, which standardizes handling of transposing instruments (see #1733 and #1800 for recent discussions).

Beware there are probably bugs in the current implementation. I have done some basic testing, but have not tried every possible combination of instrument and key signature. The code is also somewhat messier than I'd like, in part because entry at written pitch requires more things to know about the key signature and transposition settings.

### DIFF
--- a/frescobaldi_app/scorewiz/build.py
+++ b/frescobaldi_app/scorewiz/build.py
@@ -23,7 +23,6 @@ Builds the LilyPond score from the settings in the Score Wizard.
 
 
 import collections
-import fractions
 import re
 
 import ly.dom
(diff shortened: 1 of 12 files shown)


bmjcode:
What's under #1810 is finished. The additional stuff here is work in progress that may or may not be ready for a Christmas release (hence my marking it as a draft).

bmjcode:
I'm stopping work on this in favor of #1810, which includes everything except the written pitch logic. I'll leave the PR open for now in case anyone else wants to take up the challenge.

Known issues right now:

* Tab and other non-standard staffs are broken in written pitch mode.
* Written-pitch transpositions may be incorrect in some keys (I've never fully understood how python-ly does transposition).
* The code in general is very messy (as you'll see in the diff for 7308b76a879a4ed9f8efc04917711f66ecf41074) because of the need for individual parts to know about the key signature and transposition settings.

bmjcode:
This is going to be a bigger mess than I have time for right now. If there's demand for it maybe we can revisit it in a future version.

PR #1836, opened 2024-12-14
Move slow operations to a background thread

bmjcode:
This will either fix the infamous editor lag if I did it right or crash horribly if I didn't.

#473 identifies two slow, repeated operations that cause lag when editing large files: Updating the music position and building a model for autocompletion. While this lag is an old problem, it has gotten significantly worse under the Qt 6 port. This PR fixes the lag by moving the heavy processing for those operations into a separate thread as suggested in that issue. The threading is managed through a new worker API that can also be used for other slow operations if needed.

Now, about those horrible crashes. I believe my current implementation is robust, but I'd like to have a few more people test it before we consider committing. Threading introduces a lot of complexity, and I'll confess my early drafts of this did include some frankly silly thread safety mistakes. Maybe I shouldn't say that too loudly before I drive potential testers away. :)

### DIFF
--- a/frescobaldi_app/app.py
+++ b/frescobaldi_app/app.py
@@ -296,6 +296,25 @@ def job_queue():
         _job_queue = job.queue.GlobalJobQueue()
     return _job_queue
 
+
+_worker_thread = None
+
+def _stop_worker_thread():
+    if _worker_thread:
+        _worker_thread.quit()
+        _worker_thread.wait()
+
+def worker_thread():
+    """Return the global QThread object for background operations."""
+    global _worker_thread
+    if _worker_thread is None:
+        from PyQt6.QtCore import QThread
+        _worker_thread = QThread()
+        aboutToQuit.connect(_stop_worker_thread)
+        _worker_thread.start()
+    return _worker_thread
+
+
 _is_git_controlled = None
 
 def is_git_controlled():
--- a/frescobaldi_app/autocomplete/completer.py
+++ b/frescobaldi_app/autocomplete/completer.py
@@ -30,6 +30,7 @@ import app
 import textformats
 import widgets.completer
 import worker
+import locking
 
 
 class Completer(widgets.completer.Completer):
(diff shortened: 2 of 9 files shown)


ksnortum:
After using the auto-complete feature several times successfully, now after a few keystrokes the auto-complete window appears in the upper left-hand corner instead of where it should be.  I don't know what the steps were to start this, but it happens all the time now, even after a restart of Frescobaldi or start a new session.

ksnortum:
It seems to be after four key-strokes, the auto-complete window moves to the upper left-hand corner.

bmjcode:
This is proving to be all kinds of messy. I'm going to try one more thing then probably walk away from this for a bit.

bmjcode:
Oops. I guess GitHub doesn't like me renaming branches with open PRs.

bmjcode:
The lag was bothering me on one of my projects, so here I am messing with this again. Current status:

Item | Status | Notes
---- | ------ | -----
Worker API | Finished | I'm finally happy with this. Look how easy it is to use!
Thread safety | In progress | There's a locking API now to help with this, but the tricky part will be using it consistently. This especially applies to our `document` object since it's used *everywhere* in the program.
Music position | Working | Based on my testing this seems to be what reduces lag the most.
Autocomplete | Partly working | There's a lot of glitches, and I'm not sure it actually helps much with the lag.

ksnortum:
I have been testing this PR and it seems to be working pretty well.  The lag is less and the problems with autocomplete seem to be gone.

ksnortum:
Lag is back for me at about 2000 lines.

bmjcode:
This is honestly looking to be more trouble than it's worth. The biggest problem I see is locking the `document` object to prevent race conditions, which is difficult because *everything* uses it, including Qt internals that aren't aware of our locking needs. I've had it crash on me at least once so this isn't just a theoretical concern.

Threading also introduces some fun non-crashing bugs. Take a look at the autocomplete suggestions for MIDI instruments, for example -- it usually misreads the cursor position and repeats letters you've already typed.

Issue #1065, opened 2018-05-05
Windows installer erroneously requires accepting the GPL


chris-morgan:
<img src="https://user-images.githubusercontent.com/392868/39658429-ae2dc7e8-5057-11e8-846e-6d3788d88135.png" width="499" height="371" alt="Screenshot of the License Agreement page of the Windows installer, which requires accepting the license to continue.">

Yet clause five of the GPL says,

> You are not required to accept this License, since you have not signed it.

(If you want to modify and distribute it you need to accept it, but that’s not the case for the installer alone.)

The most common solution for this that I’ve seen in installers is to make it informational only: “Next >” button becomes “Continue”, the accept/reject radio buttons disappear, and the explanatory text above changes to say something like “The License Agreement is provided for your information only.”

chris-morgan:
Also for comparison, the Lilypond installer:

<img src=https://user-images.githubusercontent.com/392868/39658524-da56cd8c-5058-11e8-98bd-8d80455255a5.png width=422 height=290 alt="Lilypond installer license page: clear and correct">

fedelibre:
The new .msi file built by Briefcase doesn't even show the GPL license.
That's fine for me. Let's close this issue.

PR #1854, opened 2025-01-10
Windows installer using Pynsist/NSIS

fedelibre:
I've tested the .exe file on Windows 10 and it seems to work fine.

By default pynsist doesn't add the application to the desktop or the application bar. The installation ends saying that the installation was successful, but nothing happens. We'll have to explain the user to search Frescobaldi in the main menu, then right-click on the icon and add it to the start menu or the application bar. Not a big deal IMO.

The installation might be customized by changing the NSIS template. It's the first time I work on this stuff and I prefer to keep it simple if possible.

The license is not shown right now. If `license_file` is added to the config file, an extra step before installation will check the user’s agreement to abide by the displayed license. But in #1065 it's been suggested that the user is not required to accept the license. I would simply omit this step. The LICENSE file will be copied anyway to the installation directory.

Fingers crossed for the GitHub action :-)


bmjcode:
I'll take a look this weekend. I've worked with NSIS on other projects and might be able to improve this some.

fedelibre:
Great, feel free to add your commits to the `pynsist` branch.

bmjcode:
I just tried this. To be blunt I think relying on pynsist will cause more problems than it solves. The biggest issue is there's no way I can see to customize the NSIS script, which we need to do to set up file associations among other things.

On my own projects I've used [PyInstaller](https://pyinstaller.org/) to build the executables and written my own NSIS scripts for the installers. I'm looking to see how easily we can do this with Frescobaldi.

windows-installer.cfg
```diff
@@ -0,0 +1,29 @@
...
+# pynsist doesn't handle wheel dependencies
+pypi_wheels =
+     PyQt6==6.8.0
+     PyQt6-Qt6==6.8.1
+     PyQt6-sip==13.9.1
+     PyQt6-WebEngine==6.8.0
+     PyQt6-WebEngine-Qt6==6.8.1
+     python-ly==0.9.9
+     qpageview==1.0.0
+     pygame==2.6.1
```
bmjcode, line 22: I added `pygame` for MIDI support.

windows-installer.cfg
```diff
@@ -0,0 +1,29 @@
...
+
+# Needed only for unpublished versions.
+#local_wheels =
+#     ../qpageview/dist/*.whl
+
+[Build]
+nsi_template=windows/frescobaldi.nsi
```
bmjcode, line 29: This lets us customize the install script. Among other things this is necessary to register file associations.

windows/frescobaldi.nsi
```diff
@@ -0,0 +1,287 @@
...
+    CreateDirectory "$SMPROGRAMS\${PRODUCT_NAME}"
+    [% for scname, sc in ib.shortcuts.items() %]
+    CreateShortCut "$SMPROGRAMS\${PRODUCT_NAME}\[[scname]].lnk" "[[sc['target'] ]]" \
+      '[[ sc['parameters'] ]]' "$INSTDIR\[[ sc['icon'] ]]"
+    [% endfor %]
+  [% endif %]
+  SetOutPath "$INSTDIR"
+  [% endblock install_shortcuts %]
+
+  [% block install_associations %]
```
bmjcode, line 134: This and the corresponding uninstall block are my only additions to `pynsist`'s default script. It's a bit ugly, but it works.

bmjcode (approved):
This works for me if you're OK with my additions.

fedelibre:
Ok, let me test the new installer in the coming days.

I would also like to fix the GitHub workflow file.

fedelibre:
I've tested the latest installer and it works fine. The .ly files are associated with Frescobaldi.

I'll merge it tonight unless there are other comments.

fedelibre:
@bmjcode Do you know the legal requirements for executables? Is it Ok to omit the license in the executable? (there's no license in the installed files)

pynsist offers a config option to show the license during the installation; if you put it then the user will have to accept the GPL license to proceed with the installation. But in #1065 it's been suggested that the user should not be required to accept the license.

fedelibre:
Closing in favour of #1859 (briefcase)

PR #1861, opened 2025-01-20
Rename frescobaldi_app directory as frescobaldi

fedelibre:
@jeanas This is now needed to make briefcase work. See #1859.

Please review this PR.


CONTRIBUTING.md
```diff
@@ -4,9 +4,9 @@ CONTRIBUTING
 The Frescobaldi LilyPond sheet music editor is written in Python and uses
 Qt for its user interface, via the PyQt bindings.
 
-'frescobaldi_app' is not a real package: on startup, the absolute
-'frescobaldi_app' directory is added to sys.path and its own __path__ is
-cleared so all modules and packages inside frescobaldi_app are
+'frescobaldi' is not a real package: on startup, the absolute
+'frescobaldi' directory is added to sys.path and its own __path__ is
+cleared so all modules and packages inside frescobaldi are
```
jeanas, line 9: Re this, I wonder if the packaging tools will also require us to remove this and import stuff from `frescobaldi.*` in the normal way. (This would also help IDEs and the like.) But we can leave this for later.

jeanas (approved):
LGTM!

fedelibre:
The new way is:python -m frescobaldi

Issue #1868, opened 2025-01-29
macOS: v4.0.0 crashes when quitting


JunTamura1959:
I had the same issue when I ran Frescobaldi from the shell with "briefcase run." When I quit from Frescobaldi, the command prompt came back quickly with no error messages. The "Frescobaldi quit unexpectedly" error window was displayed a few seconds later.

Here is my environment as shown in the About Frescobaldi window.

Frescobaldi: 4.0.0
Extension API: 0.9.0
Python: 3.12.7
python-ly: 0.9.9
Qt: 6.8.1
PyQt: 6.8.0
qpageview: 1.0.0
OS: macOS-15.2-x86_64-i386-64bit
installation kind: standalone .app bundle

fedelibre:
Can you paste a screenshot of the Frescobaldi error window? Not via email please, otherwise we won't see the image.

JunTamura1959:
Here is the "Frescobaldi quit unexpectedly" window:
<img width="372" alt="Image" src="https://github.com/user-attachments/assets/9d99c348-17cd-400e-864c-ff051e37ae9d" />

Clicking the "Report..." button opens the following window:
<img width="1112" alt="Image" src="https://github.com/user-attachments/assets/7ea386ef-d156-4f3e-8453-b683847a0f93" />

Here is the contents of the Problem Report:
[20250130 Problem Report for Frrescobaldi.txt](https://github.com/user-attachments/files/18601763/20250130.Problem.Report.for.Frrescobaldi.txt)

jeanas:
That's a segfault in the macOS GUI runtime, apparently while freeing some objects (`objc_autoreleasePoolPop`), but it's hard to tell more from this backtrace.

JunTamura1959:
After updating the source code with commit 0f16d6b and rebuilding the Frescobaldi.app, the issue always seems to happen.

davidfokkema:
I don't think that commit changed anything about running and quitting the app, and leftover objects. Is Frescobaldi making use of threads that might still be running in the background? I'm not familiar with the code base, but might be responsible for any breaking changes introduced in the latest updates to packaging with Briefcase, ;-).

JunTamura1959:
I fully agree that this issue has nothing to do with the codes/scripts/config files/templates/etc. written by the Frescobaldi community. The app is usable as the error occurs after you try to quit the app. It is still annoying and will surprise ordinary users, however.

JunTamura1959:
[...]

PR #1877, opened 2025-02-15
Replace global signals with native Qt versions

bmjcode:
As discussed under #1868, I believe Qt's memory management may be interfering with our custom implementation of global signals and slots. If this is the case, it could explain the persistent crashes and "object has been deleted" errors on macOS.

This is intended to test that hypothesis by reimplementing our global signals using Qt's native signals and slots.


bmjcode:
This is a major overhaul with significant potential for breakage, so I'm going to wait for feedback before continuing further. No sense doing all this if it doesn't actually fix the problem.

fedelibre:
@bmjcode Can you rebase this on master? Then ask Jun to test this as well

bmjcode:
@fedelibre Now that I understand signals.py better I don't think this is actually a good idea. The custom signals provide features Qt's don't like slot execution priorities and the ability to work from non-QObject classes. Jun's results from #1879 also lead me to believe the problem is elsewhere after all.

Issue #1618, opened 2023-05-31
Use GitHub CI to create installers


jeanas:
As much as this project is lucky to have a couple of very dedicated contributors for building the macOS and Windows installers on every release (currently @dliessi and @19joho66), it would be great to simplify the release process by generating the installers automatically using GitHub's CI system so that we could make a release by just extracting them from the download artifacts.

Hopefully, this would also help to make the generation more reliable by always doing it in the same configuration, especially for macOS, where it has been a consistent pain point.

A substantial problem is that GitHub Actions support Homebrew on macOS, but not MacPorts, so we will have to revise the scripts for generating macOS app bundles.

fedelibre:
Fixed by #1887

PR #1887, opened 2025-03-18
Sign macOS app

### LINKED ISSUE #1584: Sign macOS application bundles with Apple Developer certificate
Signing with an Apple Developer certificate adds some security for the user and makes macOS's Gatekeeper happy, avoiding the need for the user to give special permission to run the application (see https://support.apple.com/en-us/HT202491).
However it requires to have a paid Apple Developer subscription (99$/year).

There is also the possibility of signing with a non-Apple certificate, but that's unusual and still has a cost.

See the following links for more information (thanks to @nodevnul):
* https://lapcatsoftware.com/articles/hardened-runtime-sandboxing.html
* https://eclecticlight.co/2019/01/15/code-signing-for-the-concerned-1-why/
* https://eclecticlight.co/2019/01/17/code-signing-for-the-concerned-3-signing-an-app/
* https://eclecticlight.co/2020/08/22/apple-silicon-macs-will-require-signed-code/
* https://eclecticlight.co/2020/11/03/code-signatures-3-fixing-signature-problems/

I consider this low priority until we can build a fully functional application bundle (see #1244).

See also #1576.

fedelibre:
This is what was missing for the 4.0.0 release.

Let's see if the Action will complete successfully. I guess I'll have to do some trial and error...

Fixes #1584 and #1618

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,8 @@ on:
 
 env:
   FORCE_COLOR: "1"
+  # TODO: pick the Frescobaldi version from pyproject.toml?
+  RELEASE_VERSION: "4.0.0"
 
 defaults:
   run:
@@ -34,14 +36,14 @@ jobs:
             output-format: "app"
             runs-on: "macos-latest"
             briefcase-package-args: '--identity "$CERTIFICATE_ID"'
-            filename: dist/*.dmg
+            filename: dist/Frescobaldi-$RELEASE_VERSION.dmg
 
           - target: "macOS-Intel"
             platform: "macOS"
             output-format: "app"
             runs-on: "macos-13"
             briefcase-package-args: '--identity "$CERTIFICATE_ID"'
-            filename: dist/*.dmg
+            filename: dist/Frescobaldi-$RELEASE_VERSION-Intel.dmg
 
     steps:
       - name: Checkout
@@ -126,6 +128,13 @@ jobs:
             --update --no-input --log \
             ${{ matrix.briefcase-args }} \
             ${{ matrix.briefcase-package-args }}
+
+      # Briefcase generates Frescobaldi-$RELEASE_VERSION.dmg for
+      # both macOS builds. We want to rename the Intel .dmg before
+      # uploading it to Releases.
+      - name: Rename macOS-Intel .dmg file
+        if: matrix.target == 'macOS-Intel'
+        run: mv dist/Frescobaldi-$RELEASE_VERSION.dmg dist/Frescobaldi-$RELEASE_VERSION-Intel.dmg
 
       - name: Upload Frescobaldi
         uses: actions/upload-artifact@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
           - target: "macOS"
             output-format: "app"
             runs-on: "macos-latest"
-            briefcase-package-args: "--identity '$CERTIFICATE_ID'"
+            briefcase-package-args: '--identity "$CERTIFICATE_ID"'
             filename: dist/*.dmg
 
     steps:
@@ -109,6 +109,8 @@ jobs:
             ${{ matrix.briefcase-build-args }}
 
       - name: Package Frescobaldi
+        env:
+          CERTIFICATE_ID: ${{ secrets.CERTIFICATE_ID }}
         run: |
           ${{ matrix.briefcase-package-prefix }} \
           briefcase package \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
           CERTIFICATE_PATH=$RUNNER_TEMP/build_certificate.p12
           KEYCHAIN_PATH=$RUNNER_TEMP/app-signing.keychain-db
 
-          # import certificate and provisioning profile from secrets
+          # import certificate from secrets
           echo -n "$BUILD_CERTIFICATE_BASE64" | base64 --decode -o $CERTIFICATE_PATH
 
           # create temporary keychain
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,17 +24,20 @@ jobs:
         target: ["Windows", "macOS", "macOS-Intel"]
         include:
           - target: "Windows"
+            platform: "windows"
             output-format: "app"
             runs-on: "windows-latest"
             filename: dist/*.msi
 
           - target: "macOS"
+            platform: "macOS"
             output-format: "app"
             runs-on: "macos-latest"
             briefcase-package-args: '--identity "$CERTIFICATE_ID"'
             filename: dist/*.dmg
 
           - target: "macOS-Intel"
+            platform: "macOS"
             output-format: "app"
             runs-on: "macos-13"
             briefcase-package-args: '--identity "$CERTIFICATE_ID"'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,12 +51,10 @@ jobs:
         run: |
           # create variables
           CERTIFICATE_PATH=$RUNNER_TEMP/build_certificate.p12
-          #PP_PATH=$RUNNER_TEMP/build_pp.provisionprofile
           KEYCHAIN_PATH=$RUNNER_TEMP/app-signing.keychain-db
 
           # import certificate and provisioning profile from secrets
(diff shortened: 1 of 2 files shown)


PeterBjuhr (approved):

fedelibre:
The workflow file finally run successfully! 
Many thanks to @jacques-menu and @davidfokkema

You can find the Windows and macOS packages [here](https://github.com/frescobaldi/frescobaldi/actions/runs/14071423066).

macOS package needs testing! @JunTamura1959 @marnen

ksnortum:
Downloaded and installed on Windows 11; it looks great!  One thing I noticed is that bold doesn't seem to be working -- that is, keywords that should be bold aren't.  I went into Preferences -> Fonts and clicked the bold on and off and there was no difference in the syntax highlighting.  Should I open an issue?  Does anyone else see this?

marnen:
@fedelibre: 
> macOS package needs testing! @JunTamura1959 @marnen

Thanks; this is great news! I’ll try to have a look (I haven’t been using LilyPond much lately).

JunTamura1959:
@fedelibre 
I had no problem with installation (or copying to the Application folder) and launching (no security warning), but then Frescobaldi crashed:

image

I'm running the latest macOS Sequoia 15.3.2 on a MacBookPro13,2 (MacBook Pro, 13-inch, late 2016, Four Thunderbolt 3 ports, Intel Core i5) with the help of OpenCore Legacy Patcher 2.2.0.

bmjcode:
> Downloaded and installed on Windows 11; it looks great! One thing I noticed is that bold doesn't seem to be working -- that is, keywords that should be bold aren't. I went into Preferences -> Fonts and clicked the bold on and off and there was no difference in the syntax highlighting. Should I open an issue? Does anyone else see this?

It's always been like this for me too (including back in the Qt 5 days). I've never figured out what causes it though.

JunTamura1959:
@yoshiakionishi
This sounds like the issue #1868.
Since you can run Frescobaldi on an Apple Silicon machine, the crash I observed above is likely only on an Intel Mac.

nodevnul:
> @fedelibre … but then Frescobaldi crashed:
> 
> I'm running … Intel Core i5) with the help of OpenCore Legacy Patcher 2.2.0.

This is to be expected. The crash log states:

```Log
Reason: tried: '/Applications/Frescobaldi.app/Contents/Resources/app_packages/ PyQt6/Qt6/lib/QtCore.
 framework/Versions/A/QtCore' (mach-o file, ==but is an incompatible architecture 
 (have 'arm64', need 'x86_64h' or 'x86_64==')),
```

Here note „incompatible architecure … need … x86_64…”.

You have an Intel-CPU. macOS Intel cannot run Code compiled for the Arm-Architecture. It does not provide a binary translator like e. g. Rosetta on an Arm-based Mac that would translate Intel-code.

So it appears that the package is not properly built. Afaik there are two incarnations of app packages:

1. Universal binary
2. Binary Intel only or Arm only

For Case 1. the app should not crash neither on Intel nor on Arm based hardware because the appropriate binaries for the respective platform are chosen correctly.

For Case 2. an app package Intel only would launch on Intel hardware normally, on Arm hardware it would also launch after Rosetta does its translation job. An app package for Arm only should show on Intel hardware as not executable with its app pictogram - or on trying to launch the message should be that this app cannot be run on the present system.

davidfokkema:
Briefcase supports building universal binaries for quite some time now. That _does_ depend on the availability of both architecture versions of every library that the app uses. That number has peaked now and newer versions of some libraries are dropping Intel builds. At least, my app started to crash on Intel again and that was fixed by reverting one library to an earlier version. Since it looks like Qt itself is crashing, it would be wise to check the version and see on PyPI if there are Intel builds. If not, check for earlier (or later?) versions and pin that version for the build. If you want to keep on supporting Intel hardware, that is.

yoshiakionishi:
> @yoshiakionishi
> 
> This sounds like the issue #1868.
> 
> Since you can run Frescobaldi on an Apple Silicon machine, the crash I observed above is likely only on an Intel Mac.
> 
> 
> 
> 

Thanks for the reference, @JunTamura1959. I had a feeling my issue was different from yours, though I'd forgotten that there was an issue created a month or so ago. I hope the issue you observe will be resolved. 

As for my issue, is it something that might automatically be resolved when the version is released and if I install the software using Homebrew, for example?

fedelibre:
> As for my issue, is it something that might automatically be resolved when the version is released and if I install the software using Homebrew, for example?

No, we have to understand the bug and find a fix.
The app we distribute will use the software bundled in it. Installing whatever dependency - e.g. PyQt6 - from homebrew won't help (if I understood correctly your question).

fedelibre:
> Since it looks like Qt itself is crashing, it would be wise to check the version and see on PyPI if there are Intel builds. If not, check for earlier (or later?) versions and pin that version for the build. If you want to keep on supporting Intel hardware, that is.

I don't want to pin older Qt versions for an architecture whose support is being dropped.
I look forward to newer Qt versions with bug fixes.

Tomorrow I will add some information about the supported architectures and also instructions on how to build the app with briefcase for an unsupported architecture. @JunTamura1959 had previously used briefcase successfully to build a working macOS app for the same hardware which cannot run the universal binary app built by GitHub Action.

JunTamura1959:
@fedelibre 
I can try to build the application bundle locally again. Should I git-clone the current briefcase branch?

marnen:
@fedelibre:
 
> I don't want to pin older Qt versions for an architecture whose support is being dropped. I look forward to newer Qt versions with bug fixes.
> 
> Tomorrow I will add some information about the supported architectures and also instructions on how to build the app with briefcase for an unsupported architecture. 

Just to be clear: the goal of this is to have an Intel Mac (or universal) build available, right? Intel architecture is a huge part of the user base, and support for it isn’t being dropped anytime soon.

davidfokkema:
For universal builds, the Qt version really needs to be pinned on something that supports both architectures, but if you want to have separate Intel and ARM builds, I think it should be enough to just build on Intel and just build on ARM. And _that_ can be done by specifying different runners: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories. So `macos-13` will build on an Intel runner, and `macos-latest` will build on ARM.

davidfokkema:
Unfortunately, the PyQt6 packages all seem to have universal builds on PyPI, or both arm and intel builds. So to _really_ figure out what goes wrong requires a deep dive in the build process and resulting binaries...

Or build two separate versions :/

JunTamura1959:
I was able to "briefcase create|build|package" on the current master branch (0b14cc8). The resulting Frescobaldi.app runs fine so far, except for issue #1868.

fedelibre:
The build of macOS-Intel completed successfully:
https://github.com/frescobaldi/frescobaldi/actions/runs/14117637088

I don't know what will happen with the file names when uploading the files to Releases.
I mean, briefcase will create a Frescobaldi-4.0.0.dmg file for both the universal binary .dmg and for the Intel .dmg. Can I tell briefcase to add the architecture in the .dmg file name?

JunTamura1959:
@fedelibre 
I downloaded Frescobaldi-macOS-Intel.zip and installed it on my Intel Mac. The Frescobaldi.app runs fine except for issue #1868, which is annoying but not critical, I think.

davidfokkema:
> I don't know what will happen with the file names when uploading the files to Releases. I mean, briefcase will create a Frescobaldi-4.0.0.dmg file for both the universal binary .dmg and for the Intel .dmg. Can I tell briefcase to add the architecture in the .dmg file name?

I wouldn’t know how, but you can add if-statements specific for the macos-13 build and add an extra step to just rename the .dmg to Frescobaldi-4.0.0-intel.dmg or something like that, and upload that.

fedelibre:
Yes, that's what I also thought. I'll try later

fedelibre:
The Intel file has been renamed correctly.

I would merge this now. We can slightly improve the workflow file and add briefcase instructions in the next minor release.
I want to make the release asap, maybe tomorrow.

fedelibre:
> The Intel file has been renamed correctly.

No, I was wrong. The variable is not expanded :(

https://github.com/frescobaldi/frescobaldi/actions/runs/14149804617

If I hadn't set the version, [upload-release-action](https://github.com/svenstaro/upload-release-action) would have not overwritten the file name, but we have to set the right name of the file somehow. upload-release-action has an asset_name argument, but making it work in our use case is a bit tricky.

I should have made an upload test before making the release.
I'll delete the release and try again when I find the time.

fedelibre:
Morning brought an easy fix. Workflow started now, fingers crossed...

PR #1889, opened 2025-03-20
Fix packaging step in GitHub action workflow

davidfokkema:
Fix quoting the env var and copy the secret into the workflow step environment.


davidfokkema:
Ah, I missed 836ead3, sorry. Closing now.

PR #1901, opened 2025-03-31
Add package data files to be included or excluded

### LINKED ISSUE #1898: Installing 4.0.0 from source misses installing several files
Was looking at packaging 4.0.0 for Gentoo (Linux), but installing it the same way we did 3.3.0 (using a pep517 generated wheel) misses installing non-.py files (e.g. .md, .png, .svg, .dic, .mo), which leads to e.g.
```
FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/python3.12/site-packages/frescobaldi/userguide/index.md'
```
when opening the user guide, and likely missing icons among potentially other issues.

Here's the gentoo build.log for reference if needed: [build.log.txt](https://github.com/user-attachments/files/19524593/build.log.txt)

(on a side-note, build.log shows setuptools deprecation warnings about the project.license usage -- non-issue until removal planned for 2026-02-18 though)

fedelibre:
Fixes #1898 

`tox -e mo-generate` can be called from pyproject.toml?

I've built the wheel with `python -m build` and then installed it. Languages and user manuals are available.
Please review as I don't know Python packaging best practices.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "frescobaldi"
 description = "LilyPond Music Editor"
 authors = [{name = "Wilbert Berendsen", email = "info@frescobaldi.org"}]
 maintainers = [{name = "Wilbert Berendsen", email = "info@frescobaldi.org"}]
-license = {text = "GPL"}
+license = {text = "GPL-2.0-or-later"}
 requires-python = ">=3.8"
 dependencies = ["pyqt6>=6.6", "python-ly >= 0.9.5", "qpageview>=1.0.0"]
 dynamic = ["version"]
@@ -17,7 +17,6 @@ classifiers = [
     "Environment :: Win32 (MS Windows)",
     "Environment :: X11 Applications :: Qt",
     "Intended Audience :: End Users/Desktop",
-    "License :: OSI Approved :: GNU General Public License (GPL)",
     # Natural Language :: Chinese (Hong Kong) is not yet accepted by pypi
     #"Natural Language :: Chinese (Hong Kong)",
     "Natural Language :: Chinese (Simplified)",
@@ -70,6 +69,17 @@ include = ["frescobaldi*"]
 
 [tool.setuptools.dynamic.version]
 attr = "frescobaldi.appinfo.version"
+
+# Files not added by default in the wheel but needed on runtime.
+[tool.setuptools.package-data]
+"*" = [
+  "*.dic", "*.ily", "*.js", "*.ly", "*.md", "*.mo", "*.png", "*.svg",
+  "index.theme",
+  "svgview/background.html"
+]
+
+[tool.setuptools.exclude-package-data]
+"*" = ["*~", ".git*"]
 
 [tool.briefcase]
 project_name = "Frescobaldi"


ionenwks:
> `tox -e mo-generate` can be called from pyproject.toml?

On that note, I'd be nice if we could have the same that the 3.3.0 release had, aka a release tarball with pre-generated files rather than rely on github's auto-generated archives without them. Most people would likely use that unless they're doing development making it a non-issue for them.

pyproject.toml
```diff
@@ -71,6 +70,11 @@ include = ["frescobaldi*"]
 [tool.setuptools.dynamic.version]
 attr = "frescobaldi.appinfo.version"
 
+# .mo files are not tracked in git
+# `tox -e mo-generate` needs to be run before `python -m build`
+[tool.setuptools.package-data]
+"*" = ["*.mo"]
```
ionenwks, line 76: Given that in #1898 we do not install from a full git checkout (using the tarball that we mirror and users verify with a manifest, git is inconvenient there), seems this currently (only) installs the .mo and still skips all the other files.

fedelibre:
> On that note, it'd be nice if we could have the same that the 3.3.0 release had, aka a release tarball with pre-generated files rather than rely on github's auto-generated archives without them

Ok, I will try this option and add a tarball to the 4.0.0 release.

I'm having problems also with the flatpak package for the same reason, see [here](https://github.com/flathub/org.frescobaldi.Frescobaldi/pull/69).

fedelibre:
Please try this tarball:
https://github.com/frescobaldi/frescobaldi/releases/download/v4.0.0/Frescobaldi-4.0.0.tar.gz

ionenwks:
> Please try this tarball: https://github.com/frescobaldi/frescobaldi/releases/download/v4.0.0/Frescobaldi-4.0.0.tar.gz

Thanks, it's usable but it seems a bit messy. The tarball contains another broken .tar.gz and unpacks without a subdirectory which isn't typical of source tarballs (typically would have frescobaldi-4.0.0/ subdir or so named the same as the tarball itself)

fedelibre:
Ok, this one should be ok:
https://github.com/frescobaldi/frescobaldi/releases/download/v4.0.0/frescobaldi-4.0.0.tar.gz

Here's how I created it:

```
git clone -b v4.0.0 git@github.com:frescobaldi/frescobaldi.git frescobaldi-4.0.0
cd frescobaldi-4.0.0
tox -e mo-generate
tox -e linux-generate
cd ..
tar --exclude='.git*' --exclude='.tox' --exclude='i18n/__pycache__' -czvf frescobaldi-4.0.0.tar.gz frescobaldi-4.0.0
```

fedelibre:
I think you still need the patches in this PR to make pip install work.
I'm about to push here another commit I've verified on the tarball.

pyproject.toml
```diff
@@ -71,6 +70,10 @@ include = ["frescobaldi*"]
 [tool.setuptools.dynamic.version]
 attr = "frescobaldi.appinfo.version"
 
+# Files not added by default in the wheel but needed on runtime.
+[tool.setuptools.package-data]
+"*" = ["*.mo", "*.md"]
```
ionenwks, line 75: Not overly familiar with what's needed, but is this really enough?

There's a lot of other files, .svg, .png, .dic, etc... Won't it be e.g. missing icons.

fedelibre: IIUC setuptools-scm should automatically add in the wheel all the files tracked in git. However, if that was true, I should add only the files generated by tox (that is, *.mo files and the .metainfo.xml and .desktop file). But as a matter of fact I have to explicitely add .md files which are tracked in git.

In my pip install test I see that .dic, .svg and .png files are in the installation directory.
But yesterday I noticed that some images were missing in Frescobaldi preferences.

It's such a mess.
And now I can't even run Frescobaldi due to some dependency trap:

```
$ frescobaldi 
  File "/home/fede/.local/bin/frescobaldi", line 5, in <module>
    from frescobaldi.__main__ import main
  File "/home/fede/.local/lib/python3.13/site-packages/frescobaldi/__main__.py", line 46, in <module>
    import app              # Instantiate global signals etc
    ^^^^^^^^^^
  File "/home/fede/.local/lib/python3.13/site-packages/frescobaldi/app.py", line 38, in <module>
    import PyQt6.QtWebEngineWidgets
ImportError: /home/fede/.local/lib/python3.13/site-packages/PyQt6/Qt6/lib/libQt6Widgets.so.6: version `Qt_6.8_PRIVATE_API' not found (required by /lib64/libQt6WebEngineWidgets.so.6)
```

fedelibre: I'm tempted to take back MANIFEST.in. I don't have time to waste on this stuff

ionenwks: > tracked in git

Was your test with the tarball? It won't know anything about git there, and including .git would be wasteful.

fedelibre: Doh, good point. Yes, I tested the tarball and not a git checkout.
I'll think about it when I have some spare time today.

fedelibre: I think it should work now. I could test it (I should have used pipx instead of pip to test the installation).

pyproject.toml
```diff
@@ -71,6 +70,18 @@ include = ["frescobaldi*"]
 [tool.setuptools.dynamic.version]
 attr = "frescobaldi.appinfo.version"
 
+# Files not added by default in the wheel but needed on runtime.
+[tool.setuptools.package-data]
+"*" = [
+  "*.dic", "*.ily", "*.js", "*.ly", "*.md", "*.mo", "*.png", "*.svg",
+  "frescobaldi.1", "LICENSE", "THANKS", "TODO",
+  "linux/org.frescobaldi.Frescobaldi.desktop",
+  "linux/org.frescobaldi.Frescobaldi.metainfo.xml"
```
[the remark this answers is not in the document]
fedelibre: You are right, thanks for catching it.
.ico and .icns are needed for Windows and macOS packaging.

Please let me know if the last commit is Ok.

pyproject.toml
```diff
@@ -71,6 +70,17 @@ include = ["frescobaldi*"]
 [tool.setuptools.dynamic.version]
 attr = "frescobaldi.appinfo.version"
 
+# Files not added by default in the wheel but needed on runtime.
+[tool.setuptools.package-data]
+"*" = [
+  "*.dic", "*.ily", "*.js", "*.ly", "*.md", "*.mo", "*.png", "*.svg",
+  "index.theme",
+  "frescobaldi/svgview/background.html"
```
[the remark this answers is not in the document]
fedelibre: Fixed, thanks

PR #1913, opened 2025-04-04
Use hatchling as build backend

jeanas:
It has much better defaults for files to include/exclude and is generally just a lot saner than setuptools. This parallels the changes

https://github.com/frescobaldi/python-ly/pull/165
https://github.com/frescobaldi/qpageview/pull/23


jeanas:
Wait, there's an issue with the MO files…

fedelibre:
> Wait, there's an issue with the MO files…

What issue?
If I generate the mo files and then run `python -m build`, I can see the mo files in the tarball.

fedelibre:
@jeanas Can you please tell use which issue you had with the MO files?

jeanas:
I re-tested today, no issues. Not sure what I did wrong last time.

Should we merge it?

Issue #1916, opened 2025-04-06
MIDI Input error


GeoffGroberg:
When using Frescobaldi to capture midi in OS X, playing a MIDI note triggers the following error:

Traceback (most recent call last):
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/midiinput/__init__.py", line 85, in analyzeevent
    self.noteevent(event.type, event.channel, event.note, event.value)
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/midiinput/__init__.py", line 100, in noteevent
    self.print_or_replace(note.output(self.widget().relativemode(), self._language))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/midiinput/elements.py", line 31, in output
    return pitch.output(language)   (('=' ly.pitch.octaveToString(self._pitch.octave)) if QApplication.keyboardModifiers() & Qt.Modifier.SHIFT else '')
                                                                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for &: 'KeyboardModifier' and 'Modifier'

Optionally describe below what you were doing:

Frescobaldi: 4.0.1
Extension API: 0.9.0
Python: 3.12.7
python-ly: 0.9.9
Qt: 6.8.2
PyQt: 6.8.1
qpageview: 1.0.0
OS: macOS-15.3-x86_64-i386-64bit
installation kind: standalone .app bundle

fedelibre:
I could reproduce it in Linux. It's an easy fix, fortunately.

PR #1919, opened 2025-04-09
Use Qt6 enums in MIDI input

fedelibre:
Fixes issue #1916

@bmjcode I found this bug while preparing a patch for the user manual. I want to add some instructions for MIDI playback and MIDI input.

I see that there are other `Qt.Modifier` uses in the code that should be probably modified? I'm not confident with Qt and programming in general...

```
git grep -n 'Qt\.Modifier' frescobaldi/
[...]
```

https://doc.qt.io/qt-6/qt.html#KeyboardModifier-enum

### DIFF
--- a/frescobaldi/midiinput/elements.py
+++ b/frescobaldi/midiinput/elements.py
@@ -28,7 +28,7 @@ class Note:
         else:
             pitch = self._pitch
         # also octavecheck if Shift is held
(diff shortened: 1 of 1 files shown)


bmjcode (approved):
Looks good to me. This specific replacement is one of my pet peeves with PyQt6. `Qt.Modifier` and `Qt.KeyboardModifier` are supposed to be interchangeable, but they are not in PyQt6 because of how it implements enum types.

PR #1920, opened 2025-04-10
Reorganize Preferences

### LINKED ISSUE #1414: Organize/Structure Preferences window
The structure of the Preferences window seems to be slightly unclear. Here are a couple of related issues:

– The list of “tabs” on the left is named inconsistently: some categories include “Settings” or “Preferences” in their name, some don’t. I think none of them should, the user knows they’re in the Preferences window. Also, it would be clearer to write “Music View” instead of just “Music”.
– The Music Viewer has its own tab, but two settings for it are also included (under their own heading) in the “Tools” tab. I think these two should be moved.
– There is a section with two settings for “Printing music” in the “Music View” tab. I’m not sure that’s where one would look for it and it’s unclear whether these also apply to Printing source. I would suggest a dedicated Printing tab, even if there’s not that much in it (for now?).

fedelibre:
Fix #1414

### DIFF
--- a/frescobaldi/preferences/__init__.py
+++ b/frescobaldi/preferences/__init__.py
@@ -43,13 +43,13 @@ def pageorder():
     yield LilyPond
     yield MusicViewers
     yield Midi
-    yield Helpers
+    yield Editor
+    yield Tools
     yield Paths
     yield Documentation
     yield Shortcuts
-    yield Editor
     yield FontsColors
-    yield Tools
+    yield Helpers
     yield Extensions
 
 
@@ -190,7 +190,7 @@ class General(PrefsItemBase):
     help = "prefs_general"
     iconName = "preferences-system"
     def translateUI(self):
-        self.setText(_("General Preferences"))
+        self.setText(_("General"))
 
     def widget(self, dlg):
         from . import general
@@ -201,7 +201,7 @@ class MusicViewers(PrefsItemBase):
     help = "prefs_musicviewers"
     iconName = "Audio-x-generic"
     def translateUI(self):
-        self.setText(_("Music Preferences"))
+        self.setText(_("Music View"))
 
     def widget(self, dlg):
         from . import musicviewers
@@ -212,7 +212,7 @@ class LilyPond(PrefsItemBase):
     help = "prefs_lilypond"
     iconName = "lilypond-run"
     def translateUI(self):
-        self.setText(_("LilyPond Preferences"))
+        self.setText(_("LilyPond"))
 
     def widget(self, dlg):
         from . import lilypond
@@ -223,7 +223,7 @@ class Midi(PrefsItemBase):
     help = "prefs_midi"
     iconName = "audio-volume-medium"
     def translateUI(self):
-        self.setText(_("MIDI Settings"))
+        self.setText(_("MIDI"))
 
     def widget(self, dlg):
         from . import midi
@@ -234,7 +234,7 @@ class Helpers(PrefsItemBase):
     help = "prefs_helpers"
     iconName = "applications-other"
     def translateUI(self):
-        self.setText(_("Helper Applications"))
+        self.setText(_("Helper Apps"))
 
     def widget(self, dlg):
         from . import helpers
@@ -256,7 +256,7 @@ class Documentation(PrefsItemBase):
     help = "prefs_lilydoc"
     iconName = "help-contents"
     def translateUI(self):
-        self.setText(_("LilyPond Documentation"))
+        self.setText(_("LilyPond Docs"))
 
     def widget(self, dlg):
         from . import documentation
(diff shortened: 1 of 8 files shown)


fedelibre:
I'm happy with the changes.

I'm keeping it as draft, as I have to remember to update the user manual. But I'd like to have some feedback first. See also the screenshots I put in issue #1414

fedelibre:
@bmjcode I've rebased it and updated the documentation.

bmjcode:
Looks good. Just a couple points:

The page list on the left is slightly too narrow to display the full labels on mine. Can it be made a hair wider?

image

Also, the manual mentions an external printing command can be set on Windows and macOS. That's not actually the case on any platform, as a quick glance through [preferences/musicviewers.py](https://github.com/frescobaldi/frescobaldi/blob/30153922eb8719a2d11b5b0c1c42ae136f98fec4/frescobaldi/preferences/musicviewers.py#L396) will confirm. (With the move to QtPdf, everything is now printed as a raster image.)

image

fedelibre:
On Linux at startup (with no previous conf file present), the preferences appear very shrinked:

prefs shrinked

then you can drag the window and the size is remembered. So it's not a big deal... but sure it could be improved.

I see a better behaviour with the left sidebar on Linux. When I shrink the window to the minimum it never cuts the preferences on the left:

prefs no cut

fedelibre:
> Also, the manual mentions an external printing command can be set on Windows and macOS. That's not actually the case on any platform, as a quick glance through [preferences/musicviewers.py](https://github.com/frescobaldi/frescobaldi/blob/30153922eb8719a2d11b5b0c1c42ae136f98fec4/frescobaldi/preferences/musicviewers.py#L396) will confirm. (With the move to QtPdf, everything is now printed as a raster image.)

So the tooltip in Preferences>Music View>Resolution saying "Set the resolution if Frescobaldi prints using raster images" is not accurate? If everything is now printed as a raster image, we'd better say something like "Set the resolution for printing (as Frescobaldi prints the music as raster image)".
By the way, the PDF generated by LilyPond is not raster, right? So I'd rather print from a PDF viewer than from Frescobaldi.

The part of the manual about printing doesn't seem accurate anymore:

> = Printing Music =
>
>The printing command is used to print a PostScript or PDF file.
>On Linux you don't need this, but on Windows and macOS you can
>provide a command to avoid that PDF documents are being printed
>using raster images, which is less optimal.

>`$pdf` gets replaced with the PDF filename, or alternatively,
>`$ps` is replaced with the PostScript filename.
>`$printer` is replaced with the printer's name to use.
>
>Uncheck the *Use Frescobaldi's print dialog* option when the
>printing command opens a print dialog itself, in which you can
>select e.g. the page range to print and the printer to use.
>If you don't use Frescobaldi's print dialog, you don't need to put
>the `$printer` variable in your command line.

>If Frescobaldi must fall back to printing using raster images,
>you can specify the number of dots per inch here.

I can't find the *Use Frescobaldi's print dialog* option in the code. Maybe it's been removed?
And I don't see where I could use that variables: `$pdf`, etc.

Here's what I see in the print dialog:

print dialog

bmjcode:
> > Also, the manual mentions an external printing command can be set on Windows and macOS. That's not actually the case on any platform, as a quick glance through [preferences/musicviewers.py](https://github.com/frescobaldi/frescobaldi/blob/30153922eb8719a2d11b5b0c1c42ae136f98fec4/frescobaldi/preferences/musicviewers.py#L396) will confirm. (With the move to QtPdf, everything is now printed as a raster image.)
> 
> So the tooltip in Preferences>Music View>Resolution saying "Set the resolution if Frescobaldi prints using raster images" is not accurate? If everything is now printed as a raster image, we'd better say something like "Set the resolution for printing (as Frescobaldi prints the music as raster image)". By the way, the PDF generated by LilyPond is not raster, right?

PDF uses vectors, but both screens and printers are raster devices, so it has to be rendered to a raster image either way. In qpageview >= 1.0.0 that happens in [qpageview.pdf.PdfRenderer.draw()](https://github.com/frescobaldi/qpageview/blob/4066284ea88b3605e73affc9aab13bb49e241258/qpageview/pdf.py#L226). The same logic is used for both kinds of devices.

Poppler provided a so-called vector backend, but all that really did was offload rasterization to Poppler itself. Since printers operate at a fixed resolution the final output was still a raster image.

Note that [QPrinter.setResolution()](https://doc.qt.io/qt-6/qprinter.html#setResolution) only "requests that the printer prints at *dpi* or as near to *dpi* as possible". So if that preference is set to 300dpi but your printer only supports 600dpi, the PDF will print at 600dpi. You might phrase that in the manual as "the closest resolution supported by your printer will be used" and "if unsure, leave this set to the default".

> So I'd rather print from a PDF viewer than from Frescobaldi.

I agree. I only ever print from Frescobaldi to test that code since I seem to be maintaining it now. :)

> The part of the manual about printing doesn't seem accurate anymore:
> 
> > ...snip...
> 
> I can't find the _Use Frescobaldi's print dialog_ option in the code. Maybe it's been removed? And I don't see where I could use that variables: `$pdf`, etc.

It must have been removed some time ago. I don't recall ever seeing it on any platform.

fedelibre:
Thanks for the detailed explanation. I've added one more commit

bmjcode (approved):
Looks good to me.

PR #1934, opened 2025-04-29
Macos app menu i18n

### LINKED ISSUE #1577: Wish: displaying Frescobaldi menu of macOS app bundle version in native language
The Frescobaldi menu (the menu between "Apple" menu and File menu) in macOS app bundle version is always displayed in English regardless of the language setting in Frescobaldi preference or that of the macOS. It would be nicer if the Frescobaldi menu is displayed, like all other menus, with the language set in the Frescobaldi preference. I understand that this is specific to macOS app bundle and might be tricky to fix.

fedelibre:
Fix #1577


fedelibre:
For the records, this is not  working.

We really need a way to enable debug mode from the GUI.

Issue #1467, opened 2022-11-11
RuntimeError: wrapped C/C++ object of type TabBar has been deleted


jeanas:
A user encountered this error while importing a MusicXML file (https://lists.gnu.org/archive/html/lilypond-user-fr/2022-11/msg00016.html).

```
 Traceback (most recent call last):
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/macosx/file_open_eventhandler.py", line 58, in eventFilter
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/macosx/file_open_eventhandler.py", line 50, in openUrl
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/mainwindow.py", line 418, in openUrl
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/mainwindow.py", line 442, in openUrls
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/app.py", line 106, in openUrl
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/document.py", line 242, in new_from_url
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/document.py", line 87, in new_from_url
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/document.py", line 252, in __init__
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/signals.py", line 191, in emit
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/signals.py", line 308, in call
File "/Applications/Frescobaldi.app/Contents/Resources/lib/python3.9/frescobaldi_app/tabbar.py", line 128, in addDocument
RuntimeError: wrapped C/C++ object of type TabBar has been deleted
```

PeterBjuhr:
Looks like a broken QT installation, or what do you think?

dliessi:
Another user report on a variation of this issue:

> Frescobaldi: 3.1.3
> Extension API: 0.9.0
> Python: 3.9.1
> python-ly: 0.9.7
> Qt: 5.11.3
> PyQt: 5.15.2
> sip: 4.19.24
> poppler: 20.12.1
[...]

PR #1935, opened 2025-04-30
Make mainwidget an instance attribute so it's not garbage-collected

bmjcode:
Adding a widget to a layout [changes its parent](https://doc.qt.io/qt-6/layout.html#tips-for-using-layouts) to the widget on which the layout is installed. Here the layout is on `self.mainwidget`, and the tab bar and view manager become its children. Thus if `mainwidget` is a local variable, the tab bar and view manager are garbage-collected when the constructor returns and `mainwidget` goes out of scope, even though they were constructed with the `MainWindow` instance as their parent.

Hopefully this finally fixes the infamous https://github.com/frescobaldi/frescobaldi/issues/1467.

### DIFF
--- a/frescobaldi/mainwindow.py
+++ b/frescobaldi/mainwindow.py
@@ -126,14 +126,15 @@ class MainWindow(QMainWindow):
 
         app.windows.append(self)
 
-        mainwidget = QWidget()
-        self.setCentralWidget(mainwidget)
+        self.mainwidget = QWidget()
+        self.setCentralWidget(self.mainwidget)
         layout = QVBoxLayout()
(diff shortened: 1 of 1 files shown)


fedelibre:
@bmjcode Feel free to merge this if ready.

I would then make a 4.0.2 release soon and close #1467 . If it's not fixed, we'll see soon new issues.
Does this fix only #1467 or maybe other similar issues?

bmjcode:
> Does this fix only #1467 or maybe other similar issues?

I think all the "wrapped C/C++ object has been deleted" errors share the same root cause, so this should fix all of them.

PR #1938, opened 2025-05-04
Add Qt version check

fedelibre:
If Qt version is below the required version (currently 6.6), the following error window will appear:

image

and the user will be forced to quit Frescobaldi.

### DIFF
--- a/frescobaldi/appinfo.py
+++ b/frescobaldi/appinfo.py
@@ -46,6 +46,7 @@ desktop_file_name = "org.frescobaldi.Frescobaldi"
 # required versions of important dependencies
 required_python_version = (3, 8)
 required_python_ly_version = (0, 9, 4)
+required_qt_version = (6, 6)
 
 # LilyPond doc URLs to be used in lilydoc/manager.py
 # see also lilypondinfo.LilyPondInfo.lilydoc_url()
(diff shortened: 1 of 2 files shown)


bmjcode (changes requested):
I think the idea here is good, but comparing the string values is an unreliable way to implement it. For example, the check would fail for a hypothetical Qt 6.10 because a string starting with `"6.1"` sorts before one starting with `"6.6"`.

I would instead use a tuple in `appinfo.py`:

```python
required_qt_version = (6, 6)
```

And write the check using `QVersionInfo` like so:

```python
from PyQt6.QtCore import QLibraryInfo, QVersionNumber
v = QLibraryInfo.version()
r = QVersionNumber(*appinfo.required_qt_version)
if v < r:
    error("Qt version is too old.",
        f"Frescobaldi is started with Qt {v.toString()} \
but requires at least version {r.toString()}.")
```

This ensures the version is processed numerically as intended.

fedelibre:
> I think the idea here is good, but comparing the string values is an unreliable way to implement it. For example, the check would fail for a hypothetical Qt 6.10 because a string starting with `"6.1"` sorts before one starting with `"6.6"`.

Ooops, you are right:

```
>>> v="6.10"
>>> v>"6.6"
False
```

I'll follow your advice, thanks.

PR #1939, opened 2025-05-05
Update install instructions for Frescobaldi 4 (Qt6)

### LINKED ISSUE #1900: Update INSTALL instructions for version 4
When I want to open an already existing .ly file in version 4.0.0, I get the following error:

  File "/home/falter/frescobaldi-repositories/frescobaldi/frescobaldi/musicview/documents.py", line 70, in documents
    self.update()
  File "/home/falter/frescobaldi-repositories/frescobaldi/frescobaldi/musicview/documents.py", line 96, in update
    doc = pagedview.loadPdf(filename)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/falter/frescobaldi-repositories/frescobaldi/frescobaldi/pagedview.py", line 259, in loadPdf
    return qpageview.loadPdf(filename, getRenderer("pdf"))
                                       ^^^^^^^^^^^^^^^^^^
  File "/home/falter/frescobaldi-repositories/frescobaldi/frescobaldi/pagedview.py", line 240, in getRenderer
    import qpageview.pdf
  File "/home/falter/.local/pipx/venvs/frescobaldi/lib/python3.12/site-packages/qpageview/pdf.py", line 34, in <module>
    from PyQt6.QtPdf import QPdfDocument, QPdfDocumentRenderOptions, QPdfLinkModel
ImportError: cannot import name 'QPdfLinkModel' from 'PyQt6.QtPdf' (/usr/lib/python3/dist-packages/PyQt6/QtPdf.abi3.so)

Beschreiben Sie, was Sie getan haben (optional):

Frescobaldi: 4.0.0
Extension API: 0.9.0
Python: 3.12.3
Git branch: master
on commit: f1f9c34d
python-ly: 0.9.4
Qt: 6.4.2
PyQt: 6.6.1
qpageview: 1.0.0
OS: Linux-6.8.0-52-generic-x86_64-with-glibc2.39 (Linux Mint 22.1)
Installation kind: distro package or installed from source

Also recompiling the file does not help. There are no LilyPond errors, the PDF is created, but there is no pdf  displayed. How can I woo Frescobaldi to display the PDF? Btw, the Manuscript viewer cannot be used as well, of course. Everything works in the flatpak...

fedelibre:
Fix #1900


fedelibre:
IMO we should add also pygame (for MIDI playback) to the list of dependencies in pyproject.toml.

I've already added PyQtWebEngine (for the LilyPond docs browser) in this PR (the issue discussed in #1900 was probably related to a wrong configuration system-wide and installing this package via pipx [seems to be the  easiest solution](https://groups.google.com/g/frescobaldi/c/g90TVt_pc0M)).

For sure MIDI playback is not an optional feature of Frescobaldi but an essential feature that most people use, so in the optional features of the INSTALL.md file I would leave only pycups.

bmjcode:
> IMO we should add also pygame (for MIDI playback) to the list of dependencies in pyproject.toml. [...] For sure MIDI playback is not an optional feature of Frescobaldi but an essential feature that most people use

Agreed.

> so in the optional features of the INSTALL.md file I would leave only pycups.

I don't think there's anything left that even uses that.

fedelibre:
> > so in the optional features of the INSTALL.md file I would leave only pycups.
> 
> I don't think there's anything left that even uses that.

It seems that qpageview still has something that uses it:

```
fede@fedora:~/src/frescobaldi/qpageview$ git grep -n 'import cups'
qpageview/cupsprinter.py:206:            import cups
qpageview/cupsprinter.py:222:        import cups
```

I'll remove only the part about PortMidi/pygame.

fedelibre:
@jeanas I would like your review before merging this. Please let me know if you can find the time. No hurry of course, I just want to know if waiting for you makes sense.

alerque (approved):

fedelibre:
> IMO we should add also pygame (for MIDI playback) to the list of dependencies in pyproject.toml.

Well, an annoying drawback is that on Linux now you cannot use PortMidi for MIDI playback. You are forced to use pygame (now pygame-ce) or the installation will fail. For example, the Fedora package [relies on PortMidi](https://src.fedoraproject.org/rpms/frescobaldi/blob/rawhide/f/frescobaldi.spec#_29).

I'm not sure it's a good idea. pygame-ce is installed in the Windows and macOS packages using the briefcase section of pyproject.toml (tool.briefcase.requires), which should be kept.
pygame-ce in project.dependencies may be useful only for installations from source code, but in this case we may ask to install it. I'm tempted to remove it, but I would be glad to hear other opinions. I don't want to do another mistake...

I've been hit by this during [this flatpak update](https://github.com/flathub/org.frescobaldi.Frescobaldi/pull/74#issuecomment-2966127927).
What Linux distro packager think of this issue? Should we remove pygame-ce from the required python deps? @alerque @limburgher 

Thanks

limburgher:
I patched Frescobaldi to use pygame instead of pygame-ce for Fedora since Fedora doesn't have pygame-ce. If pygame isn't actually needed on all platforms, the dependencies should certainly reflect that.

fedelibre:
pygame is not a "strong" dependency, as PortMidi is an alternative dependency to enable MIDI playback. pygame is just more convenient (as a one solution for multiple platforms) because it can be installed via pip.
pygame-ce is actually needed only for macOS, because pygame works fine on Linux and Windows only.

I think we'd better remove pygame as dependency, in order to let the user/packager choose if they prefer using PortMidi instead. I'll prepare a PR in the weekend.

PR #1940, opened 2025-05-05
4.0.2 release

fedelibre:
I'll make the release tomorrow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,22 @@ Click on the version to see the complete list of commits in a release.
 ## [unreleased]
 
 
+## [4.0.2] - 2025-05-06
+
+### Changed
+
+- Reorganize and extend the Preferences (#1920 and #1928).
+- Use hatchling as build backend (#1913).
+- Updated translations: Italian, Polish.
+
+### Fixed
+
+- Make MIDI input work again (#1919 and #1922).
+- Fix a garbage-collector issue (#1935 and #1942). This should prevent
+the error "RuntimeError: wrapped C/C++ object of type FOO has been deleted",
+which has been reported as duplicates several times.
+
+
 ## [4.0.1] - 2025-04-03
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


inkandpaper-app:
Can you please compile it with the attached mac icon? For better integration with the system.
[org.frescobaldi.Frescobaldi.icns.zip](https://github.com/user-attachments/files/20053010/org.frescobaldi.Frescobaldi.icns.zip)

fedelibre:
I cannot open ICNS files in Linux.

Can you please explain which changes you made?
I can see that the file "type" changed.

Current file info:

```
$ file frescobaldi/icons/org.frescobaldi.Frescobaldi.icns 
frescobaldi/icons/org.frescobaldi.Frescobaldi.icns: Mac OS X icon, 240065 bytes, "is32" type
```

Your version file info:

```
$ file org.frescobaldi.Frescobaldi.icns 
org.frescobaldi.Frescobaldi.icns: Mac OS X icon, 230605 bytes, "ic10" type
```

inkandpaper-app:
Don't know about the filetype, it is a working icns file. You can convert it here https://cloudconvert.com/. Time ago I compiled a mac version with this icon and worked well, but, if I remember correctly, I also had to replace the svg file with the attached one. Replacing both files only in mac flavour everything should work fine.
org frescobaldi Frescobaldi

fedelibre:
The SVG file appears with a black background, which is wrong.
I will update only the ICNS file.

inkandpaper-app:
Well, if you want a white background I can provide it (though for me aesthetically it is not the right choice). But it is essential to also replace the SVG (or neither): the problem is that on Mac when the app is active the icon on the status bar is replaced with the SVG file (I just checked it), so the two files should represent the same image.

inkandpaper-app:
[black.zip](https://github.com/user-attachments/files/20060638/black.zip)
[white.zip](https://github.com/user-attachments/files/20060639/white.zip)

I attached both versions (black and white). I prefer the black one but that's my personal opinion.

fedelibre:
The standard is using a **transparent** blackground (not black nor even white), which can adapt to any background

inkandpaper-app:

fedelibre:
Thanks, I'l look into this after 4.0.2 release, due today

PR #1942, opened 2025-05-06
Block further method calls if we encounter a RuntimeError

bmjcode:
This fixes instances of #1467 where a PyQt object's underlying C++ object is deleted before the Python object is garbage-collected.


fedelibre:
I've tested this branch and I confirm that I cannot reproduce it.

I'll merge it

Issue #887, opened 2017-02-07
shortcut to access: preferences and/or midi options


pedro-nonfree:
Great application, I started using recently. It have a lot of shortcuts so you can be very productive.
But I see that when I want to configure midi I have to do:
Edit / Preferences / Midi settings (3 clicks)

I would like to have a shortcut like Atom editor that with Ctrl+, you access options
I also see useful to have a button that goes to midi options from `Midi input` and `Midi player` tools.

I think these stuff is useful and easy to implement.

Alternatively / Temporarily, I would like to know if someone knows how to do that as a snippet!

Thanks!

wbsoft:
You can navigate the preferences with the keyboard. Depending on the GUI translation: activate the Edit menu, Activate <Preferences>, press the first letter of the preferences page you want to open.

In the case of English user interface: hold Alt and press E, E, M. Which will bring up the MIDI preferences. You could configure a keyboard shortcut to the Edit→Preferences action, reducing the number of keystrokes to 2.

But is might be a good idea to make the MIDI preferences more accessible, directly in the MIDI tools.

fedelibre:
I didn't know we could navigate the Preferences sections by typing the first letter of each section!

There are two ideas here:

1. We might add a default shortcut to the Preferences, since this is a very common place to go.  Easy and necessary IMO.
2. As Wilbert wrote, we might add a settings button in the MIDI tool to jump to the MIDI preferences. I don't have the need to change the MIDI settings often, as I do it once and forget it. But if others find it useful...

PR #1952, opened 2025-05-12
Add Ctrl+, shortcut to open the Preferences

fedelibre:
GNOME and Windows¹ do not have a standard keyboard shortcut to open the Preferences so there's no default shortcut in these platforms. We can force Ctrl+, which is the default on macOS but it's also very common on some popular crossplatform editors.

This is a partial fix for #887

¹ https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts

### DIFF
--- a/frescobaldi/mainwindow.py
+++ b/frescobaldi/mainwindow.py
@@ -1316,7 +1316,9 @@ class ActionCollection(actioncollection.ActionCollection):
         self.edit_find_next.setShortcuts(QKeySequence.StandardKey.FindNext)
         self.edit_find_previous.setShortcuts(QKeySequence.StandardKey.FindPrevious)
         self.edit_replace.setShortcuts(QKeySequence.StandardKey.Replace)
-        self.edit_preferences.setShortcuts(QKeySequence.StandardKey.Preferences)
+        # QKeySequence.StandardKey.Preferences currently has no defaults on Windows and GNOME.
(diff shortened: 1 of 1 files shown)


fedelibre:
I've added a comment in the code to explain why we do not use `QKeySequence.StandardKey.Preferences`.

I know that a user might set this shortcut in the Preferences, but many users are lazy and might not even know they can do it.
IMO a default keyboard shortcut to open the Preferences should always be present. And if QKeySequence.StandardKey.Preferences doesn't have a default on Windows and GNOME, then the only option is using the same shortcut for all platforms.

What do you think?

bmjcode (approved):
Looks good to me -- though I would suggest adding that Ctrl+, is a Mac shortcut in case anyone wonders why that specific combination.

fedelibre:
Ok, I will do it.

I can't find a default shortcut for Windows.

But `Ctrl+,` is also the new default in GNOME, as documented in [GNOME Human Interface Guidelines](https://developer.gnome.org/hig/reference/keyboard.html). Qt should be updated... I might open an issue in the Qt tracker.

immagine

fedelibre:
I tried improving both the comment and the commit message

fedelibre:
I've opened [QTBUG-137059](https://bugreports.qt.io/browse/QTBUG-137059)

bmjcode (approved):
Works for me.

Issue #1954, opened 2025-05-13
AttributeError: type object 'QFontComboBox' has no attribute 'FontDialogOption'


petersch973:
Traceback (most recent call last):
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/fonts/__init__.py", line 91, in document_fonts
    dlg = oldfontsdialog.DocumentFontsDialog(self.mainwindow())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/fonts/oldfontsdialog.py", line 50, in __init__
    self.typewriterCombo = QFontComboBox(fontFilters=QFontComboBox.FontDialogOption.MonospacedFonts)
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'QFontComboBox' has no attribute 'FontDialogOption'

Optionally describe below what you were doing:

Frescobaldi: 4.0.2
Extension API: 0.9.0
Python: 3.12.7
python-ly: 0.9.9
Qt: 6.9.0
PyQt: 6.9.0
qpageview: 1.0.0
OS: macOS-13.7.2-x86_64-i386-64bit
installation kind: standalone .app bundle

PR #1955, opened 2025-05-13
QFontComboBox does not have a FontDialogOption attribute

fedelibre:
Yet another overseen Qt6 change.
Fix issue #1954


bmjcode (commented):
You could write that more concisely as

```python
self.typewriterCombo = QFontComboBox(fontFilters=QFontComboBox.FontFilter.MonospacedFonts)
````

Also note the `.FontFilter` in the constant name.

fedelibre:
Ok, done

bmjcode (approved):
Looks good to me!

PR #1960, opened 2025-05-19
Set an explicit default font on macOS

bmjcode:
When I had a chance to briefly test Frescobaldi on a Mac, I noticed the editor font defaulted to the system sans-serif font. This PR explicitly sets a monospace default font on macOS to match the behavior on other platforms.

Do any of our Mac users have a preference for which font? I assume Monaco is the safest choice since it's been in every macOS version, but it looks like Apple favors other fonts in recent OS versions.

### DIFF
--- a/frescobaldi/preferences/import_export.py
+++ b/frescobaldi/preferences/import_export.py
@@ -103,7 +103,12 @@ def importTheme(filename, widget, schemeWidget):
 
     fontElt = root.find('font')
 
-    defaultfont = "Consolas" if platform.system() == "Windows" else "monospace"
+    if platform.system() == "Darwin":   # macOS
+        defaultfont = "Monaco"
+    elif platform.system() == "Windows":
+        defaultfont = "Consolas"
+    else:
+        defaultfont = "monospace"
     if fontElt.get('fontFamily') in QFontDatabase.families():
         fontFamily = fontElt.get('fontFamily')
     else:
--- a/frescobaldi/textformats.py
+++ b/frescobaldi/textformats.py
@@ -71,7 +71,12 @@ class TextFormatData:
         s.beginGroup("fontscolors/" + scheme)
 
         # load font
-        defaultfont = "Consolas" if platform.system() == "Windows" else "monospace"
+        if platform.system() == "Darwin":   # macOS
+            defaultfont = "Monaco"
+        elif platform.system() == "Windows":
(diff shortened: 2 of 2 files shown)


fedelibre:
Let's CC some macOS users who may not follow this repository or the macOS label

@inkandpaper-app @JunTamura1959 @marnen @jacques-menu

JunTamura1959:
@bmjcode @fedelibre 
I've just realized that I've been using Menlo 10-point font, but I don't remember when or why I chose this setting. Readability is quite similar. Menlo has slightly narrower vertical spacing, which allows for the display of more lines in the same vertical space, but some might think that Monaco is easier to read.

bmjcode:
Thanks @JunTamura1959. Any other Mac users have a strong preference either way?

inkandpaper-app:
Why not SF Pro? It is the default system font on all Apple platforms.

bmjcode:
@inkandpaper-app @marnen Yes, I meant the editor font. From what I've read Apple used Monaco, then Menlo, then SF Mono. I'm not a regular Mac user myself, so I don't know offhand which macOS versions we support or which versions used which fonts. I guess we could default to SF Mono and use [QFont substitutions](https://doc.qt.io/qt-6/qfont.html#insertSubstitution) to fall back on the older fonts if needed.

marnen:
@bmjcode:

> From what I've read Apple used Monaco, then Menlo, then SF Mono.

That sounds like what I remember seeing on my Mac. 

I don’t tend to use the default font in my text editors in any case, but I do think we should have a sane default.

JunTamura1959:
For some reason, I cannot find "SF Mono" in the font pull-down menu in Frescobaldi. "Monaco" and "Menlo" are there. The default editor font in my Xcode is set to "SF Mono."
I'm running Frescobaldi 4.0.3.dev1 (a8f79c40) on macOS 15.5 (w/OCLP 2.4.0) on MacBook  Pro late 2016 Intel Core i5.

jacques-menu:
I always use Menlo on my Macs, in Terminal and BBEdit for example. IIRW, it replaced Monaco at the time.
It's easy to read and exhibits indentation well.

fedelibre (approved):
I can test on Linux only, but I've checked that this doesn't break anything.

bmjcode:
Well, if there is a problem I'm sure we'll hear about it. :grin:

PR #1961, opened 2025-05-20
Set default preference values for page scaling and scrolling orientation

bmjcode:
This fixes a UI bug where no default is shown for those preferences on a clean installation.


fedelibre (approved):

fedelibre:
I did a test after removing the .conf file, just to be sure.

Before this patch there was no selected default:

Screenshot From 2025-05-20 21-58-39

and I confirm that this PR fixes it

PR #1962, opened 2025-05-21
Japanese update

JunTamura1959:
"Page layout" in the Music View section of the Preferences window was not translated.


fedelibre (commented):
I see you've added two translations and removed two translations. Is this really what you want to do?

JunTamura1959:
There is one new translation for: "Page layout"
One correction  for: "Staff group"
Two deletions for: "Output filename:", "Include octave indication"
The one correction and the two deletions are for the ones originally suggested by the Poedit, which I overlooked. I've found where in the UI the "Staff group" appears, I mean, I understand the context. I could not find where in the UI the "Output filename:" and "Include octave indication" appear, so I decided to leave them untranslated.

fedelibre:
> I could not find where in the UI the "Output filename:" and "Include octave indication" appear, so I decided to leave them untranslated.

I think that these strings are self explanatory and you don't really need to know where they appear in the UI in order to translate them.
However, for better understanding, Poedit has a `View>Show Code Occurences` menu item, which shows where the string is used in the code. This should help a little bit. For example, look at this:

Screenshot From 2025-05-26 20-30-57

The file location in the top left tells you where it appears: the score wizard and specifically the containers part. There's also a doc string which explains everything.

You can use the same method for any string.

I hope it helps. You can add more commits to this PR if you want.

bmjcode:
> I could not find where in the UI the "Output filename:" and "Include octave indication" appear, so I decided to leave them untranslated.

"Include octave indication" is used for guitar parts in the Score Wizard. It toggles between `treble_8` clef and standard clef transposed an octave down.

JunTamura1959:
@fedelibre @bmjcode Thanks for your advice. Do you have any schedule for the 4.0.3 release? If it's a few weeks away, then I would fix the above and add a few more.

fedelibre:
I don't have any plan at the moment. I would say not before the 10/15 of June.

JunTamura1959:
@fedelibre Thanks! I will translate a bit more by 10th June.

fedelibre:
@JunTamura1959 Friendly reminder: I confirm I'll make a new release next week on 9th or 10th. In case you want to complete this PR before the new release...

JunTamura1959:
@fedelibre Thank you for the reminder. I'll complete the PR by the end of Sunday (Japan time; around noon EU time).

PR #1989, opened 2025-06-05
Papioara fresco menus

papioara:
Updated and revised menus, tooltips etc.; translated all untranslated entries


fedelibre:
Replaced by #1990

PR #1990, opened 2025-06-06
i18n: Update German translation

### LINKED ISSUE #1957: German localization Frescobaldi 4
Hello all,

I have just finished reviewing my own localization from 2023 of Frescobaldi 3.3.* (that was the state of the art back then) and made quite a few changes to existing entries. The strings are more consistent now, and I have also found better translations in some places and corrected a few typos.

There are, however, quite a few entries where I am not quite sure either what exactly they mean or in what context they appear. Sometimes I only want someone's confirmation about something I either inferred or conjectured.

And I have two general questions, because Poedit keeps warning me about it:

1. Why do many entries finish with a full stop although at least grammatically the full stops do not always make sense?
2. Very few entries start with small instead of capital letters. Is this intentional or are they typos? If so: can I change that myself?

I have counted the questions I have: it's 162 sometimes easy, sometimes more complex ones. Maybe I should hand them in in packages of ten each? That would be easier to handle. I really hope for some help especially from people who have done other localizations in their own languages.

Frescobaldi is such a powerful tool that even as an advanced user there are things you never see or need, because they have nothing to do with what you do with LilyPond or Frescobaldi.

Thank you in advance

fedelibre:
Thanks to @papioara
Fix #1957

### DIFF
--- a/frescobaldi/userguide/credits.md
+++ b/frescobaldi/userguide/credits.md
@@ -115,7 +115,8 @@ The following people contributed to {appname} (in alphabetical order):
     Davide Liessi
 
 {lang_de}:
-:  !Henrik Evers,
+:  !Hajo Bäß,
+    Henrik Evers,
     Georg Hennig,
     Markus W. Kropp,
     Urs Liska,
--- a/i18n/frescobaldi/de.po
+++ b/i18n/frescobaldi/de.po
@@ -3,13 +3,13 @@
 #
 # Georg Hennig <georg.hennig@web.de>, 2009, 2010, 2011, 2014.
 # Updates Urs Liska <git@ursliska.de>, 2012, 2013.
-# Update Hajo Bäß <habae@posteo.de>, Version 3.0.0; 2023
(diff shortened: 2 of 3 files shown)


papioara:
Thank you! I have been away, hence my belated reply. So now my translation is in the master branch, right?
I have downloaded a book on Git and will study it carefully before trying to contribute again. I basically understand the Git concept, but what often has confused me are some terms which - at least for me - have not been defined and explained sufficiently clearly. Explanations often sound like being from programmers for programmers, which I am not at all.

fedelibre:
Yes, it's in master and in 4.0.3 release.

Good study of Git! :-)

papioara:
As for the small problems I have mentioned earlier already (Grace Notes and that mixed English-German error message): I will open a new issue to keep things organized.

First, however, I will install the latest Frescobaldi version on my machine and see if they are still there. BTW: the annoying bilingual context menu (when you right-click in the editor) is gone now - we talked about it when it occurred in FB 3.3.0.

PR #1993, opened 2025-06-07
Update credits.md

Pruchniakowski:
Update credits.md

Greetings,
Grzegorz

### DIFF
--- a/frescobaldi/userguide/credits.md
+++ b/frescobaldi/userguide/credits.md
@@ -129,6 +129,7 @@ The following people contributed to {appname} (in alphabetical order):
 
 {lang_pl}:
 :  !Piotr Komorowski
+:  !Grzegorz Pruchniakowski
 
 {lang_gl}:
 :  !Manuel A. Vázquez


bmjcode:
@fedelibre I think that's something specific to Frescobaldi's user guide. It looks like our formatter uses `&` to indicate the underlined letter for keyboard accelerators (which is also how it's done in the menu code), so `!` is the escape character to treat `&` and possibly others as literal punctuation. That's just a guess from a quick glance at what you linked though.

Pruchniakowski:
Please don't type yours thoughts about this file here. I translated Frescobaldi into Polish, so please add me to the list of contributors, just like other people. Please create a new issue for this problem, if necessary. Thank you.

fedelibre:
@bmjcode I pointed you to the wrong file. You should read [this](https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi/userguide/read.py#L64). And now I think it's clear: the purpose of ! is marking a string as not translatable, that is it should not be included in the PO files.

For the sake of this PR, @Pruchniakowski did it well. All the names should be marked as not translatable.

But the file credits.md contain some inconsistencies which should be fixed:

```
$ git grep 'Marnen' i18n/userguide/it.po
$ git grep 'macOS packaging' i18n/userguide/it.po
$ git grep -A3 'Homebrew formula for macOS' i18n/userguide/it.po
i18n/userguide/it.po:msgid "Homebrew formula for macOS"
i18n/userguide/it.po-msgstr "Formula Homebrew per macOS"
i18n/userguide/it.po-
i18n/userguide/it.po-#: ../../frescobaldi/userguide/credits.md:46
```

I'll open a new issue for this.

Pruchniakowski:
Thanks!

PR #2000, opened 2025-06-13
Build the tarball in the release.yml workflow

fedelibre:


fedelibre:
I've already tested this PR on my fork (excluding macOS, because the secrets to sign macOS packages should be available on the main repo only):

- https://github.com/fedelibre/frescobaldi/releases/tag/v4.0.4.dev1
- https://github.com/fedelibre/frescobaldi/actions/runs/15635665260

PR #2007, opened 2025-06-17
Remove pygame from project.dependencies

fedelibre:
Let the user/packager decide the tool for MIDI playback.

On Linux the preferred choice is probably PortMidi installed as a system package.
An alternative is pygame from PyPI, but it currently works fine on Linux and Windows only, while macOS needs pygame-ce (community fork).

Discussed here:
https://github.com/frescobaldi/frescobaldi/pull/1939#issuecomment-2966386546

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -57,17 +57,18 @@ You will also need a few packages from your operating system:
 * Git, the version control software used for Frescobaldi's source code,
 * GNU Gettext, a suite of tools for working with translations, for generating
   compiled MO catalogs.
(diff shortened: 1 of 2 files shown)


bmjcode:
Is pygame(-ce) still included when building our Briefcase packages?

fedelibre:
Yes, pygame-ce is still included in the `tool.briefcase.requires` section so Windows and macOS packages still have it.

bmjcode (approved):
This isn't really my area of expertise but it looks clear enough to me.

PR #1988, opened 2025-06-03
Add a GitHub workflow to lint code using ruff

fedelibre:
Fix #1525

PR #2010, opened 2025-06-19
Fix F401 unused import errors

fedelibre:
https://docs.astral.sh/ruff/rules/unused-import/

This commit is the result of this command:

ruff check . --select F401 --fix

but I had to undo the changes in frescobaldi/__main__.py otherwise the application would not run.

### DIFF
--- a/frescobaldi/about.py
+++ b/frescobaldi/about.py
@@ -22,12 +22,11 @@ About dialog.
 """
 
 
-from PyQt6.QtCore import QSettings, QSize, Qt, QUrl
+from PyQt6.QtCore import QSize, Qt, QUrl
 from PyQt6.QtWidgets import (
     QDialog, QDialogButtonBox, QLabel, QLayout, QTabWidget, QTextBrowser,
     QVBoxLayout, QWidget)
 
-import app
 import appinfo
 import icons
 import helpers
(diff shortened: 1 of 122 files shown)


fedelibre:
There are still other unused imports found by ruff:

```
$ ruff check . --statistics --select F401
71	F401	unused-import
Found 71 errors.
[*] 12 fixable with the `--fix` option.
```

but they are in special files like `__main__.py` and several `__init__.py` files and I guess they've been put there for a reason.
@bmjcode What do you think it's the best way to deal with them? See attached list.

[PR2010_F401_errors.txt](https://github.com/user-attachments/files/20825627/PR2010_F401_errors.txt)

bmjcode:
@fedelibre Some of the imports in `__main__.py` execute code when they are imported even though we aren't using any classes or functions from them. It's an unconventional but deliberate way to implement loose coupling between modules. I haven't looked through that whole list but I suspect a lot of "unused" imports for our own modules are the same way.

We'll need to review these case by case and add exceptions for the ones that work as I described.

fedelibre:
Ok. I would leave this task to you, as I don't feel confident with touching this stuff.
Feel free to add commits to this PR.

fedelibre:
I'm closing this PR as I don't feel confident with touching this stuff.

I've decided to skip rule F401 for the moment, see #1988

bmjcode:
I agree with skipping this rule. Reworking all our imports to be F401-compliant would be a nontrivial effort for not a lot of gain.

Issue #2008, opened 2025-06-18
Find some solutions to reduce the duplication of issue reports


fedelibre:
> @steveschow I think @fedelibre just intended that as a suggestion, not as criticism. We get a lot of duplicate bug reports, so it helps us if you check that something wasn't already reported or fixed in a newer version. But you do raise a good point; since that dialog doesn't "know" about open issues here, it tends to encourage a lot of duplicate reports. I also recognize that bug reporting etiquette isn't necessarily obvious to first-time reporters. 

 _Originally posted by @bmjcode in [#1999](https://github.com/frescobaldi/frescobaldi/issues/1999#issuecomment-2968671279)_

kledeen:
I agree entirely,  There has been email discussion of the midi input bug (on both macOS and Ubuntu) I created this morning, but I could not find any existing bug to reference.  Do all issues end up tracked in github?

ksnortum:
Emails sent to the googlegroups.com email do not create GitHub issues, but reporting a bug from inside Frescobaldi does.

kledeen:
Consistent the need to avoid duplication, would it be worthwhile to have someone ensure that issues reported on the list server be moved over?  I assume it's much easier to track progress, search for duplicated, etc. in github that on googlegroupd

ksnortum:
I'm not sure who that someone might be, but I can certainly keep my eye out for possible issues.  But this issue is about reducing the number of duplicate issues created by the "Report a bug" window in Frescobaldi and has nothing to do with the Google Group email list.

steveschow:
My suggestion is remove the button from the dialog.  The assumption by any user is that we are doing you a favor by hitting the button and we are not going to go to GitHub to look through reports to find out whether we should use the button or not.  The program crashed, and gave a dialog saying we can send you the crash log by pressing the button.  That is the most you are going to get from me or most people.  If you don't want duplicate reports then either remove the button or don't automatically create a new ticket every time the program crashes and someone hits the button, thinking they are helping.

Or find a way to gather those button reported crashes into a bucket to sort through occasionally yourself to decide if it's related to an existing bug or needs a new ticket or whatever.  Myself I will never press the button again.  Thanks for all you do but my time is also valuable for different things.

bmjcode:
[...]

PR #2011, opened 2025-06-19
New issue template

fedelibre:
As discussed in #2008, using a default template for new issues might help reducing the issue duplication.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-issue.md
+++ b/.github/ISSUE_TEMPLATE/new-issue.md
@@ -1,0 +1,27 @@
+---
+name: New Issue
+about: Create a bug report or feature request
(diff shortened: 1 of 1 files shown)


fedelibre:
I've used a single template for bug reports and feature requests, to keep it simple. But we may use two different templates, as explained [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository), if you think it's useful.

<s>I'm not sure what will happen when a new issue is filed from Frescobaldi using the following link:
https://github.com/frescobaldi/frescobaldi/issues/new

Will it skip the template or not?</s>

PR #2012, opened 2025-06-19
Fix ruff E722, no bare excepts

ksnortum:
Fix two instances of bare excepts:

    $ ruff check --select E722
    All checks passed!


bmjcode:
~I think this is fine, but since [signals.py](https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi/signals.py) is the kind of relatively low-level code that sometimes needs to bend the rules, I want to come back and look at it with fresh eyes before approving anything.~

PR #2013, opened 2025-06-20
Improve formatting of CONTRIBUTING.md

bmjcode:
This fixes some things I noticed while scrolling through earlier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,7 +111,7 @@ New features that add menu actions can be imported in `menu.py`, while building 
 
 ## Preparing patches or Pull Requests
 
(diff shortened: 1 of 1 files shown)


fedelibre (commented):
Why did you change  to one line per paragraph? Breaking a paragraph after n characters (usually 70/90) is quite common. Probably you prefer let your editor window do the breaking automatically?

The rest looks good to me.

bmjcode:
@fedelibre I don't see the point of breaking lines within paragraphs for formats like Markdown and HTML. The formatter ignores those breaks and any sane text editor can wrap long lines automatically. Breaking manually just creates unnecessary work for human editors if they need to change something mid-paragraph.

ksnortum:
Can I make one minor suggestion?  Line 144, "Change existing files as less as possible." to "Change existing files as **little** as possible."

bmjcode:
@ksnortum Good catch. Done.

PR #2015, opened 2025-06-20
Fix ruff check E713, 'not in' warnings

ksnortum:
Change `not <element> in <list>` to `<element> not in <list>`.

    $ ruff check --select E713
    All checks passed!

I was able to test all files except `frescobaldi/vbcl/__init__.py`, which has to do with extensions, I think.  I couldn't find any Frescobaldi extensions on the web to add and remove, but the change is a trivial one, so I think we're okay.

### DIFF
--- a/frescobaldi/extensions/__init__.py
+++ b/frescobaldi/extensions/__init__.py
@@ -599,7 +599,7 @@ class Extensions(QObject):
         """
 
         inactive = self.inactive_extensions()
-        active = [key for key in self._infos.keys() if not key in inactive]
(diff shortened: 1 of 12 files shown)


igneus:
>  I couldn't find any Frescobaldi extensions on the web to add and remove

@ksnortum there is the sample extension
https://github.com/frescobaldi-extensions/sample-extension
and my daily companion
https://github.com/igneus/in-adiutorium-frescobaldi

fedelibre:
@bmjcode @igneus Can you please have a look and give the approval to this PR?

PR #2018, opened 2025-06-22
delete Python 2 compatibility measures

igneus:
Since Python 3 is now required, these constitute dead execution branches.


bmjcode (approved):
Looks good to me.

PR #2019, opened 2025-06-22
vbcl: cover with unit test

igneus:
In #2015 the `frescobaldi/vbcl/__init__.py` is mentioned as hard to test (manually).

Since the module is pure functional logic, let's cover it with unit tests, refactor boldly and never be worried again.

(I picked `pytest` because that's what I use by default. I'm not opposed to switching to any other testing tool.)

### DIFF
--- a/tests/vbcl/test_vbcl.py
+++ b/tests/vbcl/test_vbcl.py
@@ -1,9 +1,58 @@
 import pytest
 
+# global function _() is required by the vbcl module
+import i18n
+i18n.install('en')
+
 from vbcl import *
+
+def test_empty_file():
+    assert parse_file('/dev/null') == {}
 
 def test_empty():
     assert parse('') == {}
 
-def test_empty_file():
-    assert parse_file('/dev/null') == {}
+def test_empty_with_defaults():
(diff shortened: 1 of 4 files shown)


fedelibre (commented):
Can you please rename the workflow and the file to describe what is meant for? `tests.yml` and `name: Tests` is fine for me.

I'm asking because there's already a release.yml file. And lint.yml will be added later, see #1988

tests/vbcl/test_vbcl.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # comments at the end of meaningful lines are not supported
+    assert parse('key: my value # not a comment') == {'key': 'my value # not a comment'}
+
+def test_one_line_value():
+    assert parse('key: my value') == {'key': 'my value'}
+
+    # empty value
+    assert parse('key:') == {}
+    assert parse('key: ') == {'key': ''}
```
igneus, line 41: inconsistency potentially worth fixing

igneus:
(Merged master in order to resolve conflicts.)

PR #2022, opened 2025-06-24
Changes to the Score Wizard to remove unused variables.

ksnortum:
The last two instances of `ruff check --select F841` have been fixed.


bmjcode (approved):
Sure, let's go ahead and do this.

PR #2028, opened 2025-06-28
Add installation instructions for non-technical Linux users

### LINKED ISSUE #2026: INSTALL.md needs instructions for non-technical Linux users
`INSTALL.md` currently has instructions for Windows and Mac, but for non-technical Linux users there is only `apt` and `flatpak`.  `apt` may be out of date and some Linux distros don't have `flatpak` installed.  There should be instructions on how to install from the compressed tar files provided in the release section.

I'd like to expand `INSTALL.md` to include these instructions with a PR.  I have one question: is it correct that `tox` is not needed in a non-development environment?  I figure pinning Frescobaldi to the dock is as good as a desktop entry and the "mo" files don't need to be generated with the compressed tar file, is that right?

ksnortum:
I added a section in INSTALL.md for non-technical Linux users who are downloading the compressed tar file in the release website.  Fixes #2026

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,10 +15,13 @@ so, installing it is the easiest route, even though it might be out of date.
 Run `sudo apt install frescobaldi` on Debian/Ubuntu or `sudo dnf install
 frescobaldi` on Fedora.
 
-You can also install Frescobaldi from Flathub using Flatpak to get a more
-up-to-date application.
+You can also install [Frescobaldi from Flathub](https://flathub.org/apps/org.frescobaldi.Frescobaldi) using Flatpak to get a more up-to-date application:
 
-https://flathub.org/apps/org.frescobaldi.Frescobaldi
+```
+flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
+flatpak install flathub org.frescobaldi.Frescobaldi
+flatpak run org.frescobaldi.Frescobaldi
+```
 
 If you previously installed Frescobaldi via your distribution, remember to
 remove it before installing it through Flatpak. Otherwise, they will coexist on
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,7 +25,64 @@ remove it before installing it through Flatpak. Otherwise, they will coexist on
 your computer; while this is not a problem per se, you could get the old
 Frescobaldi version when you launch Frescobaldi graphically.
 
+If you are *not* a developer and none of the options above are right for you, then install the compressed tar file (.tar.gz) using the instructions below.
 
+### Installing the compressed tar file
+
+It is a good idea to uninstall any previous versions before continuing.  See [Uninstalling other versions](#uninstalling-other-versions) for how to do this.
+
+#### Download, extract and move the Frescobaldi directory
+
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -25,7 +25,55 @@ remove it before installing it through Flatpak. Otherwise, they will coexist on
...
+#### Install using `pipx`
+
+At the command line, `cd` into your home directory and type
+
+    pipx install ./frescobaldi-x.y.z
+
+After a few moments, `pipx` will say Frescobaldi is installed.  If this is your first time installing using `pipx`, type the following:
+
+    pipx ensurepath
+
```
fedelibre, line 71: You forgot to install the desktop file. Maybe better entering the tar dir from the beginning?

```
cd ~/frescobaldi-x.y.z
pipx install .
pipx ensurepath

desktop-file-install --dir ~/.local/share/applications/ --set-icon $PWD/frescobaldi/icons/org.frescobaldi.Frescobaldi.svg  linux/org.frescobaldi.Frescobaldi.desktop

update-desktop-database ~/.local/share/
```

fedelibre (changes requested):

fedelibre (approved):

INSTALL.md
```diff
@@ -25,7 +25,64 @@ remove it before installing it through Flatpak. Otherwise, they will coexist on
...
+You should now have the directory `frescobaldi-x.y.z` in your download directory.  Move it to your home directory by typing
+
+    mv frescobaldi-x.y.z ~
+
+#### Install the necessary tools
+
+At this point, if you have been using GUI tools, you will need to get to the command line.  We will now install some tools needed to install Frescobaldi.  You only have to do this once.  The next release you install, you can skip this.  If you are on a Debian/Ubuntu-like distribution, type:
+
+    sudo apt install pipx gettext libportmidi0
+
```
fedelibre, line 57: Can you add this package for Ubuntu and derivatives? `libxcb-cursor0`

As [discussed here](https://lists.gnu.org/archive/html/lilypond-user/2025-06/msg00143.html), this might be needed for users of Linux Mint and maybe similar outdated distros.

I don't think this is a problem on "modern" distros like Fedora, where this package is correctly set as a dependency of the main Qt6 package:

```
$ dnf repoquery --installed --requires qt6-qtbase-gui | grep libxcb
libxcb-cursor.so.0()(64bit)
libxcb-glx.so.0()(64bit)
libxcb-icccm.so.4()(64bit)
libxcb-image.so.0()(64bit)
libxcb-keysyms.so.1()(64bit)
libxcb-randr.so.0()(64bit)
libxcb-render-util.so.0()(64bit)
libxcb-render.so.0()(64bit)
libxcb-shape.so.0()(64bit)
libxcb-shm.so.0()(64bit)
libxcb-sync.so.1()(64bit)
libxcb-xfixes.so.0()(64bit)
libxcb-xkb.so.1()(64bit)
libxcb.so.1()(64bit)
```

fedelibre: For the records, the package in Fedora is called `xcb-util-cursor`.

fedelibre:
@ksnortum I've squashed your commits into one and added one more commit to show the flatpak install commands for the lazy user.

I'll merge this tonight.

PR #2046, opened 2025-07-24
Fix Flathub quality guidelines checks

fedelibre:
https://github.com/flathub/flathub/issues/6745


fedelibre:
This is ready to be merged

PR #2050, opened 2025-07-26
Fixes frescobaldi/frescobaldi#2041

### LINKED ISSUE #2041: After closing the window and leaving the app in the background, the next time you open it the connection to the MIDI ports is lost.
https://github.com/user-attachments/assets/0cc3f9f7-1953-463f-9fcb-66a7588431f4

As can be seen from the attached video. 

Refreshing the MIDI ports from the preferences restores the connection. 
A solution, of course, would be to programmatically refresh the ports every time you start playing (although this might be excessive, since the problem only occurs when closing and reopening the app).

inkandpaper-app:
Fix #2041

### DIFF
--- a/frescobaldi/miditool/widget.py
+++ b/frescobaldi/miditool/widget.py
@@ -243,6 +243,7 @@ class Widget(QWidget):
                 self.loadSong(files.current)
 
     def loadSong(self, index):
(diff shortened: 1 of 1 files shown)


fedelibre:
The right fix is in #2053

PR #2052, opened 2025-07-26
Icon theme for macos. Fix frescobaldi/frescobaldi#1945.

### LINKED ISSUE #1945: review icons in macOS

inkandpaper-app:
I've created an icon theme based on Google Material Icons (with some modifications). I also modified the code so that the app recognizes when switching from light to dark themes (but only on macos, as I don't have the ability to test it on other platforms). The theme is activated by checking the 'system_icons' checkbox in the preferences (if unchecked, the app's icon theme will remain the traditional TangoExt). Changing icons in the preferences causes some minor and pre-existing issues (some icons don't update, or, more rarely, the app quits). Once the preference is set, the app is stable.
This change has no effect on platforms other than macOS, although both icon themes are likely usable on Windows as well.

Fix #1945


inkandpaper-app:
Just one check to do: that the ly and ily file association icon on other platforms still works.

fedelibre (changes requested):
@inkandpaper-app Thanks for this contribution, even if I find it a bit clumsy and hard to review.

You should rebase to current origin/master and remove the unnecessary commits (such as the qpageview version changes in pyproject.toml).

The commit messages are quite poor.
For example, `Add files via upload` doesn't really describe what you are doing in 26401dbe2fbd14b16c92ffbd5b43e7a10d2cbdf8

It's the most important commit where you modified some Python code for the icon lookup. It deserves a good commit message.

fedelibre:
I guess that we should also check what to do to comply with [Google Material Icons licensing](https://developers.google.com/fonts/docs/material_icons#licensing)

fedelibre:
It could be interesting to be able to select these two new icon themes - for any operating system - in the Frescobaldi preferences, under Style. It would also be easier to really test this patch.

inkandpaper-app:
continues in #2065

PR #2056, opened 2025-07-27
Set the maximum zoom for the PDF viewer to 800%

### LINKED ISSUE #2035: Zooming in on PDF viewer crashes program
On my system (Frescobaldi 4.0.3, native archlinux package) zooming in on the pdf view increases memory usage (also having a big spike during and rendering) and computation time. This means that setting zoom too high will result in Frescobaldi freezing and producing tons of RAM to eventually crash.

Just to get an idea: With one particular score Frescobaldi sits at 515M at 100%. At 214% rendering is almost instantly, and we sit at 655M. At 418% it taskes only a minimal delay, and memory sites at 822M. At 814% we are at 958M. At 1083% Redering takes maybe half a second, memory peaks to 3.2G and falls back to 1.1G. At 1745% we take maybe 1 second and a peak of 7G which falls back to 2.3G. At 2323% memoy peaks to over 12G and falls to 3.5G.

Looking at these measurements

Image

it seems reasonable to think that memory of the viewer scales roughly by the square of the scale, which makes me believe that Frescobaldi generally renders the whole score at the given scale, rather than just the current area. This of course means that once rendering is done we get very smooth scrolling, but is means scrolling is chunky and requires a lot of memory.

Now, this comes at an even bigger issue: Once you reach a certain level of zoom (~ 1000%) due to the freezing it becomes hard to adjust your zoom when zooming with mousewheel, and actually (even when doing [Ctrl]+[+]) this input event seems to be duplicated, resulting in Frescobaldi doing multiple zooming steps after each other. This does as well apply to mousewheel zooming.

Now, what this means that when zooming in (e.g. to check alignment of some things) not only the zooming experience is not nice, but also once you go over a certain level Frescobaldi starts doing multiple steps and crashes due to high RAM demand, which is clearly not ideal.

Of course what should actually happen is that the score is only rendered in say cached chunks around the current view. E.g. when I open the file in Okular and zoom in to even 10000% memory barely changes, but once I start scrolling around memory will rise to some point, at which Okular is dumping cached tiles and memory keeps constant. This leads to really fluent zooming, but more CPU for scrolling after, which is probably fine. What we can also see with Okular is that rendering is done in a separate thread, keeping zooming and scrolling totally smooth while content is rendered. Meanwhile Frescobaldi totally freezes up when rendering, including the Code pane, the menu, probably saving and stuff.

But this is probably not going to be trivial to implement, so I’d suggest to at least limit zooming level to maybe 800% to avoid crashing people’s sessions when they try to zoom in using mousewheel.

bmjcode:
This was suggested as a workaround for qpageview performance issues, but I believe it is justified anyway since I'm not aware of any practical need for higher zoom levels in Frescobaldi. I would prefer to make this change here rather than in qpageview in case other applications using the latter have more exotic needs.

Fixes #2035.


fedelibre:
You wrote (see below) that this issue is limited to the QtPdf backend and it would not be present if a python-poppler-qt6 module existed. Perhaps you may add a note in the commit log? Just in case someone one day decides to port python-poppler-qt to Qt6.


>
>>    Of course what should actually happen is that the score is only rendered in say cached chunks around the current view.
>
>This actually is how qpageview works internally. The issue is that QtPdf -- the backend library we use now -- does not currently support rendering less than a whole page at a time, so tiling is effectively bypassed when rendering PDF files.
>
>Our previous renderer used Poppler, which does support partial rendering. I believe that's what Okular uses as well. It is without question a superior library. It's also notoriously difficult to package with Python applications, and none of us could get the bindings to work under PyQt6. So we went with what was available and working when we did the port.

bmjcode:
I'll add a note if you'd like one, but the reason I implemented this is it seemed like a reasonable idea *regardless of there being an issue*. I don't see how python-poppler-qt6 existing would make gigantic zoom factors any more useful. Other applications also limit zoom factors for presumably similar reasons; for example, LibreOffice Writer's maximum is 600%.

fedelibre:
Ok, then no need to add a note. The discussion here is enough.

fedelibre (approved):

PR #2057, opened 2025-07-27
Adapt pyproject.toml to Briefcase 0.3.24

fedelibre:


fedelibre:
I need this to be able to run the release workflow.

The shorter name is to prevent this bad readability (can't see the platform on the left sidebar):
Screenshot From 2025-07-27 21-30-24

bmjcode (approved):
Looks good to me. I'm not as familiar with Briefcase as you are, but I trust that you've tested it.

PR #2060, opened 2025-07-30
MIDI input improvements

### LINKED ISSUE #2009: Midi Input not working
This has been described extensively elsewhere, but I was unable to find an actual bug submission for the issue.  Midi input is working fine for all other applications (e.g. Logic Pro), but not for Frescobaldi.  I’d be glad to help diagnose the issue, and suspect that it may be configuration-related.

I have tried restarting Frescobaldi, rebooting the machine itself. shutting off all other apps that listen to midi, etc.

Frescobaldi: 4.0.3
Extension API: 0.9.0
Python: 3.12.8
python-ly: 0.9.9
Qt: 6.9.0
PyQt: 6.9.0
qpageview: 1.0.0
OS: macOS-15.3.1-arm64-arm-64bit
installation kind: standalone .app bundle

bmjcode:
This partially fixes #2009. It also improves the overall readability and maintainability of the MIDI input code.

### DIFF
--- a/frescobaldi/midiinput/__init__.py
+++ b/frescobaldi/midiinput/__init__.py
@@ -29,10 +29,12 @@ import documentinfo
 
 from . import elements
 
+# What this does was originally undocumented. It appears intended to match
+# chord and pitch names, but not commands or variables (thanks @ksnortum)
 LY_REG_EXPR = re.compile(
     r'(?<![a-zA-Z#_^\-\\])[a-ps-zA-PS-Z]{1,3}(?![a-zA-Z])[\'\,]*'
-   '|'
-   r'(?<![<\\])<[^<>]*>(?!>)'
+    '|'
+    r'(?<![<\\])<[^<>]*>(?!>)'
 )
 
 # Event codes from the MIDI specification


inkandpaper-app:
The changes work, with one caveat: you should keep the midi prefixes in the preferences query:

```
portname = s.value("midi/input_port", midihub.default_input(), str)
pollingtime = s.value("midi/polling_time", 10, int)
```

because without doing this, the default port is always chosen.

Or (and I think this is the best option, since they seem redundant) remove them from the preferences as well (and also remove the 'player' prefix from 'player/output_port', since it only appears in one place in the code).

fedelibre:
> Or (and I think this is the best option, since they seem redundant) remove them from the preferences as well (and also remove the 'player' prefix from 'player/output_port', since it only appears in one place in the code).

I agree: the prefixes don't seem to add any value.

bmjcode:
Good catch. I removed the redundant prefixes and added some code to automatically clean up the config files.

Note the couple remaining uses of prefixes are deliberate:

```python
# This line from frescobaldi/miditool/widget.py...
p = QSettings().value("midi/output_port", midihub.default_output(), str)

# is equivalent to...
s = QSettings()
s.beginGroup("midi")
p = s.value("output_port", midihub.default_output(), str)
```

fedelibre:
> Note the couple remaining uses of prefixes are deliberate:
> 
> ```python
> # This line from frescobaldi/miditool/widget.py...
> p = QSettings().value("midi/output_port", midihub.default_output(), str)
> 
> # is equivalent to...
> s = QSettings()
> s.beginGroup("midi")
> p = s.value("output_port", midihub.default_output(), str)
> ```

Yes, I know about this equivalence. But I had the impression - I may be wrong of course - that this could not be true in macOS. Otherwise, why MIDI settings worked fine on Linux and Windows but not on macOS?
@inkandpaper-app may know this better than me.

inkandpaper-app:
I wasn't sure what was happening, but it seemed like an extra /midi/ was appearing in the preferences file without the beginGroup line. However with these changes everything seems to be working fine.

bmjcode:
> I wasn't sure what was happening, but it seemed like an extra /midi/ was appearing in the preferences file without the beginGroup line.

The extra `/midi/` was present on all platforms because the [preferences code used both `beginGroup()` and prefixes](https://github.com/frescobaldi/frescobaldi/blob/1f92c4f6518ff4dd052d2cd57690b3db304f048d/frescobaldi/preferences/midi.py#L135).

fedelibre:
I can't do a MIDI input test before Monday night. But if you tested it already and feel confident about this PR, feel free to merge it. I want this to be part of the next release.

bmjcode:
I've tested it briefly, but since this isn't a feature I regularly use I'd rather wait for more feedback before merging.

FYI, @ksnortum sent me an improved version of `LY_REG_EXPR` that I'd also like to try at some point.

fedelibre:
> I've tested it briefly, but since this isn't a feature I regularly use I'd rather wait for more feedback before merging.

I haven't had time to test it yet.

I've launched the workflow to build the Windows and macOS packages and I will ask some users who reported problems with MIDI input to test these packages.

@bmjcode As far as I can see, I can manually trigger the GitHub Action workflows only on branches pushed to the upstream repository. This time I pushed your midiinput branch to the upstream repository (and I'll have to remember to delete it once this PR is merged). As you have push access on the upstream repository, why don't you make a habit of always pushing there so we can easily run the workflows if needed as in this case? Branches are deleted once the PRs are merged so it's not going to mess around the active branches.

bmjcode:
> As you have push access on the upstream repository, why don't you make a habit of always pushing there so we can easily run the workflows if needed as in this case?

Yeah, I've been meaning to switch that over but just hadn't gotten around to it yet.

fedelibre:
> FYI, @ksnortum sent me an improved version of `LY_REG_EXPR` that I'd also like to try at some point.

@bmjcode Do you want to try it later? Can we merge this now?
The macOS user who reported the problem tested this PR and confirmed that it's been solved.

bmjcode:
@fedelibre Let's merge this now so it's working in this release. We'll save further improvements for the next one.

PR #2064, opened 2025-08-08
Change basedir() to default to the user's Documents directory

### LINKED ISSUE #2063: How to set document path for Frescobaldi? Feature? Bug? Enhancement? User error?
Win 10 64b, Frescobaldi 4.0.3, Lilypond 2.24.4

Everything works, except...

Every time I start Frescobaldi and open the file menu, it opens in Windows\System32. I have the shortcut set to start in C:\users\me\documents, which is my documents folder. The shortcut editor navigates to that folder just fine. Other shortcuts open their programs in that folder. I can't find a folder setting for it in Frescobaldi's folder options.

I can navigate to the proper folder in Frescobaldi's file open dialog, and all the temp files are deposited appropriately, as well as the final product. I can navigate to another directory, and that works as well. But every time I close and restart it, it's in System32 again.

The part I'd consider a bug is that if a folder is specified and it can't find that folder, it should open in the user's default documents folder, and not in a windows system folder. 

But if this folder IS specified, where is that setting? And/or how can I change it?

bmjcode:
Fixes #2063.

### DIFF
--- a/frescobaldi/app.py
+++ b/frescobaldi/app.py
@@ -29,7 +29,7 @@ import platform
 import importlib.util
 import weakref
 
-from PyQt6.QtCore import QObject, QSettings, Qt, QThread
+from PyQt6.QtCore import QObject, QSettings, Qt, QThread, QStandardPaths
 from PyQt6.QtWidgets import QApplication, QMenuBar
 
 ### needed for QWebEngine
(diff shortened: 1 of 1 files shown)


fedelibre:
I'm testing this patch on Linux and it's not working as expected.

First of all, the folder detection works fine. Here's a Python shell example:

```python
$ python
Python 3.13.5 (main, Jun 12 2025, 00:00:00) [GCC 15.1.1 20250521 (Red Hat 15.1.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from PyQt6.QtCore import QObject, QSettings, Qt, QThread, QStandardPaths
>>> print(QStandardPaths.writableLocation(QStandardPaths.StandardLocation.DocumentsLocation))
/home/fede/Documenti
```

However, this folder is not used as default when no basedir is set.
The behaviour is the following:

- if I launch Frescobaldi from the source tree (`python -m frescobaldi` or `hatch run frescobaldi`), the default basedir is the source tree path, in my case `/home/fede/src/frescobaldi/frescobaldi`.
- if I launch Frescobaldi from my desktop environment, the default basedir is my home, that is `/home/fede`.

fedelibre:
Wait, it does work as expected if I remove this line in my settings file:

```
[General]
basedir=
```

So the problem here is that maybe when I removed my custom default folder, instead of removing basedir from the settings file, Frescobaldi just made it empty and this doesn't play well with your patch?

bmjcode:
> So the problem here is that maybe when I removed my custom default folder, instead of removing basedir from the settings file, Frescobaldi just made it empty and this doesn't play well with your patch?

The problem is my logic was wrong. I was only handling when `basedir` was unset rather than when it was explicitly set blank. I just pushed an updated version that corrects it.

fedelibre (approved):
Now it works fine!

PR #2065, opened 2025-08-10
Icon theme dark/light for MacOS and Windows

### LINKED ISSUE #1945: review icons in macOS

inkandpaper-app:
I've created an icon theme based on Google Material Icons (with some modifications). I also modified the code so that the app recognizes when switching from light to dark themes. The theme is activated by checking the 'system_icons' checkbox in the preferences (if unchecked, the app's icon theme will remain the traditional TangoExt). Changing icons in the preferences causes some minor and pre-existing issues (some icons don't update, or, more rarely, the app quits). Once the preference is set, the app is stable. This change should work on Windows (>= 10) and MacOS (but it was tested on MacOS only). It has no effect on Linux platforms:  in this case, unlike the other two, the 'system_icons' preference makes sense, and to implement the new theme I would have had to modify the preferences mechanism more invasively (which my limited knowledge of the app prevents me from doing). However, it shouldn't be difficult for a more experienced programmer to add the theme on Linux as well. Fix #1945


inkandpaper-app:
The Apache 2.0 and GNU GPL v3 licenses are compatible, while it seems that Apache 2.0 and GNU GPL v2 are not. Maybe there's a way around this, but I'm no expert in this area.

inkandpaper-app:
A new icon theme is based on a MIT licensed theme. Pull request #2067

PR #2066, opened 2025-08-11
Modifications to fix ruff warnings, redefined while unused (F811)

ksnortum:
Modifications have been made so that the ruff linter warnings for "redefined while used" (F811) are gone.

	$ ruff check --select F811
	All checks passed!

There were two instances where the code took this form:

	def foo():
		f = Bar()
		global foo
		foo = lambda: f
		return f

I'm assuming this is a Pythonic way to create a singleton and I marked the code so that ruff would skip the checks on those.  I don't know of a way to create a singleton in Python that ruff will pass.

### DIFF
--- a/frescobaldi/debug.py
+++ b/frescobaldi/debug.py
@@ -25,24 +25,24 @@ def doc_repr(self):
 document.Document.__repr__ = doc_repr
 
 @app.documentCreated.connect
-def f_doc_created(doc):
+def doc_created(doc):
     print("created:", doc)
 
 @app.documentLoaded.connect
-def f_doc_loaded(doc):
+def doc_loaded(doc):
     print("loaded:", doc)
 
 @app.documentClosed.connect
-def f_doc_closed(doc):
+def doc_closed(doc):
     print("closed:", doc)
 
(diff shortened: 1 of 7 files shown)


bmjcode:
> There were two instances where the code took this form:
> 
> ```
> def foo():
> 	f = Bar()
> 	global foo
> 	foo = lambda: f
> 	return f
> ```
> 
> I'm assuming this is a Pythonic way to create a singleton and I marked the code so that ruff would skip the checks on those. I don't know of a way to create a singleton in Python that ruff will pass.

The form we generally use is:

```python
_foo = None
def foo():
    global _foo
    if _foo is None:
        _foo = Bar()
    return _foo
```

ksnortum:
Could someone check the debugging output for me?  I'm getting stair-stepping (line feeds without carriage returns) on some of my output.  I think it's just something in my setup, but I want to make sure I didn't cause it with my changes to `debug.py`.

bmjcode:
Debugging is working for me. You haven't changed anything that would affect the output, so it must be specific to your setup.

Issue #1199, opened 2019-10-11
Dark theme icon alternatives


uliska:
Related to #1194 

![image](https://user-images.githubusercontent.com/1812148/66663815-abc49400-ec4b-11e9-8dfd-165dff4b34f7.png)

I *think* we should provide at least one normal and one "dark" variant for the custom icons. We won't manage to make the custom icons look consistent with arbitrary available desktop themes, but as can be seen from the screenshot dark/light variants at least are crucial for usability.

Ideally we'd provide custom variants adjusted to *all* available themes, but I think this is unrealistic to expect. The management should not be very complicated, but of course we'd need someone to *design* all these icons/variants...

RyuKurisu:
I'm just getting started with Frescobaldi, and the Flatpak does automatically pickup the dark theme setting from Fedora Kinoite/Silverblue but crashes after a minute or 2.  But has there been any updates on this? 🙂

fedelibre:
@RyuKurisu can you provide more details on the crash?

fedelibre:
Nevermind, I've seen #1373

inkandpaper-app:
I've created an icon theme based on Google Material Icons (with some modifications). I also modified the code so that the app recognizes when switching from light to dark themes (but only on macos, as I don't have the ability to test it on other platforms). The theme is activated by checking the 'system_icons' checkbox in the preferences (if unchecked, the app's icon theme will remain the traditional TangoExt). Changing icons in the preferences causes some minor and pre-existing issues (some icons don't update, or, more rarely, the app quits). Once the preference is set, the app is stable.
This change has no effect on platforms other than macOS, although both icon themes are likely usable on Windows as well.

https://github.com/inkandpaper-app/frescobaldi/tree/icon_theme_for_macos

![Image](https://github.com/user-attachments/assets/8aca66a8-2456-4020-abdc-e511cfb2a8c7)

![Image](https://github.com/user-attachments/assets/718fe567-89d6-4fa9-98da-c2e3989f9779)

fedelibre:
I was quickly reviewing your PR, but I've seen the screenshots only now. Very nice look!

I would be interested to test these styles in Linux (and Windows), if you can make them available for all platforms.

fedelibre:
Can we consider this fixed by #2067?

The Tango Icon Theme project is [inactive since 17 years](https://gitlab.freedesktop.org/tango/tango-icon-theme)...
[...]

PR #2067, opened 2025-08-12
Icon theme dark/light for MacOS and Windows

### LINKED ISSUE #1945: review icons in macOS

inkandpaper-app:
I've created an icon theme based on Tabler Icons https://github.com/tabler/tabler-icons (with some modifications). I also modified the code so that the app recognizes when switching from light to dark themes. The theme is activated by checking the 'system_icons' checkbox in the preferences (if unchecked, the app's icon theme will remain the traditional TangoExt). Changing icons in the preferences causes some minor and pre-existing issues (some icons don't update, or, more rarely, the app quits). Once the preference is set, the app is stable. This change should work on Windows (>= 10) and MacOS (but it was tested on MacOS only). It has no effect on Linux platforms: in this case, unlike the other two, the 'system_icons' preference makes sense, and to implement the new theme I would have had to modify the preferences mechanism more invasively (which my limited knowledge of the app prevents me from doing). However, it shouldn't be difficult for a more experienced programmer to add the theme on Linux as well.

Fix #1945 #1199


fedelibre:
I should have time to test this PR next week

inkandpaper-app:
Ok. For testing purposes on Linux, you can comment out the two lines:

if platform.system() == "Windows" or platform.system() == "Darwin":

in /frescobaldi/__main__.py  ,  /frescobaldi/icons/__init__.py  and  /frescobaldi/icons/change_theme_eventhandler.py

fedelibre:
I'm testing this PR on WIndows and I see some issues:

1. This icon theme is the new default, if "Use System Icons" is checked (otherwise the Tango icons bundled in Frescobaldi are used). Don't know if this is what we want. A softer approach could be adding this theme as an opt-in.
2. If I change the system to dark mode, Frescobaldi does not change icon colors immediately. I have to close and restart. But I'm testing the application from the terminal (hatch run frescobaldi) and maybe if properly installed it could change immediately.
3. This is the most important: when system theme is set to dark, Frescobaldi uses the dark theme (light icons) but the Qt interface remains light so the visibility is broken. See screenshot below.

immagine

inkandpaper-app:
1. Currently, on Mac, when "Use system icons" is selected, the icons are those shown in the screenshot: not very consistent, because not all the required icons are present at the system level. You should test it on Windows, but I don't think the situation changes much. The "Use system icons" option only makes sense on Linux, where it's common for icon themes to replace all the required ones.

Screenshot 2025-08-29 alle 11 41 46

2. Even on Mac, changing the theme from dark to light or vice versa while the app is open doesn't redesign all the elements, only the main window. 




However, reopening the app restores the correct appearance:



3. This looks to me like a Qt issue with that particular version of Windows (the app window should be dark). If closing and reopening the app doesn't fix the problem, I don't think there's anything we can do.

inkandpaper-app:
If Qt isn't yet ready to enable the transition from light to dark themes on all systems, a less impactful solution could be to make the two themes optional via a drop-down menu (instead of the 'Use system icons' checkbox). I think bmjcode could do this quickly, but I prefer not to touch that part of the code, as it requires switching to a new version of the preferences.

fedelibre:
It may depend on the Qt styles. The only one that works fine *with dark themes* in Windows is the Fusion style (this has been reported by @bmjcode as well). So if I choose the Fusion style and check the "Use System Icons" option I can finally see how your dark theme looks on Windows:

immagine

inkandpaper-app:
Note that the lilypond and wizard icons aren't changed immediately, but only when the app is restarted. This is also true on Mac. A minor issue, though.

fedelibre:
Yes, that is a minor issue.

I like these icons and especially the fact that they work well with dark theme.
I think that Fusion style - being the only one that supports both light and dark themes - should be the default for Windows. What do you think?

inkandpaper-app:
I agree. Aside from the issue of supporting light and dark modes, they seem very similar to me, so I don't see the point of keeping both.

fedelibre:
I've also quickly tested it on Linux following your hints above.

A few problems:

- the music view icon of the new theme is missing
- the theme becomes the default for Linux, which is not what I would want. In general, even for Windows and macOS, I would like this theme to be optional
- if I change my system to dark mode, the dark theme is applied to Frescobaldi but it doesn't work well (and changing to the Fusion style does not help as it did on Windows)

Some screenshots. Light mode:

Screenshot From 2025-08-31 19-23-20

Dark mode:

Screenshot From 2025-08-31 19-24-17

inkandpaper-app:
1) It should be fixed now. I've always wondered what the point of leaving icons in the parent directory of themes was.
2) There are several problems on Linux: if you've noticed, some system icons are replaced: 'Aiuto, 'Annulla', 'Ripristina'...
To prevent this from happening, you should change the names of all icons by adding an app-specific prefix, and simultaneously change every icon reference within the code. It shouldn't be too complicated, but something might be missed.
3) This is a problem with your Linux distribution's theme: the window in the dark theme must be dark...

inkandpaper-app:
On Windows and Mac the default theme is actually Tango, and it's used when 'Use system icon' is unchecked. There's no system theme on Mac and Windows, so it's fine to leave this as an alternative.

inkandpaper-app:
On Linux, you can already set your preferred theme without adding any additional options to the app: just put your theme in /usr/share/icons and it will remain consistent not only with the app but also with the entire system.
I think you should keep this change only for Mac, or at most for Windows, setting Fusion as the default theme (and removing the other one).

inkandpaper-app:
With the last commit should work on linux as well. Can you test it?

fedelibre:
> With the last commit should work on linux as well. Can you test it?

Sorry for giving it up on this. (I had some important life changes in the last six months)

I confirm that the Music View icon in the preferences now is correct:

immagine

fedelibre:
> 3. This is a problem with your Linux distribution's theme: the window in the dark theme must be dark...

I've just realized that this is a problem when I run Frescobaldi in "development mode" using the PyQt6 installed via pip in a virtualenv. If I use `hatch run frescobaldi` I don't have access to some styles (like Adwaita-dark) that are available on system level.

If I have Qt6 installed at system level and start the application using `python -m frescobaldi`, then the dark theme works fine even in GNOME:

immagine

fedelibre:
I must say that the Adwaita-dark style is quite inconsistent compared to the new theme in this PR:

immagine

fedelibre (commented):
@bmjcode Can you share your thoughts on this PR? I think it's a good contribution but should be carefully checked, as it has an impact on the look of Frescobaldi.

I would prefer to add this theme but set it as optional. I don't have any expertise in this field and I hope you can have a better view of how we can manage the themes in Frescobaldi.

frescobaldi/icons/__init__.py
```diff
@@ -24,8 +24,8 @@
 
 import os
 
-from PyQt6.QtCore import QDir, QFile, QFileInfo, QSettings, QSize
-from PyQt6.QtGui import QIcon
+from PyQt6.QtCore import QDir, QFile, QFileInfo, QSettings, QSize, Qt
```
bmjcode: Do we use the `Qt` class at all in this file? I don't see any references to it in your new code.

frescobaldi/icons/__init__.py
```diff
@@ -88,7 +88,10 @@ def update_theme():
     """Change the theme for icon lookup after change of style preference"""
     s = QSettings()
     if s.value("system_icons", True, bool):
-        QIcon.setThemeName(s.value("guistyle", "", str))
+        if QGuiApplication.palette().window().color().value() > QGuiApplication.palette().windowText().color().value():
```
bmjcode, line 91: Clever 👍

frescobaldi/icons/change_theme_eventhandler.py
```diff
@@ -0,0 +1,50 @@
+# This file is part of the Frescobaldi project, http://www.frescobaldi.org/
+#
+# Copyright (c) 2013 - 2014 by Wilbert Berendsen
```
bmjcode, line 3: Is this all your code? If so, you should put your own name and the current date.

frescobaldi/__main__.py
```diff
@@ -210,8 +210,12 @@ def main(debug=False):
             splashscreen.show()
 
     # application icon
-    import icons
-    QApplication.setWindowIcon(icons.get("org.frescobaldi.Frescobaldi"))
+    if platform.system() != "Darwin":
```
bmjcode, line 213: Why is this not done on macOS? There should be a comment here to explain that.

frescobaldi/menu.py
```diff
@@ -350,7 +350,7 @@ def menu_tools_format(mainwindow):
 
 def menu_tools_transform(mainwindow):
     m = Menu(_('submenu title', "Musical &Transformations"), mainwindow)
-    m.setIcon(icons.get('Audio-x-generic'))
+    m.setIcon(icons.get('audio-x-generic'))
```
bmjcode, line 353: Does this change affect other icon themes?

bmjcode (commented):
I like what I'm seeing in the screenshots, but tbh I haven't really been keeping up with the discussion. Let me come back when I have time to look more closely at this.

For now I've highlighted a couple spots in the code where I had questions or suggestions. Overall this looks pretty good; I just want to make sure I fully understand what you're doing before making a decision on whether/when to merge.

frescobaldi/menu.py
inkandpaper-app: No, because I've renamed the file in TangoExt folder as well lowering the first letter.

frescobaldi/__main__.py
inkandpaper-app: On macOS this feature is not recommended, because it changes the launcher icon in the dock while the app is in use.

frescobaldi/icons/change_theme_eventhandler.py
inkandpaper-app: Ok

frescobaldi/icons/__init__.py
inkandpaper-app: Correct

inkandpaper-app:
Finally, you could change the text 'use system icons' to 'use symbolic icons'. But that would involve translators, I'd rather leave the change up to you.

frescobaldi/__main__.py
bmjcode: Weird, but completely in character for macOS :)

frescobaldi/icons/org.frescobaldi.Frescobaldi.icns
bmjcode, line 1: I'm guessing this was accidental, but better change it back. I don't think we want the icon file to be executable.

inkandpaper-app: Ok

fedelibre:
> Finally, you could change the text 'use system icons' to 'use symbolic icons'. But that would involve translators, I'd rather leave the change up to you.

The text *must* be changed to reflect the actual behaviour, if we agree on this change. *Use Symbolic Icons* looks good to me.
As far as I can understand from testing this branch, this is the new behaviour:

- Use Symbolic Icons unchecked => the bundled Tango theme will be used
- Use Symbolic Icons checked => the bundled Dark/Light theme will be used.
- System icons cannot be longer used

Eventually we'll have to update the user manual [here](https://github.com/frescobaldi/frescobaldi/blob/master/frescobaldi/userguide/prefs_general.md).
But this can be done after this PR has been  merged.

inkandpaper-app:
With this last commit I'm satisfied with the theme, I won't make any further changes. You can merge whenever you want.

bmjcode:
I've finally had a chance to try out this PR. Overall the new theme solves the technical problems with dark mode, but I have concerns about the usability of some of the specific icon designs.

The new/open/save/close icons are nonstandard, and only differentiated by a small symbol in the corner:

image

A user seeing those for the first time is not likely to understand what they mean compared to the more familiar folder and floppy icons. In addition, the smallness of the corner symbol makes them difficult to distinguish for users with weaker eyesight; besides being widely known, a major strength of the traditional icons is their very distinct shapes.

Similarly, it's unclear at first glance what the Score Wizard and engrave icons are meant to represent:

image

I get that the engrave icon is meant to be a stylized flower like in the LilyPond logo, but it doesn't actually *resemble* the one in that logo, straining the already tenuous connection from flower -> LilyPond -> engraving.

Less critically, some of the icons also just look odd to me, such as those for "Code Formatting" and "Pitch" below:

image

image

Please don't read this as dismissing the work you've put into this; it's quite an impressive effort. But I'm also not comfortable merging it yet because I think some of the new designs will cause too much confusion.

inkandpaper-app:
Could you check if the issues have been resolved with the latest changes? In particular, could you test the menus? (the menu icons are disabled on Mac and I'm working blind)

bmjcode:
A vast improvement! I'll have to look more closely when I have more time, but this at least resolves my previous concerns.

inkandpaper-app:
After this last upload the theme looks good to me. Let me know if you think I should make any further changes.

fedelibre:
I like the new engrave icon.

I would like the **score wizard icon** to be replaced. The current document icon with the bullet list is way too generic IMO. What about these options?

- https://tabler.io/icons?icon=wand
- https://tabler.io/icons?icon=sparkles
- https://tabler.io/icons?icon=file-spark
- https://tabler.io/icons?icon=file-music (+ sparkle) <== to be created

fedelibre (dismissed):
What I'd like to be addressed in this PR is the option in the Preferences. "Use System Icons" should be replaced by another short text and a new tooltip describing the new behaviour. Proposal:

- Use Symbolic Icons
- Uses the new minimal, high-contrast icons designed for both light and dark themes. Otherwise, falls back to the classic colored Tango icons.

I would keep this option checked by default.

Once this is merged, I can update the user guide in another PR.

bmjcode:
@fedelibre I think it would be better for the symbolic icons to replace the Tango icons completely. Having two separate icon sets just creates more maintenance work, and could potentially confuse users if what they see doesn't match what's in the documentation.

So far my month has been off to an exceptionally busy start but I will test this PR again when I have more time.

inkandpaper-app:
I updated the wizard icon in the latest upload. I am currently on vacation, so it is a bit difficult for me to make code changes. It would be better if you handled that later on—perhaps after deciding whether to keep the old icons.

fedelibre:
> @fedelibre I think it would be better for the symbolic icons to replace the Tango icons completely. Having two separate icon sets just creates more maintenance work, and could potentially confuse users if what they see doesn't match what's in the documentation.

I would set the new symbolic icons as default and keep Tango icons as fallback for a while (until we decide to remove them completely). You know, I bet some old Frescobaldi user could not live without the good old Tango icons :-)

inkandpaper-app:
As far as I'm concerned, the icon theme is fine. If there are no further comments, I won't make any more changes.

fedelibre:
Here we'll wait for a final review by @bmjcode 

Ben, please squash all the commits into one before merging. It's a rather messy commit history.

bmjcode:
Sorry I haven't been keeping up with this as closely as usual. Among other things I've been busy getting ready to move. But I will test this out as soon as I have a spare minute!

bmjcode:
This looks nice. I particularly like the new Score Wizard icon. Great work, @inkandpaper-app!

ksnortum:
I just pulled this PR in from the master branch.  Very nice!  I have two comments.<s>  One, the old icons are gone.  Is this what you want?  Because some people may want to old icons still.</s> (Strike that, if you uncheck symbolic icons, you get the old icons back.)

Second, in the tabs for the documents, there is an icon that shows if the code was successfully run.  It is now a very light green (I think) check mark.  It's too light; it's hard to see on my system (Ubuntu Linux).  

Frescobaldi-tabs 

Can we make it darker?  (I am red/green color blind, so that may make a difference.)

inkandpaper-app:
If the reviewers agree, I think it would be better to remove the colors entirely. I am attaching two versions anyway. However, I wouldn't want to open a pull request for such a small change; I believe the reviewers can make the edit themselves if they wish (and adjust the colors, if necessary).

[black_and_white.zip](https://github.com/user-attachments/files/30878802/black_and_white.zip)
[colored.zip](https://github.com/user-attachments/files/30878803/colored.zip)

bmjcode:
@inkandpaper-app Don't worry about opening small PRs for our sake. Honestly it's easier to review, and we prefer doing everything through a PR anyway for accountability reasons.

PR #2068, opened 2025-08-15
Bump version to 4.0.5.dev1

JunTamura1959:


fedelibre:
Thanks for the reminder.

You should also bump the version in pyproject.toml (used by Briefcase to build Windows and macOS packages).

JunTamura1959:
@fedelibre Thank you! I was not aware of pyproject.toml. I've updated the file.

PR #2080, opened 2025-08-30
Use Python 3.13 in GitHub Actions workflow

fedelibre:
Python 3.13 is mature and it's what I've been using for months on Linux and Windows.
Let's use it to build the Windows and macOS packages.


bmjcode (approved):
Fine with me.

PR #2100, opened 2025-10-02
Better regex for finding pitch names in MIDI input

ksnortum:
I am very nervous about this PR!  I don't have a way to test it in Frescobaldi because I can't get MIDI input to work for me. (I *have* tested it in a Python program I wrote and at Regex101.com.)

 I would love volunteers to test MIDI input with the "re-pitch" checkbox checked.  I believe this will take an existing piece, find the next pitch name, and replace it with the MIDI input.  It would be great to test it before and after the PR is applied so that we can tell if re-pitch is actual working.  My hunch is that it won't work well in some languages.

Ideally, this PR would be testing with "a b c" pitch name languages (like English or Netherlands) and "do re mi" pitch name languages (like Italian or French).  @kledeen, would you be willing to test this?

### DIFF
--- a/frescobaldi/midiinput/__init__.py
+++ b/frescobaldi/midiinput/__init__.py
@@ -29,12 +29,16 @@ import documentinfo
 
 from . import elements
 
-# What this does was originally undocumented. It appears intended to match
-# chord and pitch names, but not commands or variables (thanks @ksnortum)
+# Match LilyPond pitch names and chords, but not variables or commands
 LY_REG_EXPR = re.compile(
-    r'(?<![a-zA-Z#_^\-\\])[a-ps-zA-PS-Z]{1,3}(?![a-zA-Z])[\'\,]*'
-    '|'
-    r'(?<![<\\])<[^<>]*>(?!>)'
+    r"(?<![a-zA-Z_\\:#\"-])" # Ensure no letters, underscores, or special characters precede
+        r"(?:(?:(?:do|re|ré|mi|fa|sol|la|si)[db]{0,2})" # do re mi pitch names and modifiers
+        r"|" # or ...
+        r"(?:(?:[a-h](?:s|is|x|f|es|as){0,2})))" # a b c pitch names and modifiers
+        r"[',]*" # optional octave indicators
+        r"(?![a-zA-Z_\"-])" # Ensure no letters, or special characters follow
+    r"|" # or ...
+    r"(?<![<\\])<[^<>]*>(?!>)" # Match angle brackets with content inside
 )
 
 # Event codes from the MIDI specification


fedelibre:
@ksnortum Do you have push access to this repository?

If you push a new branch to the frescobaldi repository instead of your fork, then you can run the packages workflow on your branch here:
https://github.com/frescobaldi/frescobaldi/actions/workflows/release.yml

then ask users who may not be able to test a patch from git source code to install the new installer.

That said, I have a MIDI keyboard to test this patch (from source) but I don't think I'll have time this week.

ksnortum:
> I don't have push permissions to frescobaldi/frescobaldi.

I do have those permissions now and I'll create a release for the branch `pitch-name-regex`.

kledeen:
I'm glad to do testing, but I won't be home for another week, possibly
more.  If the issue is still open, I'll check then.

ksnortum:
I have pushed the branch `pitch-name-regex` to this repository and created a release for Windows and Mac users [here](https://github.com/frescobaldi/frescobaldi/actions/runs/18431364318). Linux users I think will still need to go to the branch and download the tarball from there.

fedelibre:
> I am very nervous about this PR! I don't have a way to test it in Frescobaldi because I can't get MIDI input to work for me.

Why you can't get MIDI input to work? On Linux? I'm planning to add some MIDI instructions in the user manual so I'm curious to know about the current problems.

Honestly, I have a MIDI keyboard but I never use it so I'm not so motivated to test this PR. I'll do it though, as I'm sorry this PR has been ignored so far. Let's see what I find...

fedelibre:
Well, we should understand what this patch is trying to fix exactly.

I've tested only master so far and I've found that re-pitch works fine with the nederland note names (lilypond default) but it doesn't work well with Italian note names: if the replaced note name is the same, it changes the next note instead of "confirming" the current note. See this video:


https://github.com/user-attachments/assets/9d6f4d37-0f49-48c6-8829-a4f2470ca342

ksnortum:
> >I don't have a way to test it in Frescobaldi because I can't get MIDI input to work for me.
> 
> Why you can't get MIDI input to work? On Linux? I'm planning to add some MIDI instructions in the user manual so I'm curious to know about the current problems.

If I remember correctly, I was having problems with the MIDI input in other programs as well as Frescobaldi.  I was trying it on a Windows machine but I could try it on Linux too to see if my MIDI setup is working at all.

ksnortum:
I got MIDI input to work on a Linux system and this patch seems to fix the problem shown in the video.  I will do more testing later.

ksnortum:
I have pushed a commit to improve the pitch-finding regex.  It seems to be working as expected in all the scenarios I could think of.  Here's some of the LilyPond code I used to test it:

```
\version "2.24.4"
\language "italiano"

{
  % italiano
  s1 % do'2 re'4 mib'_"mi" fa'2-\tweak Y-offset #-1 ^"test"  sol' 
}

{
  % netherlands
  s1 % c'2 d'4 es'_"accent" f'2-\tweak Y-offset #-1 ^"test" g' 
}

{
  % netherlands chords
  s1 % <c' e'>4 <d' f'>4_"accent" <es' g'>4-\tweak Y-offset #-1 ^"test" <f' a'> 
}

{
  % italiano chords
  <do' mi'>4 <re' fa'>4_"accent" <mib' sol'>4-\tweak Y-offset #-1 ^"test" <fa' la'>
}
```

I would welcome any more testing.

fedelibre:
I've now tested your branch. I confirm it solves the bug I showed in the video. I've also played with your testing file and it seems to work fine.

I'm fine with this PR.
Feel free to merge it if you think it's complete.

fedelibre (approved):

Issue #2126, opened 2025-12-11
Frescobaldi crashes when selecting text [Arch Linux]


HagenBarz:
When selecting a part of the text in the editor window via mouse or [shift] + arrow keys [up down left right] Frescobaldi crashes with 

```
qt.qpa.services: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an application ID")
Speicherzugriffsfehler     (Speicherabzug geschrieben) frescobaldi
```
(memory access error) 

I tried 4.0.4, 4.0.3 and 4.0.2.
It seems an issue of its dependencies. Frescobaldi was running smoothly before.

Frecobaldi 4.0.4
lilypond 2.24.4-5
python 3.13.11-1
python-ly 0.9.9-1
python-pygame 2.6.1-2
python-pyqt6-sip 13.10.3-1
python-pyqt6-webengine 6.10.0-1
python-qpageview 1.0.0-1
qt6-base 6.10.1-1
qt6-svg 6.10.1-1

Nodd:
I also have this bug, with the error message:
`fish: Job 1, 'frescobaldi …' terminated by signal SIGSEGV (Erreur de frontière d’adresse)`

```console
> frescobaldi -V
Frescobaldi: 4.0.4
Extension API: 0.9.0
Python: 3.13.11
python-ly: 0.9.9
Qt: 6.10.1
PyQt: 6.10.1
qpageview: 1.0.0
OS: Linux-6.17.9-arch1-1-x86_64-with-glibc2.42 (Arch Linux)
Installation kind: distro package or installed from source
```

aiomaster:
Maybe it is some Wayland QT problem. I made a backtrace with gdb. Maybe it is usefull.
```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
#0  0x00007fffd1a2ee83 in QtWaylandClient::QWaylandClipboard::setMimeData(QMimeData*, QClipboard::Mode) () from /usr/lib/qt6/plugins/platforms/../../../libQt6WaylandClient.so.6
#1  0x00007ffff1d1a8cf in ?? () from /usr/lib/libQt6Widgets.so.6
#2  0x00007ffff1d1e442 in QWidgetTextControl::processEvent(QEvent*, QTransform const&, QWidget*) () from /usr/lib/libQt6Widgets.so.6
[...]

PR #2131, opened 2025-12-26
Stop crash when selecting text

ksnortum:
A per @mitya57, set parent of m before selecting text to prevent crash.  Fixes issue #2126


mitya57:
Per https://www.riverbankcomputing.com/pipermail/pyqt/2025-December/046417.html, let’s put it on hold for now.

bmjcode:
I agree with putting this on hold. This looks to be PyQt's bug, not ours.

ksnortum:
As per #2126, the release of python-pyqt6 6.10.2-1 solved the problem without this patch.  Closing.

PR #2135, opened 2026-01-08
Use treble_8 clef by default for classical guitar

bmjcode:
We have discussed before whether an octave or plain clef is more appropriate for guitar (style guides differ on this point). My impression last time was that octave clefs are more accepted/preferred in the classical guitar world, while plain clefs are more common in popular music. This patch sets the Score Wizard's defaults accordingly for each type of guitar.

### DIFF
--- a/frescobaldi/scorewiz/parts/plucked_strings.py
+++ b/frescobaldi/scorewiz/parts/plucked_strings.py
@@ -343,6 +343,9 @@ class Guitar(TablaturePart):
         layout.addLayout(box)
         self.octaveClef = QCheckBox()
         layout.addWidget(self.octaveClef)
+        # Use treble_8 by default for (classical) Guitar but not derived
+        # types, since it is less common in the styles those are intended for
+        self.octaveClef.setChecked(type(self) is Guitar)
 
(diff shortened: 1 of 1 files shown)


fedelibre:
We discussed it last year [here](https://github.com/frescobaldi/frescobaldi/pull/1810#issuecomment-2612651317).
Your main argument was: the other notation apps do not use the octave clef.

AFAICS octave clefs are used especially in educational/theoretical contexts, but they are often omitted in any music (not only popular).

I don't have a strong opinion on this matter.

bmjcode:
Honestly I might just be second-guessing myself for changing the default here. I've never seen a guitar book that uses octave clefs (speaking as a rock/pop musician), but guitar also isn't really my area of expertise. My thinking this time was

* [LilyPond's developers use them](https://lilypond.org/doc/v2.24/Documentation/notation/guitar),
* we've historically followed what they do,
* people who know more than me have very strong opinions, and
* it's pretty technically simple to please and/or annoy both camps equally.

I know it's so uncharacteristic of me to overthink something simple like this. :)

fedelibre (approved):

PR #2139, opened 2026-01-18
Add the vcs-browser URL to the metainfo file

fedelibre:
Flathub linter considers this missing tag only a warning now, but warnings can be promoted to errors in the future. Let's fix it.


bmjcode (approved):
You know more about this than I do, so if you say it's needed, let's do it.

Issue #2144, opened 2026-01-28
“Length” sometimes not available


gabriel-ellsworth:
“Length” is sometimes missing from my user interface in Frescobaldi for no apparent reason. Usually, if I quit Frescobaldi and open it again, I will see the duration counter. But sometimes that doesn’t work, and there is no way (to my knowledge) to force “Length” to come back in those instances.

Frescobaldi: 4.0.4
Extension API: 0.9.0
Python: 3.12.8
python-ly: 0.9.9
Qt: 6.9.0
PyQt: 6.9.1
qpageview: 1.0.1
OS: macOS-26.1-arm64-arm-64bit
installation kind: standalone .app bundle

fedelibre:
I had to look at the code to understand this issue :-)
It's a feature I never noticed.

When you select some notes in the text view, the *Pos:* field at the bottom of the text view changes to *Length:*.

<img width="508" height="398" alt="Image" src="https://github.com/user-attachments/assets/d7230b70-13c4-47fc-9204-421d2a6e9a91" />

<img width="508" height="398" alt="Image" src="https://github.com/user-attachments/assets/778bdf01-e25f-4f89-98c9-044d0a2c17f7" />

I did a few tests and Length always appeared when I selected a few notes. But, as you say, it happens seldomly so it might be a difficult bug to debug.

IMO the Length counter, which shows the duration counter in the current bar, is much more useful than the Pos counter, which shows the duration counter from the first bar up to the current bar. What if Length counter became the new default? @bmjcode @ksnortum What do you think?

I mean, in this screenshot you see that 69/16 is not giving any useful advice for someone who is entering music:

<img width="508" height="398" alt="Image" src="https://github.com/user-attachments/assets/a93290cd-33a3-4f91-8a2f-f090fb1ffacb" />

ksnortum:
I would vote to leave it as it is.  I think some people may rely on the position counter and would complain if it were removed.  I can see the facility of having both the position from the beginning and a length of selected notes.

bmjcode:
I've honestly never paid much attention to that counter either way, at least while editing scores. I know the code pretty well because I think it's one of the causes of #473.

gabriel-ellsworth:
Thank you all! I rely on Length fairly often. This snippet gives an example use case.

[Unmetered music: An alternative to \cadenzaOn
](https://wiki.lilypond.community/wiki/Unmetered_music:_An_alternative_to_%5CcadenzaOn)

[...]

PR #2145, opened 2026-01-30
Add a toggle to hide the music position display in the status bar

bmjcode:
~Turning this off may reduce lag since, as discussed in #473, calculating this position requires re-parsing the document after every change.~

Edit: This doesn't help much with lag because autocomplete has a much larger impact. I'd still like this for its own sake though.

### DIFF
--- a/frescobaldi/musicpos.py
+++ b/frescobaldi/musicpos.py
@@ -22,7 +22,7 @@ Shows the time position of the text cursor in the music.
 """
 
 
-from PyQt6.QtCore import QTimer
+from PyQt6.QtCore import QSettings, QTimer
 from PyQt6.QtWidgets import QLabel
 
 import weakref
@@ -43,6 +43,8 @@ class MusicPosition(plugin.ViewSpacePlugin):
         view = space.activeView()
         if view:
             self.slotViewChanged(view)
+        app.settingsChanged.connect(self.reloadSettings)
+        self.reloadSettings()
 
     def slotViewChanged(self, view):
         old = self._view()
@@ -62,12 +64,13 @@ class MusicPosition(plugin.ViewSpacePlugin):
 
     def startWaitTimer(self):
         """Called when the document changes, waits longer to prevent stutter."""
-        self._waittimer.start(900)
-        self._timer.stop()
+        if self._visible:
+            self._waittimer.start(900)
+            self._timer.stop()
 
     def startTimer(self):
         """Called when the cursor moves."""
-        if not self._waittimer.isActive():
+        if self._visible and not self._waittimer.isActive():
             self._timer.start(100)
 
     def slotTimeout(self):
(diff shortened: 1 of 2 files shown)


ksnortum:
I had high hopes for this PR!  Alas, it doesn't seem to reduce lag, for me at least.  Here's how I tested it.

1) After downloading the PR, toggling off the option, and restarting Frescobaldi, I loaded a large file, like [this one](https://github.com/ksnortum/chopin-opus-54/blob/main/scherzo-op54-parts.ily).

2) Near the top of the file, I typed in the following: `{ c4 c c c }` 

3)  I edited the line to add octave marks: `{ c'''4 c c c }`.  Typing in the octave marks, I noticed a definite lag.

It's hard to tell if the lag is less than it was, but it doesn't seem like it.  But this PR may be one piece of the puzzle.

bmjcode:
What about turning off both the music position and autocomplete?

ksnortum:
Sorry, how do I turn off autocomplete?

bmjcode:
Tools -> Automatic Completion.

FYI, I'm working on another idea that might actually fix the lag rather than work around it. Watch #473. :)

ksnortum:
From my unscientific experiments, turning autocomplete off has a large effect on lag and turning position display off has a small effect.  I can't tell the difference between when position display is disabled or not when autocomplete is off.  But this PR may still do some good.  I'm looking forward to testing PR #2147.

bmjcode:
> From my unscientific experiments, turning autocomplete off has a large effect on lag and turning position display off has a small effect. I can't tell the difference between when position display is disabled or not when autocomplete is off.

That makes sense. Autocomplete requires much more processing so of course it would lag more. I was just so focused on this small improvement that I forgot for a moment that autocomplete exists. :)

fedelibre:
My humble assumption: the autocomplete should not cause lag **IF** it was limited to the known LilyPond syntax. The problem is that the autocomplete module wants to also check for functions defined in the included files and this causes troubles when the included files are many or big files. Correct?

Then a possible workaround could be to limit the autocomplete to the LilyPond official/standard syntax and forget about the user defined functions in the included files? Just in case #2147 does not solve the lag issue (I haven't tested it yet).

ksnortum:
> My humble assumption: the autocomplete should not cause lag **IF** it was limited to the known LilyPond syntax. The problem is that the autocomplete module wants to also check for functions defined in the included files and this causes troubles when the included files are many or big files. Correct?

I have not tested this, but I will.

> Then a possible workaround could be to limit the autocomplete to the LilyPond official/standard syntax and forget about the user defined functions in the included files? Just in case #2147 does not solve the lag issue (I haven't tested it yet).

I would hate that change, personally.  At least make it a flag that can be turned on and off.

ksnortum:
I tried editing my 5500 line file without any includes, and the lag is just as bad without them.  (This is my unscientific opinion.)

fedelibre:
I'm now testing this PR.

@bmjcode I don't know if it is intended: right now this preference turns off also the length of selected notes. You may change the description to "Show music position or selected notes length in the status bar". Otherwise you may make the two "counters" independent and add a preference for each. What do you think? We discussed about this feature in #2144

bmjcode:
@fedelibre The same widget handles both. It just changes its behavior based on whether there's a selection or not.

bmjcode:
This will need some reworking now that we've merged #2152. I may come back to it at some point, but it's not a priority.

PR #2147, opened 2026-01-31
Generate music.Document instances in a separate thread

bmjcode:
This is yet another attempt to solve #473. In theory this is the simplest possible implementation: We use threading internally to do the heavy lifting in the background, but maintain a synchronous external interface so callers do not need to be thread-aware.

**This will need extensive testing before we consider merging it.** This is a significant architectural change to a heavily-used section of the code, and while I think I know what I'm doing here, I'm not yet 100% certain of it. Right now I would consider any of these outcomes equally likely:

* Best case – it works exactly like I think it does and completely solves the problem.
* Middle case – it doesn't really help anything, but doesn't break anything either.
* Worst case – I didn't think this through properly at all and just introduced all manner of exciting new crashes.

### DIFF
--- a/frescobaldi/documentinfo.py
+++ b/frescobaldi/documentinfo.py
@@ -28,7 +28,7 @@ import os
 import re
 import weakref
 
-from PyQt6.QtCore import QSettings, QUrl
+from PyQt6.QtCore import QCoreApplication, QMutex, QSettings, QThread, QUrl
 
 import document
 import qsettings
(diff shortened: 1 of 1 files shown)


bmjcode:
No crashes here so far, but I also can't tell if it's really improved anything. Anyone else want to try?

ksnortum:
I'm using it.  Same: no crashes but no noticeable improvement either.

bmjcode:
Another one for our list of failed experiments then.

No point keeping this open if it isn't helping. I'm sure I'll take another stab at it eventually...

PR #2152, opened 2026-02-11
Use a background thread for music related parsing/computation while editing

bmjcode:
This is my latest attempt to solve #473.

Most of the lag described in that issue comes from two methods in the `DocumentInfo` class. The data returned by these methods is slow to generate, yet must be regenerated after each document change. Right now that work is performed in the main thread the first time one of those methods is called after a change (the results are then cached for future calls). This PR moves that data generation to a background thread, where it is continuously updated as the document changes, so `DocumentInfo` only ever has to retrieve it from cache in the main thread.

The usual caveats apply: Threading is complicated, and it's quite possible I don't understand the effects of introducing it here as thoroughly as I think I do. This could cause Frescobaldi to crash in exciting new ways, or just not actually help with the problem. (That said, testing with @ksnortum's [Chopin scherzo](https://github.com/ksnortum/chopin-opus-54) did show a noticeable improvement on my machine.)

### DIFF
--- a/frescobaldi/documentinfo.py
+++ b/frescobaldi/documentinfo.py
@@ -144,13 +144,10 @@ class DocumentInfo(plugin.DocumentPlugin):
             doc.changesStopped.connect(self._processChanges)
             doc.closed.connect(self._reset)
 
-    # connect to this to be notified when lydocinfo() and music() are
-    # updated; otherwise, they return cached values that may be a couple
-    # keystrokes behind while the user is typing
     contentsChanged = signals.Signal()
 
     def _reset(self):
-        """Clear cached data when the document is changed or closed."""
+        """Clear cached data when the document is opened or closed."""
         self._waitForWorker()
         self._lydocinfo = None
         self._music = None
@@ -317,16 +314,19 @@ class DocumentInfo(plugin.DocumentPlugin):
         return []
 
     def _invalidate(self):
-        """Called when the document is changed.
+        """Invalidate the cache when the document is changed.
 
-        This invalidates cached data so the worker is forced to
-        regenerate it next time _processChanges is called.
+        This forces the worker to regenerate cached data the next time
+        _processChanges is called. The old data remains in the cache
+        until the worker finishes, so DocumentInfo will return slightly
+        outdated information while the worker is running. (This is the
+        dirty secret to how we avoid lag.)
 
         """
         self._documentChanged = True
 
     def _processChanges(self):
-        """Called when the document is changed.
+        """Update cached data when the document is changed.
 
         This triggers a worker to regenerate _lydocinfo and _music
         in a background thread. These are slow operations, and
@@ -335,6 +335,11 @@ class DocumentInfo(plugin.DocumentPlugin):
         versions (see issue #473).
 
         """
+        # Note the worker operates on the live Document instance,
+        # so it always sees the latest changes even if they happen
+        # while the worker is running. There is no need to ever
+        # retrigger an active worker; doing so will only cause more
+        # of the very lag this design was intended to reduce.
         if self._documentChanged and not self._workerActive:
             try:
                 worker = self._worker_instance
@@ -383,6 +388,8 @@ class _Worker(QObject):
     @classmethod
     def preferredThread(cls):
         """Return the QThread where we want workers to live."""
+        # all workers can share the same thread since we only have one
+        # active document, and therefore one active worker, at a time
         try:
             thread = cls._thread
         except AttributeError:
@@ -392,7 +399,7 @@ class _Worker(QObject):
         return thread
 
     # these are static methods so we can also call them directly from the
-    # main thread as needed (for example, in the DocumentInfo constructor)
+    # main thread as needed
 
     @staticmethod
     def lydocinfo(document):


bmjcode:
After tracing through its execution a few times, I feel more confident in the new code's correctness. I've found (and fixed) a couple minor glitches, but my assumptions about thread safety seem to be holding up, and nothing has crashed here so far. The big question now is, does it provide enough of an improvement to be worth the effort?

ksnortum:
Is it ready for me to fetch it, or should I wait?

bmjcode:
Fetch away!

ksnortum:
So far, so good!  There is no (perceptible) lag on big files and no craches yet!

fedelibre:
I've built installers for macOS and Windows here:
https://github.com/frescobaldi/frescobaldi/actions/runs/22003173948

in case someone wants to test this PR without having to deal with running Frescobaldi from source.

fedelibre:
And here's a flatpak package for Linux users:

```
flatpak install --user https://dl.flathub.org/build-repo/256705/org.frescobaldi.Frescobaldi.flatpakref
```

Launch it from terminal so you are sure it's the version you will use:

```
flatpak run org.frescobaldi.Frescobaldi//test
```

and you can easily compare the performance with the flathub stable branch:

```
flathub run org.frescobaldi.Frescobaldi//stable
```

fedelibre:
@bmjcode Sorry to say that I'm not seeing any improvement. I've tested Knute's file of Chopin op54: if I try to add some more notes to the big file it lags, especially when I start typing a command like `\glissando`.

I'm comparing the current stable and the test flatpak packages mentioned above. I've verified that the test branch contains the code of this PR:

```
$ flatpak run --command=sh org.frescobaldi.Frescobaldi//test
[📦 org.frescobaldi.Frescobaldi frescobaldi]$ 
[📦 org.frescobaldi.Frescobaldi frescobaldi]$ find /app -name app.py
/app/lib/python3.12/site-packages/frescobaldi/app.py
[📦 org.frescobaldi.Frescobaldi frescobaldi]$ find /app -name app.py | xargs grep worker_thread
_worker_thread = None
def worker_thread():
    global _worker_thread
    if _worker_thread is None:
        _worker_thread = QThread()
        _worker_thread.finished.connect(_worker_thread.deleteLater)
        _worker_thread.start()
    return _worker_thread
```

ksnortum:
That is strange.  I just checked out the master branch and confirmed that there was lag. (I couldn't see lag typing `\glissando` but I did when adding octave marks repeatedly.)  I then checked out  the PR2152 branch and the lag went away, at least it wasn't perceivable.  I wonder if it has to do with our systems specs.  I'm running an AMD CPU @ 5.26 GHz with almost 30 gigs of memory.

bmjcode:
> @bmjcode Sorry to say that I'm not seeing any improvement. I've tested Knute's file of Chopin op54: if I try to add some more notes to the big file it lags, especially when I start typing a command like `\glissando`.

Does it lag any differently, though? I'm asking that sincerely. The lag is a side effect of slow but necessary processing; this PR doesn't eliminate it completely, it just tries to hide as much of it as possible.

fedelibre:
It seems it lags in the same way.

I think @ksnortum may be right when he says that it might depend on the system specs. I did the test on a 15 year old laptop with the following specs:

```
Intel® Core™ i7-3537U × 4
RAM 8GB
SSD disk
```

I'll try later with the PC at work, which has a slightly better hardware.

I will ask on the lilypond-user mailing list if anyone is interested in testing this PR. The more people test it, the better we can understand how much helps with this old issue.

fedelibre:
It seems that my problem is that my laptop processor has only two physical cores, so probably it cannot take advantage of this patch. Knute probably has 6 or more, as he has a modern processor.

bmjcode:
Interesting. I'll have to see what it does on my laptop. It's a little newer (6th generation Core i5) but still only dual-core IIRC.

papioara:
I just did a test with three candidates on my machine (Kernel: 6.14.0-37-generic arch: x86_64 bits: 64
  Desktop: Cinnamon v: 6.6.7 Distro: Linux Mint 22.3 Zena)
Candidate 1: Frescobaldi installed via pipx. The lilypond file compiled ok with just one little warning (see screenshot below):

Bildschirmfoto vom 2026-02-17 14-22-34

Candidates 2 and 3 were the Frescobaldi test flatpak and the stable one. Both aborted with the same error message:

Bildschirmfoto vom 2026-02-17 14-22-03

fedelibre:
My PC at work has a quad core: Intel® Core™ i5-3350P × 4.

In fact the performance experience is slightly better than in my laptop. That is true for the version before this PR and for this PR version.
Yes, I notice an improvement using this PR but it's not a big one. Still it is useful.

For anybody who wants to test this PR: we are not evaluating the speed of lilypond, so it's not about compiling the score. We are evaluating the editing speed of Frescobaldi.
Try to edit the music and see if the interface is responsive as it should or if it becomes slow; this happens especially when typing a command, like \glissando. Remember that you have to type at least two letters for the autocomplete to be triggered (as in \gl). You may notice that sometimes the command list appears immediately and sometimes you have to wait for a few seconds.

ksnortum:
I just tested this PR on a system with a Pentium Sliver @ 1.50 GHz and four cores.  There is definitely lag, although in different places.  I think it's a little better than the master branch, although it is still not very workable.  We may have to have "suggested minimum system requirements."  I still think this is a step in the right direction.

bmjcode:
If any of you are running from source, try profiling Frescobaldi with:

`python3 -m cProfile -s cumtime -m frescobaldi > results.txt`

then post your `results.txt` here. That lists all Frescobaldi's function/method calls ordered by cumulative execution time, which I've been using to identify where the slowdown is occurring.

papioara:
I have done another test of the test flatpak, and this time (thanks to the clarification above) I have focused on the responsiveness of the editor and the autocomplete feature. And this time after some editing of the .ly file which basically was deleting some strings and then restoring them manually, there were no compilation errors at all.

Only good news: Autocomplete works perfectly, and the editor feels responsive on my machine without any lag whatsoever. Point and Click on a note in the Music View opens the parts.ily file immediately.

For me everything feels good here...

ksnortum:
Here is my `results.txt` file for a 1.5 GHz processor with four cores.
[results-1.5Ghz.txt](https://github.com/user-attachments/files/25376411/results-1.5Ghz.txt)

ksnortum:
Here is my `results.txt` for a 5.25 GHz processor. 
[results-5.26GHz.txt](https://github.com/user-attachments/files/25376499/results-5.26GHz.txt)

fedelibre:
Don't know if it makes sense posting the results of a dual core.. The perceived performance was bad on both master and threaded-documentinfo branch.

[dualcore-pr2152-results.txt](https://github.com/user-attachments/files/25384910/dualcore-pr2152-results.txt)
[dualcore-master-results.txt](https://github.com/user-attachments/files/25384911/dualcore-master-results.txt)

papioara:
Sorry - I forgot to communicate the specs of my machine. Here's the inxi output:

`CPU: 6-core AMD Ryzen 5 5600G with Radeon Graphics (-MT MCP-)
speed/min/max: 2675/400/4465 MHz Kernel: 6.14.0-37-generic x86_64 Up: 27m
Mem: 2.74/30.7 GiB (8.9%) Storage: 4.14 TiB (32.5% used) Procs: 477
Shell: Bash inxi: 3.3.34
`
Hope that helps, but as I already said above, the test flatpak works well here without any lags even when I add some command to the ...parts.ily file.

bmjcode:
Thanks all! I'll look at these when I have a bit more time.

> Don't know if it makes sense posting the results of a dual core.. The perceived performance was bad on both master and threaded-documentinfo branch.

That's exactly why those results are useful -- to find out if there's anything we can improve on that machine.

bmjcode:
I just pushed a change that might slightly improve performance. Haven't yet had a chance to look at profiling results but hopefully I will get to that soon.

bmjcode:
I just did another small fix in `document.py`. This one seems so obvious in hindsight I wonder why I didn't just do it that way before. My thinking with the last couple fixes is this code runs *constantly*, so even tiny improvements will quickly add up.

ksnortum:
Having re-pulled this PR, I'm noticing something I didn't before.  If I type `\relative {` and `<enter>`, the editor doesn't indent the next line (this is using a big file.)  The same thing happens with anything that usually indents, like typing `{` and `<enter>`.

bmjcode:
Interesting. I'm not sure offhand what would cause that, but based on past attempts, I'm also not surprised that this PR has weird side effects...

bmjcode:
I finally got my own Linux machine up and running, and wow. You all weren't kidding about the lag. Still need to dig into those profiling results to find out what's going on.

bmjcode:
I think I fixed the typing lag.

There's still some weirdness with the autocomplete popup and other things but those are relatively minor in comparison.

bmjcode:
Force-pushed to remove a couple unrelated commits (fixes for unused imports) that snuck in.

bmjcode:
One more force-push to rebase this on the current master.

This is ready to test now.

ksnortum:
So far, so good!  No lag and no crash!

fedelibre:
I'm testing this branch on my laptop and I see BIG improvements!

Frescobaldi doesn't freeze anymore while I edit the Chopin piece. Sometimes it lags a bit when searching the available options for a \foo command, but much less than before.

bmjcode:
> So far, so good! No lag and no crash!

> I'm testing this branch on my laptop and I see BIG improvements!

Excellent!

> Sometimes it lags a bit when searching the available options for a \foo command, but much less than before.

That's because now you can type faster than the completion logic can process your changes. I haven't yet figured out a solution to that one.

bmjcode:
You all have spent more time actually using this than I have, so: Does this need more work before we consider merging it? Or are the rough spots like autocomplete minor enough to save for a future PR?

I'm asking mostly in terms of the user experience; I'm feeling pretty good about the technical side.

fedelibre:
I think we can consider merging it.

Kudos to Ben for the perseverance in fixing this difficult bug! This was one of the oldest and most annoying Frescobaldi bug for many users.

bmjcode:
It feels good to be nearly done with this one.

I just did a little bit of cleanup and a lot of commenting and internal documentation. The overall implementation turned out to be simpler than I'd expected, but the actual logic of it still feels somewhat counterintuitive.

This is about as clean now as I can probably get it, so let me know if you think it's ready to merge.

bmjcode:
Done!

Issue #2132, opened 2026-01-02
TypeError: NamedTemporaryFile() got an unexpected keyword argument 'delete_on_close'


taostream1-cell:
Happy New Year and Greetings from the UK
Have been running Frescobaldi without any issues for several months without issues but since reinstalling the Linux VE and apps on a Chromebook I'm getting the following: (dev help gratefully received as I can't intra-download or point to any external downloads of any Lilypond versions)

Traceback (most recent call last):
  File "/home/taostream1/.local/pipx/venvs/frescobaldi/lib/python3.11/site-packages/frescobaldi/preferences/lilypond.py", line 351, in downloadFinished
    with tempfile.NamedTemporaryFile(delete_on_close=False) as tfile:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: NamedTemporaryFile() got an unexpected keyword argument 'delete_on_close'

Please describe below what you were doing:

Frescobaldi: 4.0.4
Extension API: 0.9.0
Python: 3.11.2
python-ly: 0.9.9
Qt: 6.10.0
PyQt: 6.10.1
qpageview: 1.0.1
OS: Linux-6.6.99-08879-gd6e365e8de4e-x86_64-with-glibc2.36 (Debian GNU/Linux 12 (bookworm))
Installation kind: distro package or installed from source

bmjcode:
The `delete_on_close` keyword was [added in Python 3.12](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile). Are you able to install a newer Python version on your Chromebook?

taostream1-cell:
Many thanks for your reply. I will see if I can update with a Python 3.12
version and report back.

taostream1-cell:
Greetings. Thank you, bmjcode....installation of Python 3.12 was the solution : )
Wishing you and all a very happy, healthy and successful 2026! Thanks for all you do for this fantastic project.

bmjcode:
Glad you got it working, and happy new year to you too!

PR #2157, opened 2026-02-12
Requires 3.10 as minimum Python version

fedelibre:
as 3.8 is end of life since 2024.

Remove the classifiers in pyproject.toml as they
are needed only to publish to PyPI, where
Frescobaldi is not distributed.

### DIFF
--- a/frescobaldi/appinfo.py
+++ b/frescobaldi/appinfo.py
@@ -44,7 +44,7 @@ appname = "Frescobaldi"
 desktop_file_name = "org.frescobaldi.Frescobaldi"
 
 # required versions of important dependencies
-required_python_version = (3, 8)
+required_python_version = (3, 10)
 required_python_ly_version = (0, 9, 4)
 required_qt_version = (6, 6)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ description = "LilyPond Music Editor"
 authors = [{name = "Wilbert Berendsen", email = "info@frescobaldi.org"}]
 maintainers = [{name = "Wilbert Berendsen", email = "info@frescobaldi.org"}]
 license = {text = "GPL-2.0-or-later"}
-requires-python = ">=3.8"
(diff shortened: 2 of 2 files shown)


bmjcode:
We should probably require at least 3.12; see #2132.

fedelibre:
The idea of bumping the Python version came exactly from #2132. However, we should try to require the very minimum version possible. That method is used for downloading lilypond, which is a nice feature but not essential.

In Windows and macOS packages there's an embedded Python automatically installed by briefcase and it's currently 3.12 IIRC.
So - excluding the few who run Frescobaldi from source and know what they do - this setting should affect only Linux users. Most Linux recent distros has at least 3.12, even Mint and Debian stable. If we set a minimum of 3.10 we do not close the door to users of old distros.

I don't have a strong opinion on this actually. I'm just following the conservative approach normally used for version requirements. But if you think we should really avoid #2132 then let's bump the version to 3.12.

bmjcode:
> However, we should try to require the very minimum version possible. [...] If we set a minimum of 3.10 we do not close the door to users of old distros.

Fair. I may need to revisit that bit of code then. As I recall, the issue there was Python changed `tempfile`'s default behavior in 3.12. At the time I couldn't think of an elegant way to maintain both correct functionality and backwards compatibility.

fedelibre:
If we consider EOL dates to decide when to bump the minimum version, then we'll have to bump it again in 8 months from now:

Python EOL dates:
* 3.10: October 2026
* 3.11: October 2027

Should we merge this PR now or wait for you to see if you can revisit that bit of code?

bmjcode:
I'll try to revisit that code when I get a chance. I've already done something similar for Qt in #2171. After just setting up my own Linux system for debugging, I'm reminded you can't rely on the distros to ship current versions of *anything*...

bmjcode:
I finally got around to fixing LilyPond downloads. See #2181.

fedelibre:
Let's merge this one then

PR #2162, opened 2026-02-16
Add a GitHub Action workflow to build a flatpak package

fedelibre:
A GitHub workflow in this repo seems to be the best and easiest way to provide flatpak packages for Linux users who want to easily test a PR without having to deal with the setup of an installation from source.

I still have to verify that it actually works...

Other ways I've excluded:

- Flathub beta channel, as it tends to create stale versions.
- Draft PRs on github.com/flathub/org.frescobaldi.Frescobaldi. This is what I've done so far, but it's quite boring to do.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -1,0 +1,61 @@
+name: Flatpak CI
+
+# Run only manually to let Linux users easily test some PRs.
+on:
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


fedelibre:
It works, as shown in the [last build](https://github.com/frescobaldi/frescobaldi/actions/runs/22100577408). There's a duplicate artifact in that build but I've fixed that.

I'll merge this PR by Friday.

PR #2166, opened 2026-02-26
flatpak: upgrade to 6.10 KDE runtime

fedelibre:
The Flathub manifest [now uses version 6.10](https://github.com/flathub/org.frescobaldi.Frescobaldi/pull/92) so we have to use the same version in our workflow file.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -24,34 +24,43 @@ jobs:
       - name: Prepare CI manifest
         run: |
           cp -r flathub-manifest/. .
-          # Remove Flathub frescobaldi module
-          awk '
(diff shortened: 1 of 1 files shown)


fedelibre:
It [builds fine now](https://github.com/frescobaldi/frescobaldi/actions/runs/22445753931), I'll merge it.

For the records, the build failed first because flatpak-github-actions/flatpak-builder modifies the .yaml manifest file before starting the build and it was not able to interpret correctly the `6.10` version (it changed it to 6.1). So I made the actual fix in the flathub manifest file, wrapping the runtime versions in quotes, as in `"6.10"`.

However, I'll keep yq in place of awk for the yaml modification as it is more reliable.

PR #2171, opened 2026-03-08
Check for specific Qt features rather than the overall version

bmjcode:
This allows Frescobaldi to run with reduced functionality when certain features are unavailable and the alternative would be outright crashing. It is mainly intended as a workaround for Linux users since many distributions ship outdated Qt packages that are difficult to upgrade.

A corresponding fix for qpageview is implemented in frescobaldi/qpageview#49.

### DIFF
--- a/frescobaldi/checks.py
+++ b/frescobaldi/checks.py
@@ -80,12 +80,25 @@ but requires at least version {r[0]}.{r[1]}.")
 
 # Check Qt version
 from PyQt6.QtCore import QLibraryInfo, QVersionNumber
-v = QLibraryInfo.version()
-r = QVersionNumber(*appinfo.required_qt_version)
-if v < r:
-    error("Qt version is too old.",
-        f"Frescobaldi is started with Qt {v.toString()} \
-but requires at least version {r.toString()}.")
+qt_version = QLibraryInfo.version()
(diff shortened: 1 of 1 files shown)


bmjcode:
~Shall I merge it?~ I just saw you already approved, so done. :)

frescobaldi/checks.py
```diff
@@ -96,6 +109,19 @@ def error(title, message, can_continue=False):
...
+try:
+    from PyQt6.QtPdf import QPdfLinkModel
+except ImportError:
+    error("Point-and-click unavailable",
+        f"You are using an older Qt version ({qt_version.toString()}) "
+        "that lacks support for links in PDF documents. "
+        "Frescobaldi will run, but point-and-click will be unavailable."
+        "\n\n"
+        "To fix this, install Qt 6.6 or newer.",
+        can_continue=True)
```
igneus: Does it make sense to use Frescobaldi when point and click is not available?
I would hardly notice not having SVG support, but without point and click there's IMO no point in running the editor.

igneus: ... too late. Sorry, I've had the tab open for some time, only got to writing the comment now and didn't notice the PR has been merged in the meantime. Ignore the comment.

PR #2172, opened 2026-03-09
LilyPond preferences: allow Custom menu in flatpak

### LINKED ISSUE #2055: LilyPond development version bundled in Flatpak cannot be used
From #1672 :
  
> * Change the "Custom..." button of the "Add" menu in the LilyPond preferences, so that it displays an informative message if running under Flatpak, and refuses to proceed.

Unfortunately this is causing a problem. I've just noticed in a clean flatpak installation that I cannot add the development version bundled in the flatpak package (which is not automatically detected). The best fix would be to detect all the lilypond installations: not only the first lilypond instance in the PATH but all the occurrences:

```
$ flatpak run --command=sh org.frescobaldi.Frescobaldi 
[📦 org.frescobaldi.Frescobaldi frescobaldi]$ which -a lilypond
/app/bin/lilypond
/app/dev/bin/lilypond
```

Summary (and memo also for myself):

* Yes, I know that I can download the development version and use it in the flatpak installation. But why should I do it if it's already available?
* No, I cannot/don't want to remove the unstable lilypond version in flatpak, because there's no LilyPond binary for aarch64 (architecture supported by Flathub).
* The best fix is improving the LilyPond binary detection
* A quick solution could be relaxing the flatpak filter and allow the Custom installation in flatpak installations. We may just warn the user that a self-compiled (not static) lilypond build is not going to work inside flatpak.

fedelibre:
Change the Error to Warning.

Fix #2055

### DIFF
--- a/frescobaldi/preferences/lilypond.py
+++ b/frescobaldi/preferences/lilypond.py
@@ -204,13 +204,13 @@ class InfoList(widgets.listedit.ListEdit):
     def createItem(self):
         if linux.inside_flatpak():
             QMessageBox.warning(self, app.caption(_("Warning")),
-                _("""You are using a Flatpak installation of Frescobaldi. """
-                  """Use this menu only to add a flatpak-bundled LilyPond installation """
-                  """that was accidentally removed or not detected by Frescobaldi. """
-                  """Because Flatpak isolates apps from the rest of the system, custom LilyPond """
-                  """binaries are not supported. Please either let Frescobaldi auto-install """
-                  """a given LilyPond version, using the “Add…” menu, or if you wish to use """
-                  """a custom build, install Frescobaldi from a distribution package or from source."""))
+                _("You are running a Flatpak installation of Frescobaldi. Use this menu "
+                  "only to add a LilyPond installation in /app (flatpak bundle) that "
+                  "was accidentally removed or not detected by Frescobaldi.\n\n"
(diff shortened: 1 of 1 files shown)


frescobaldi/preferences/lilypond.py
```diff
@@ -203,13 +203,14 @@ def updateSelection(self):
...
         if linux.inside_flatpak():
-            QMessageBox.critical(self, app.caption(_("Error")),
-                                 _("""You have installed Frescobaldi from Flathub using \
-Flatpak. Because Flatpak isolates apps from the rest of the system, it is not possible \
-to use custom LilyPond binaries in this configuration. Please either let Frescobaldi auto-install \
-a given LilyPond version, using the “Add…” menu, or if you wish to use a custom build, \
-install Frescobaldi using a distribution package or from source."""))
-            return None
+            QMessageBox.warning(self, app.caption(_("Warning")),
+                _("""You are using a Flatpak installation of Frescobaldi. """
```
bmjcode: Regular quotes (`"`) would work fine here. Triple quotes (`"""`) are useful if you have multiple lines and need to preserve the breaks, which you'd have to escape as `\n` otherwise. They're also conventional for [docstrings](https://peps.python.org/pep-0257/).

I would personally write this just like you did, except with regular instead of triple quotes.

fedelibre:
I've slightly changed the formatting and the first two sentences to make it more clear.

image

Issue #2160, opened 2026-02-13
Homebrew Frescobaldi 4.0.5 crashing on Tahoe 26.2 (25C56)


dmingram66:
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------
Process:             Frescobaldi [37883]
Path:                /Applications/Frescobaldi.app/Contents/MacOS/Frescobaldi
Identifier:          org.frescobaldi.frescobaldi
Version:             4.0.5 (1)
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           org.frescobaldi.frescobaldi [42626]
User ID:             501

Date/Time:           2026-02-13 09:31:13.5297 +0000
Launch Time:         2026-02-13 09:31:10.5798 +0000
Hardware Model:      MacBookPro18,2
OS Version:          macOS 26.2 (25C56)
Release Type:        User

Crash Reporter Key:  EC77B8D2-B379-207A-8336-7360B1594AD2
Incident Identifier: 72463AF5-6055-4853-B299-17EB2297EC16

Sleep/Wake UUID:       E1523A22-24E7-4E68-97D5-267B7901C62E

Time Awake Since Boot: 260000 seconds
Time Since Wake:       88481 seconds

System Integrity Protection: enabled

Triggered by Thread: 0, Dispatch Queue: com.apple.main-thread

Exception Type:    EXC_BREAKPOINT (SIGTRAP)
Exception Codes:   0x0000000000000001, 0x00000001972a1c7c

Termination Reason:  Namespace SIGNAL, Code 5, Trace/BPT trap: 5
Terminating Process: exc handler [37883]

Application Specific Backtrace 0:
0   CoreFoundation                      0x00000001929718ec __exceptionPreprocess + 176
1   libobjc.A.dylib                     0x000000019244a418 objc_exception_throw + 88
2   Foundation                          0x0000000194abd424 -[NSMutableDictionary(NSMutableDictionary) initWithContentsOfFile:] + 0
3   AppKit                              0x0000000197db2a5c -[NSImageSymbolRepProvider _bestRepresentationForImage:hints:] + 3376
...
83  Python                              0x0000000101fe9ad0 builtin_exec + 396
84  Python                              0x0000000101ed5730 cfunction_vectorcall_FASTCALL_KEYWORDS + 88
85  Python                              0x0000000101ffb048 _PyEval_EvalFrameDefault + 35876
[...]

PR #2188, opened 2026-04-06
Workaround for Qt 6.10.1 macOS crash

sherrman:
This PR's workaround addresses a macOS crash reported in: https://github.com/frescobaldi/frescobaldi/issues/2160

Note: I used Claude Code heavily. This is my first time looking at the code for Frescobaldi and Qt. It took a lot of guiding Claude and lldb to root cause and reproduce the bug in C++.

The root cause is a bug in Qt itself. My priority is getting Frescobaldi working on macOS. Later I'll try to submit the C++ reproduction to the Qt bug tracker.

## The workaround

A previous proposal (in #2160) felt more intrusive and hacked. The change here basically does the same thing by disabling the usage of system icons on macOS.

## Demonstration of crash

Here's a reproduction of the crash. Run with PyQt 6.9.1 and it works. I've tested this with PyQt 6.10.1 and it crashes. I have not tried this bug reproduction with 6.11.1 (but Frescobaldi with 6.11.1 crashes with a similar if not same stack). Note: The following was written completely by Claude Code.

```py
"""
Minimal PyQt6 reproducer for QAppleIconEngine crash in Qt 6.10.1 on macOS.

Root cause: QAppleIconEngine::actualSize() uses integer division
(size.width() / size.height()) instead of floating-point division.
When width >> height for a non-square SF Symbol, the result truncates
to a zero-width size, which crashes AppKit's symbol rep provider.

Run with PyQt6 linked against Qt 6.10.1 to trigger the crash.
"""

import sys
from PyQt6.QtWidgets import QApplication
from PyQt6.QtGui import QIcon
from PyQt6.QtCore import QSize

def main():
    app = QApplication(sys.argv)

    # "edit-paste" maps to SF Symbol "document.on.clipboard" (native {17, 18}).
    # Any non-square SF Symbol icon should work.
    icon = QIcon.fromTheme("edit-paste")
    if icon.isNull():
        print("edit-paste icon not found (not on macOS or no SF Symbols?)")
        sys.exit(1)

    print("Requesting pixmap at size (11, 1) — this triggers the crash on Qt 6.10.1")
    print("  actualSize({11, 1}) does integer division: 11/1 = 11")
    print("  11 > 0.944 (icon aspect ratio) → rwidth = 1 * 0.944 = 0 (int truncation)")
    print("  Result: {0, 1} → paint() calls [NSImage setSize:{0,1}] → NaN → crash")

    # This is the crash trigger.
    pm = icon.pixmap(QSize(11, 1))

    print(f"No crash — bug is not present. Pixmap result: {pm.size().width()}x{pm.size().height()}")

if __name__ == "__main__":
    main()
```

### DIFF
--- a/frescobaldi/icons/__init__.py
+++ b/frescobaldi/icons/__init__.py
@@ -23,6 +23,7 @@ Icons.
 
 
 import os
+import sys
 
 from PyQt6.QtCore import QDir, QFile, QFileInfo, QSettings, QSize
 from PyQt6.QtGui import QIcon
@@ -86,6 +87,13 @@ def initialize():
 
 def update_theme():
     """Change the theme for icon lookup after change of style preference"""
+    if sys.platform == "darwin":
+        # On macOS, QAppleIconEngine (SF Symbols) crashes with zero-dimension
+        # rects in PyQt6 6.10.1+ due to integer division in actualSize().
+        # Always use our bundled theme to avoid the crash. See:
+        # https://bugreports.qt.io/browse/QTBUG-TBD
+        QIcon.setThemeName("TangoExt")
+        return
     s = QSettings()
     if s.value("system_icons", True, bool):
         QIcon.setThemeName(s.value("guistyle", "", str))
--- a/frescobaldi/icons/__init__.py
+++ b/frescobaldi/icons/__init__.py
@@ -23,7 +23,7 @@ Icons.
 
 
 import os
-import sys
+import platform
 
 from PyQt6.QtCore import QDir, QFile, QFileInfo, QSettings, QSize
 from PyQt6.QtGui import QIcon
@@ -87,7 +87,7 @@ def initialize():
 
 def update_theme():
     """Change the theme for icon lookup after change of style preference"""
-    if sys.platform == "darwin":
+    if platform.system() == "Darwin":
         # On macOS, QAppleIconEngine (SF Symbols) crashes with zero-dimension
         # rects in PyQt6 6.10.1+ due to integer division in actualSize().
         # Always use our bundled theme to avoid the crash. See:
--- a/frescobaldi/preferences/general.py
+++ b/frescobaldi/preferences/general.py
@@ -22,6 +22,8 @@ Keyboard shortcuts settings page.
 """
 
 
+import sys
+
 from PyQt6.QtCore import QSettings
 from PyQt6.QtWidgets import (
     QApplication,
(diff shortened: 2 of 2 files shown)


frescobaldi/icons/__init__.py
```diff
@@ -86,6 +87,13 @@ def initialize():
 
 def update_theme():
     """Change the theme for icon lookup after change of style preference"""
+    if sys.platform == "darwin":
```
bmjcode: Frescobaldi's preferred check is to use `platform.system()`. See [CONTRIBUTING.md](https://github.com/frescobaldi/frescobaldi/blob/master/CONTRIBUTING.md#contributing-coding-style).

frescobaldi/preferences/general.py
```diff
@@ -163,6 +173,11 @@ def translateUI(self):
         self.systemIcons.setToolTip(_(
             "If checked, icons of the desktop icon theme "
             "will be used instead of the bundled icons."))
+        if sys.platform == "darwin":
+            self.systemIconsWarning.setText(_(
+                "System icons are disabled on macOS due to a crash in "
+                "PyQt6 6.10.1+. The bundled icon theme is "
+                "used instead."))
```
bmjcode: I appreciate the intent here, but adding this string means more maintenance burden for our translators, which again is excessive for a temporary workaround.

bmjcode (changes requested):
Thanks for the attempt @sherrman, but frankly I thought the simpler fix you proposed in #2160 was better. This makes needlessly excessive changes -- including rearranging the settings dialog -- for what is supposed to be a temporary workaround. I've added some comments explaining my concerns in more detail.

I'd also ask that you please not rely on AI assistance or vibe coding, for two reasons. First, we want to keep Frescobaldi a tool created with and for human creativity. Second, more pragmatically, I've found checking behind the AI and correcting its errors usually takes longer than implementing something myself manually.

If you have questions or need help understanding the codebase, please, just ask us! We are happy to mentor (we were new to this once too), and we love talking about our work. :)

sherrman:
AI stuff noted and I totally understand. I'm ok to abandon this PR if you'd like.

The code change is now trivial. It did write a good comment, one that I'm often too lazy to write myself lol.

I removed all the settings changes. And that makes sense. What do you think about the change now? It overrides the settings and instead always uses the non-system icons for Mac. The clarity of the workaround could go either way IMO.

We either ignore which theme (system or app) in the settings at the time of theme setter, or we simply override the theme at time of icon getter.

image

bmjcode:
> We either ignore which theme (system or app) in the settings at the time of theme setter, or we simply override the theme at time of icon getter.

My thought is we do whatever requires the smallest change since it's only a temporary workaround, and presumably we'd want to revert it once the actual bug is fixed.

sherrman:
> > We either ignore which theme (system or app) in the settings at the time of theme setter, or we simply override the theme at time of icon getter.
> 
> My thought is we do whatever requires the smallest change since it's only a temporary workaround, and presumably we'd want to revert it once the actual bug is fixed.

Sounds good. Let's keep the change as is in the PR. It's a self contained if statement that just gets deleted when the workaround is done.

But for the bigger issue, how about we work on resolving it on the issue #2160?

bmjcode:
While we're waiting for the Qt folks to investigate this, I think our best bet is just to ship an older Qt version on macOS per #2190. That way our Mac users aren't losing any functionality, even purely cosmetic things like native icons.

inkandpaper-app:
Could you reconsider not using system icons, especially now that they're causing problems? As you can see from the attached video, the theme is inconsistent (also note the app icon changing when opening and closing the program). 
I think the solution I proposed here https://github.com/frescobaldi/frescobaldi/pull/2067 is good, but it hasn't been considered. I haven't yet determined whether it's been tested and what problems it might cause.

https://github.com/user-attachments/assets/7acd5159-a4ac-4c18-a041-5b8d232becba

PR #2190, opened 2026-04-07
Downgrade PyQt6 to 6.9 for macOS packages + re-enable macOS signing

fedelibre:
We must downgrade PyQt6 for macOS packages to version 6.9 because of a crash discussed in #2160.

This is a quick workaround to that problem, until we have a better understanding of the Qt bug.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -13,10 +13,10 @@ jobs:
 
     steps:
       - name: Checkout PR repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Checkout Flathub manifest
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           repository: flathub/org.frescobaldi.Frescobaldi
           path: flathub-manifest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,9 +53,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: ["Windows"]
-        # macOS disabled until we have a new certificate
-        #target: ["Windows", "macOS", "macOS-Intel"]
+        target: ["Windows", "macOS", "macOS-Intel"]
         include:
           - target: "Windows"
             platform: "windows"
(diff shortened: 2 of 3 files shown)


bmjcode (approved):
This seems like a reasonable workaround to me.

fedelibre:
We have a problem: the certificate needed to sign macOS packages seems [expired](https://github.com/frescobaldi/frescobaldi/actions/runs/24124351138/job/70385291327).

@jacques-menu Could you please renew it?

fedelibre:
Now I remember that we cannot distribute unsigned packages for macOS.
briefcase has the --adhoc-sign option which allows to build an unsigned package but this is just for personal testing; it won't work on any machine other than the machine where it was built. So it's useless for us.

I don't think we want to depend on someone paying Apple $99 per year.

I would disable the building of .dmg files and find a way to let macOS users run it from source.

This is a frustration rant... I'll see if I come up with a better idea in the future. But I don't want to further delay the Frescobaldi release.

fedelibre:
@kledeen Can you check if this patch fixes the crash you are seeing on macOS using Frescobaldi >=4.0.5?
Unfortunately we cannot build macOS packages at the moment. See below instructions to run Frescobaldi from source code.

## Instructions

I can't remember if you've ever run Frescobaldi from source.
If you haven't, first download it:

    git clone git@github.com:frescobaldi/frescobaldi.git

Otherwise, enter frescobaldi repository and just run a pull to update:

    git pull

Then checkout this branch:

    git checkout macos-qt-downgrade

Now you can [use Briefcase](https://github.com/frescobaldi/frescobaldi/blob/master/PACKAGES.md#briefcase) to build and run Frescobaldi. Can you run it without crashes?

fedelibre:
Thanks to @walkymatt for providing a new certificate! We can make macOS packages again.

@kledeen You can test this PR by installing the DMG files you can find here (you must be logged in with your Github account):
https://github.com/frescobaldi/frescobaldi/actions/runs/25389551147

kledeen:
Hi Frederico,

I tested it. Crashed on startup

The crash occurred while Frescobaldi was trying to *draw a toolbar button
icon*. Specifically, macOS's AppKit threw an Objective-C exception when
trying to render an SF Symbol (a system icon), and instead of catching it
gracefully, the app hit +[NSApplication _crashOnException:] — meaning the
exception propagated all the way up uncaught.

The key frames in the stack trace tell the story:

   1. A QToolButton needed to be painted (a button in Frescobaldi's toolbar)
   2. Qt asked for the button's icon via QAppleIconEngine
   3. That called into AppKit's NSImageSymbolRepProvider to look up a
   symbol image
   4. AppKit threw an exception from NSMutableDictionary
   initWithContentsOfFile: — likely a *corrupted or missing symbol
   configuration file*, or a *compatibility issue* between the bundled Qt
   version and macOS 26.2
   5. The exception was uncaught → crash

Hope this helps.  Let me know if you want the full dump.  I hope I was
running the correct build.  It was Frescobaldi-4.0.7.dev0

Best,

Ken

fedelibre:
I thought that this should not occur with Qt 6.9.
Please paste here all the versions you see in Help>About Frescobaldi>Version and also the full dump.

fedelibre:
> @kledeen You can test this PR by installing the DMG files you can find here (you must be logged in with your Github account):
> https://github.com/frescobaldi/frescobaldi/actions/runs/25389551147

@sherrman @dmingram66 Can you also test those DMG files?

I want to understand if downgrading Qt to 6.9 actually helps or not. Thanks.

sherrman:
> > @kledeen You can test this PR by installing the DMG files you can find here (you must be logged in with your Github account):
> > https://github.com/frescobaldi/frescobaldi/actions/runs/25389551147
> 
> @sherrman @dmingram66 Can you also test those DMG files?
> 
> I want to understand if downgrading Qt to 6.9 actually helps or not. Thanks.

I'm on Tahoe 26.4.1 (25E253) now fwiw.

Happened to have 4.0.5 installed; that crashes.

This build 4.0.7.dev0 runs fine!

PR #2191, opened 2026-04-08
Add QoL option to lilypond not found warning

### LINKED ISSUE #1657: Automatically suggest installing LilyPond when no installation is configured
Currently, if you don't have LilyPond installed (neither `lilypond` command on `PATH`, nor installation configured in the Preferences), you get

image

But it would be nicer to at least have a button in that dialog to open the preferences. Better yet, detect on startup (not on the first compilation) that LilyPond is not installed, and open the Preferences directly. Especially in conjunction with #313, that would make for an easier onboarding experience.

JaumePujantell:
This adds a button to oppen the relevant Preferences dialog on the checkLilyPondInstalled warning dialog. And changes the text to make it clearer that lilypond can be isntalled from frescobaldi itself.

Fix #1657

### DIFF
--- a/frescobaldi/engrave/__init__.py
+++ b/frescobaldi/engrave/__init__.py
@@ -26,7 +26,7 @@ import os
 
 from PyQt6.QtCore import QSettings, Qt, QUrl
 from PyQt6.QtGui import QAction, QKeySequence, QTextCursor
-from PyQt6.QtWidgets import QApplication, QMessageBox
+from PyQt6.QtWidgets import QApplication, QMessageBox, QPushButton
 
 import app
 import actioncollection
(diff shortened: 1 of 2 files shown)


fedelibre:
Looks good to me, from the user point of view. Ben will comment on the code.

Here's what it looks like (remove the LilyPond installations and compile a file to reproduce):

image

bmjcode:
This looks really nice. I'd maybe tweak the wording of the second line of instructions -- something like:

> Open the Preferences dialog to download LilyPond. If you have an existing installation that was not detected, you can also add it manually there.

(This is off the top of my head, so it might not be the clearest possible wording. What I want to emphasize is the ability to download and install LilyPond automatically.)

Other than that suggestion I am good with merging this.

JaumePujantell:
Thanks, I've modified the text with the suggestion.

fedelibre (approved):

bmjcode:
Let's do this!

PR #2192, opened 2026-04-19
lilypond.exe is the correct file name

### LINKED ISSUE #1581: lilypond-windows.exe is no longer distributed by LilyPond >=2.24
When you install Frescobaldi on Windows, it automatically adds `lilypond-windows.exe` to the LilyPond preferences, but this file was removed from LilyPond binaries since version 2.24. We should use `lilypond.exe`.

Schermata del 2023-04-05 10-08-21

fedelibre:
used by LilyPond packages for Windows.
This default value won't work until the user
adds the LilyPond package location to the PATH.

Fix #1581

### DIFF
--- a/frescobaldi/lilypondinfo.py
+++ b/frescobaldi/lilypondinfo.py
@@ -100,11 +100,11 @@ def saveinfos():
 def default():
     """Returns a default LilyPondInfo instance with the default LilyPond command.
 
-    On Windows, the default command is "lilypond-windows.exe",
+    On Windows, the default command is "lilypond.exe",
     on other platforms simply "lilypond".
 
     """
-    lilypond = "lilypond-windows.exe" if platform.system() == "Windows" else "lilypond"
+    lilypond = "lilypond.exe" if platform.system() == "Windows" else "lilypond"
     return LilyPondInfo(lilypond)
 
 
@@ -118,7 +118,7 @@ def preferred():
     s = QSettings()
     s.beginGroup("lilypond_settings")
     # find default version
-    defaultCommand = "lilypond-windows.exe" if platform.system() == "Windows" else "lilypond"
+    defaultCommand = "lilypond.exe" if platform.system() == "Windows" else "lilypond"
     userDefault = s.value("default", defaultCommand, str)
     if userDefault != defaultCommand:
         for info in infos_:


fedelibre:
Ok, this is probably the simplest patch ever :-)

At least we have the right binary name.
A nice have could be detecting that lilypond.exe cannot be found and ask the user to provide the correct path (which we cannot guess). What do you think?

fedelibre:
Actually the missing binary has already been handled by #2191 

So this simple patch can be useful only for those that already have lilypond.exe in the PATH.

bmjcode:
> So this simple patch can be useful only for those that already have lilypond.exe in the PATH.

That's a pretty small group, I think. Windows users wouldn't have LilyPond in their PATH unless they added it manually (and I don't remember the old installers even offering to). I saw the discussion about Chocolatey, but I'm not sure I understand what their needs are since I've never used that software.

fedelibre:
I don't have the time to test it on Windows today, but this is my understanding:
If you install LilyPond, or Frescobaldi and LilyPond, via Chocolatey, lilypond.exe will be in the PATH.
The maintainer of the Chocolatey packages asked how to change the default binary name for Windows to lilypond.exe, which is what this PR does.

According to the [numbers on Chocolatey website](https://community.chocolatey.org/packages?q=frescobaldi), there are several downloads for LilyPond (11k) and Frescobaldi (1,7k).

I think I'll test it tonight.

bmjcode:
I mean, the code looks fine to me. Changing the default to `lilypond.exe` shouldn't break anything even with older LilyPond versions. I'm just not sure I understand why the Chocolatey package is relying on LilyPond being in the `PATH`, since that's unreliable and non-standard on Windows. *(edited to clarify my specific point of confusion)*

fedelibre:
> I mean, the code looks fine to me. Changing the default to `lilypond.exe` shouldn't break anything even with older LilyPond versions. I'm just not sure I understand why the Chocolatey package is relying on LilyPond being in the `PATH`, since that's unreliable and non-standard on Windows. _(edited to clarify my specific point of confusion)_

It's Frescobaldi which relies on lilypond-windows.exe (currently) being in the PATH, as this is how things use to work on Linux and Wilbert the original author of Frescobaldi is a Linux user.

I did a quick test and installed Frescobaldi 4.0.5 from Chocolatey: LilyPond (chocolatey package) is automatically installed and then found by Frescobaldi.
I can see that lilypond.exe is located in a path present in the PATH variable, that is `C:\ProgramData\chocolatey\bin`.

My question for @teknowledgist is: as current Frescobaldi is looking for lilypond-windows.exe and chocolatey (correctly) provides lilypond.exe, did you change some Frescobaldi setting to make it work?
I wonder if this PR is going to break your Chocolatey package.

teknowledgist:
It's been a couple years since I thought about this... 

The Chocolatey package for Frescobaldi does install Lillypond (and VC++) as dependencies, but otherwise it is about as bare-bones as you can get. You can [see the install parameters here](https://community.chocolatey.org/packages/frescobaldi#files), but basically, it identifies the URL and checksum and lists the arguments for a silent, machine-wide, logged install, and then kicks off the install. 

I never got around to tracking down where the "lillypond-windows.exe" setting is stored so that I could then have the package modify it. It sounds like I don't need to now! 😄

bmjcode (approved):
Well, either way this fixes more things than it potentially breaks, so let's merge it.

PR #2197, opened 2026-04-25
Rethink (and document) low-DPI options for copy-to-image

bmjcode:
My thought is there are two major use cases for this feature: screenshots, which should match common screen densities, and inserting snippets into documents, which should match common print densities.

This patch adds 96 and 192 dpi as the lower-DPI options for screen use. It removes 72, 100, and 200 dpi as standard options, since they are rarely used now for either screen or print due to their poor quality.


bmjcode:
Note my initial version of this PR just added 96 dpi as an option. I rewrote it because, after further consideration, it didn't make sense to have both 96 and 100 dpi as options with no clear distinction between them. With things like this, I don't think we need to provide every possible option, just the ones that are most likely to be useful to our users.

fedelibre (approved):

Issue #2159, opened 2026-02-12
Memory leak after extended use in 4.0.5


bameylan:
Merci de décrire le problème ou la demande de fonctionnalité.
Fournissez un maximum d'informations.

Frescobaldi: 4.0.5
Extension API: 0.9.0
Python: 3.12.12
python-ly: 0.9.9
Qt: 6.9.3
PyQt: 6.9.1
qpageview: 1.0.3
OS: Linux-6.17.0-14-generic-x86_64-with-glibc2.40 (KDE Flatpak runtime)
Installation kind: Flatpak

Avec la mise à jour 4.04 -> 4.05, l'auto complétion ne fonctionne plus.
With the 4.04 -> 4.05 update, auto-completion no longer works.

Cordialement.

Bernard Meylan

fedelibre:
I've just tried 4.0.5 flatpak package and autocompletion seems to work normally:

<img width="266" height="234" alt="Image" src="https://github.com/user-attachments/assets/e76f1442-1a00-407d-84c1-bc9e43b133ce" />

Can you show what exactly is not working for you?

bmjcode:
I'm curious too. We didn't change anything that affects autocomplete between 4.0.4 and 4.0.5.

bameylan:
Actually, I should have talked about the "erratic behavior" of the auto-completion, not its malfunction. The situation is the same as weth _issue #473_: a file of almost 2000 lines (but not very "heavy": 82 KB), which means that after about 30 minutes of work, Frescobaldi reacts very slowly; if I type notes too quickly, some don't appear. It's probably the same problem that happened with the auto-completion… I didn't wait long enough. Because I'm currently working with a smaller file, and there's no longer a problem. Sorry if I alerted the community too soon…

fedelibre:
Ok, then it's the good old #473 issue. Let's close this issue.

There's a promising PR currently open, see #2152 
Would you like to test it? Can you figure out how to run that branch from source or do you need any help? I might even create a flatpak artifact if this is convenient for you.

bameylan:
I don't have the knowledge to test #2152; I'm not very skilled. But with help, I'd be willing to try.

bmjcode:
@bameylan All you need to do is install and run it for a bit, then tell us two things:

1. Does it run faster?
2. Does it crash?

fedelibre:
[...]

PR #2202, opened 2026-05-03
Reduce link cache memory usage

bmjcode:
Previously link areas were cached as QRectF objects, which carry a lot of overhead. This patch reduces memory usage by caching them as plain tuples and only converting them to QRectF when needed.

This may help with the memory leak identified in #2159.

### DIFF
--- a/frescobaldi/musicview/pointandclick.py
+++ b/frescobaldi/musicview/pointandclick.py
@@ -34,8 +34,6 @@ import util
 import textedit
 import pointandclick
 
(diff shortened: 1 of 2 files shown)


fedelibre (approved):
It seems reasonable.

I cannot test this PR as I should have a real LilyPond file to work on... and  I've not used Frescobaldi for typesetting for months..

ksnortum:
Just to reassure you, I've been working with this PR  for days and the memory usage and way down.

PR #2203, opened 2026-05-03
Fix a potential memory leak by circular reference

bmjcode:
This patch changes the worker's reference to the document to a weak reference to prevent both objects keeping one another in memory longer than needed by circular reference.

This may help with the memory leak identified in #2159.

### DIFF
--- a/frescobaldi/documentinfo.py
+++ b/frescobaldi/documentinfo.py
@@ -27,6 +27,7 @@ import functools
 import os
 import re
 import collections
+import weakref
 
 from PyQt6.QtCore import (
     QCoreApplication, QObject, QSettings, QThread, QTimer, QUrl, pyqtSignal)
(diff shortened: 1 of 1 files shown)


ksnortum:
I checked out your PR and did several engravings while watching a resource monitor for memory usage and the Frescobaldi task keeps using more memory each engraving.  Ubuntu Linux 25.10, Frescobaldi 4.0.7dev0.

bmjcode:
As it turns out, once again we can [blame qpageview](https://github.com/frescobaldi/qpageview/pull/52).

This is one of those PRs that I thought made sense when I wrote it, but looking at it again, I'm not sure it actually does anything. Suffice it to say I find Python's approach to memory management deeply confusing and difficult to debug. Where it might help if it does anything is freeing memory when closing a document, though that's a relatively small improvement.

fedelibre:
@bmjcode  Do you want set this PR to draft or close it?

bmjcode:
Let's close this. I'm fairly certain now I had misunderstood what was keeping what alive here, so this PR actually does nothing.

Issue #2097, opened 2025-09-30
AttributeError: module 'snippet.model' has no attribute 'collection'


eefweenink:
I am importing snippets.xml(s) which were created on another macbook. Regards, Eef 

Traceback (most recent call last):
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/snippet/widget.py", line 305, in slotImport
    import_export.load(filename, self)
  File "/Applications/Frescobaldi.app/Contents/Resources/app/frescobaldi/snippet/import_export.py", line 207, in load
    ac = model.collection()
         ^^^^^^^^^^^^^^^^
AttributeError: module 'snippet.model' has no attribute 'collection'

Please describe below what you were doing:

Frescobaldi: 4.0.4
Extension API: 0.9.0
Python: 3.12.8
python-ly: 0.9.9
Qt: 6.9.0
PyQt: 6.9.1
qpageview: 1.0.1
OS: macOS-15.7-x86_64-i386-64bit
installation kind: standalone .app bundle

bmjcode:
This is on my radar. I just haven't had too much time to focus on Frescobaldi stuff lately.

PR #2213, opened 2026-06-06
Fix error on template and snippet import

oleastre:
I just wanted to import a simple template from another computer and was stuck with #2097 . 
I took a quick look at the existing changes and this patch should solve the issue (at least I managed to import my template)

### DIFF
--- a/frescobaldi/snippet/import_export.py
+++ b/frescobaldi/snippet/import_export.py
@@ -40,6 +40,7 @@ import userguide
 import widgets.dialog
 
 from . import model
+from . import tool
 from . import snippets
 from . import builtin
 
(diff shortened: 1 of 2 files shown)


bmjcode:
I'm not super familiar with that part of the code, but from a quick glance this looks correct.

Before I merge this, I noticed there's an apparently unused `import actioncollection` on [line 32 of model.py](https://github.com/frescobaldi/frescobaldi/blob/1e26308f62077de1f51fef0b0e71c35bef67009d/frescobaldi/snippet/model.py#L32). Does anything break if that's removed? I'm thinking the code now in tool.py must have been moved from model.py at some point, but if model.py no longer needs that import, let's remove it while we're here.

oleastre:
I'll test this during the week. 
On the change itself, I had a look at the change log, a bit of playing with `git blame` to find where that "collections" was moved and changed the reference to point to the new location.
My main use case was to import a template, I did not test the shortcut import. But at least it does not crash any more.

bmjcode:
Very nice work @oleastre. Like I said, I think this is the correct fix. The only reason I'm waiting to merge it is so we can also clean up that import without having to open a separate PR.

oleastre:
Just pushed the remove unused import patch.

bmjcode (approved):
Looks good. Let's merge this.

PR #2235, opened 2026-08-10
Removed the color from a couple of icons to improve their readability.

inkandpaper-app:


inkandpaper-app:
The approach in the latest commit should be a better alternative that keeps the colors. Regardless of that, I noticed that in the dark theme, the color of unselected tabs is white—at least on macOS—which shouldn't be the case. There doesn't seem to be a way to change this within the theme's color settings.

Screenshot 2026-08-11 alle 13 53 02

bmjcode:
Gotta love those weird Mac things 😂 I haven't had a chance to try this yet but I am watching the PR thread.

fedelibre:
I like the traffic light icons.

Light theme:

immagine

Dark theme:

immagine

bmjcode:
The traffic light icons look nice, if a little small on my Windows system:

image

My one concern would be whether they're clear enough for colorblind users. Can anyone provide feedback?

I'm not sure I understand the QTabBar style changes. Is that CSS code applied on all platforms, or only macOS? I did see the discussion earlier in the thread, and if those changes fix the problem I'm OK with including them. I just want to understand exactly what's being changed before I approve merging them.

ksnortum:
I am red/green color blind and the icon is clear to me.  If there were another icon with the traffic signal and a red light it might be confusing, but there isn't (I think!)

inkandpaper-app:
The style comment, as you can see from the lines just above it, only applies to MacOS. It was likely written before the dark theme was introduced. I removed the light background color requirement for tabs, which prevented the text from being visible when the dark theme was selected.

Screenshot 2026-09-05 alle 08 55 49

bmjcode:
Cool, then let's merge this.

